PR #12, opened 2026-08-13
chore: split mise mac and aws profiles

lmilbaum:
## Summary
- keep common tools in shared mise.toml
- add AWS-specific mise profile
- add WIP mac profile that identifies itself only
- document the profile layout in README

## Validation
- mise -E mac config
- mise -E mac,aws config
- git diff --check


mise.aws.toml
```diff
@@ -0,0 +1,3 @@
+[tools]
+aws-cli = "latest"
+clusterawsadm = { version = "2.10.2", binary = "clusterawsadm" }
```
shrinedogg: This just needs updated to "2.13.0" to match upstream.

shrinedogg (approved):
LGTM!

PR #15, opened 2026-08-13
feat: make management lifecycle profile aware

lmilbaum:
## Summary
- make bootstrap select mac or AWS behavior from the mise profile environment
- keep the management kind cluster command shared by both profiles
- move the AWS bootstrap task into mise.aws.toml
- make teardown delete only kind for mac and retain full AWS cleanup for AWS
- document the profile-aware lifecycle in the README

## Validation
- bash -n bootstrap.sh teardown.sh
- mise -E mac tasks
- mise -E aws tasks
- git diff --check

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -3,28 +3,39 @@
 # Everything after this script runs is driven by GitOps (Flux).
 set -euo pipefail
 
+PROFILE="${KNR_OPS_PROFILE:-${1:-aws}}"
+case "$PROFILE" in
+  mac|aws) ;;
+  *)
+    echo "ERROR: unsupported profile '$PROFILE' (expected 'mac' or 'aws')" >&2
+    exit 1
+    ;;
+esac
+
 # ── Prerequisites check ───────────────────────────────────────────────────────
 for cmd in kind helm kubectl; do
   command -v "$cmd" >/dev/null 2>&1 || { echo "ERROR: $cmd not found in PATH"; exit 1; }
 done
 
-: "${GITHUB_TOKEN:?GITHUB_TOKEN must be set (a PAT with read access to the repo)}"
-: "${GIT_REPO_URL:?GIT_REPO_URL must be set}"
+if [ "$PROFILE" = aws ]; then
+  : "${GITHUB_TOKEN:?GITHUB_TOKEN must be set (a PAT with read access to the repo)}"
+  : "${GIT_REPO_URL:?GIT_REPO_URL must be set}"
 
-# GITHUB_USER is only used as the basic-auth username; any non-empty value
-# works with a GitHub PAT, so default to "git".
-GITHUB_USER="${GITHUB_USER:-git}"
+  # GITHUB_USER is only used as the basic-auth username; any non-empty value
+  # works with a GitHub PAT, so default to "git".
+  GITHUB_USER="${GITHUB_USER:-git}"
 
-# ── SOPS age key ──────────────────────────────────────────────────────────────
-# Flux decrypts SOPS-encrypted secrets in Git (e.g. the CAPA AWS credentials)
-# using an age private key loaded into the cluster as the `sops-age` secret.
-# AGE_KEY_FILE defaults to ./age.agekey (gitignored).
-AGE_KEY_FILE="${AGE_KEY_FILE:-age.agekey}"
-if [ ! -f "$AGE_KEY_FILE" ]; then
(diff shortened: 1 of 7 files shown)


bootstrap.sh
```diff
@@ -97,6 +87,32 @@ echo ">>> Waiting for cluster node to be ready..."
...
 kubectl wait --for=condition=Ready node --all --timeout=120s
 
+if [ "$PROFILE" = mac ]; then
+  echo ""
+  echo ">>> Mac profile complete: management kind cluster is ready"
+  echo ">>> No Flux or AWS resources were provisioned"
+  exit 0
+fi
+
+: "${GITHUB_TOKEN:?GITHUB_TOKEN must be set (a PAT with read access to the repo)}"
```
shrinedogg: Warning: Moving the AWS-only env checks (GITHUB_TOKEN, GIT_REPO_URL, and the age-key check below) after kind cluster creation drops the fail-fast behavior main had. Step 1 deletes and recreates an existing capi-mgmt cluster before these run, so a missing GITHUB_TOKEN on the aws path now fails after destroying a potentially live management cluster. Suggest wrapping this block in `if [ "$PROFILE" = aws ]` and keeping it ahead of the kind cluster step (the mac exit above already skips it).

mise.toml
```diff
@@ -57,13 +57,9 @@ done
 """
 
 [tasks.bootstrap]
-description = "Bootstrap the kind management cluster and hand off to Flux"
+description = "Bootstrap the management kind cluster"
```
shrinedogg: Suggestion: This description matches the mac path, but on the default aws path the task still installs Flux and hands off to GitOps. Consider keeping a mention of the Flux handoff so `mise tasks` output stays accurate for the primary workflow.

shrinedogg (changes requested):
## Code Review Summary

**Verdict: Request changes** (1 warning, 2 suggestions)

Verified locally on the PR branch: `bash -n` passes on both scripts, `mise tasks` under the default, mac, and aws profiles shows the intended task layout, `KNR_OPS_PROFILE` propagates through `[env]`, unknown profiles are rejected with exit 1 before any side effects, and CI is green (kustomize build, YAML lint; the konflate skip is expected for fork PRs).

### Warning
- **bootstrap.sh:97**: AWS-only env validation (GITHUB_TOKEN, GIT_REPO_URL, age key) moved after kind cluster creation. See inline comment.

### Suggestions
- **.env.example:32**: still says `mise run aws-bootstrap`, which now errors with "no task aws-bootstrap found" outside the aws profile. Update to `mise -E aws run aws-bootstrap`.
- **mise.toml:60**: the bootstrap description fits the mac path but undersells the default aws path (see inline comment).
- **teardown.sh:36**: `AWS_ONLY=1` combined with the mac profile deletes the kind cluster, even though AWS_ONLY means "skip k8s steps". Edge case, but a short guard or usage note would prevent a surprise.

### Looks good
- Profile selection is validated up front in both scripts and fails closed on unknown values
- aws-bootstrap now lives next to the clusterawsadm tool pin in mise.aws.toml
- The mac early-exit in teardown happens before the EXIT trap is armed, so no double-delete path
- README documents the mac lifecycle accurately (kind-only bootstrap, kind-only teardown)

bootstrap.sh
lmilbaum: Makes sense. Fixing it.

shrinedogg (approved):
Re-review of `667fa2b` after the requested changes.

All four items from the previous review are addressed:

- **bootstrap.sh**: the AWS-only env checks (GITHUB_TOKEN, GIT_REPO_URL, age key) are back ahead of the kind cluster step, gated on the aws profile. Verified locally: a missing GITHUB_TOKEN now aborts at line 21 before any mutation, while the mac profile still skips these checks and exits cleanly after the kind cluster is ready.
- **.env.example:32**: now reads `mise -E aws run aws-bootstrap`, matching the task's new home in mise.aws.toml.
- **mise.toml:60**: the bootstrap description mentions the Flux/GitOps handoff again.
- **teardown.sh**: AWS_ONLY=1 combined with the mac profile now fails fast with a clear error instead of deleting the kind cluster.

Also re-verified on this head: `bash -n` passes on both scripts, unknown profiles are still rejected before any side effects, the default, mac, and aws task layouts are all correct, and CI is green (the konflate skip is expected for fork PRs).

Approving.

PR #16, opened 2026-08-13
fix: pull Flux charts anonymously, require PAT only for git sync

lmilbaum:
## Summary

- Pull the Flux Operator chart anonymously from GHCR.
- Require `GITHUB_TOKEN` only for the AWS profile’s Flux Git sync.
- Keep the Mac profile free of GitHub and AWS credential requirements.
- Document profile-specific dotenv requirements.
- Use an empty Helm registry configuration to prevent cached credentials from being used.

## Profile behavior

- `mac`: creates `capi-mgmt`, installs the Flux Operator anonymously, and stops before GitOps sync or AWS provisioning.
- `aws`: creates `capi-mgmt`, installs the Flux Operator anonymously, configures GitHub and SOPS secrets, and starts GitOps reconciliation.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -108,19 +108,26 @@ echo ">>> Waiting for cluster node to be ready..."
 kubectl config use-context kind-capi-mgmt
 kubectl wait --for=condition=Ready node --all --timeout=120s
 
-if [ "$PROFILE" = mac ]; then
-  echo ""
-  echo ">>> Mac profile complete: management kind cluster is ready"
-  echo ">>> No Flux or AWS resources were provisioned"
-  exit 0
-fi
-
-# ── Step 2: Install the Flux Operator via Helm ────────────────────────────────
+# ── Step 2: Install the Flux Operator ────────────────────────────────────────
 echo ">>> Installing Flux Operator..."
+ANONYMOUS_REGISTRY_CONFIG="$(mktemp)"
+printf '{}\n' > "$ANONYMOUS_REGISTRY_CONFIG"
+cleanup_registry_config() {
+  rm -f "$ANONYMOUS_REGISTRY_CONFIG"
+}
+trap cleanup_registry_config EXIT
 helm install flux-operator oci://ghcr.io/controlplaneio-fluxcd/charts/flux-operator \
   --namespace flux-system \
   --create-namespace \
-  --wait
+  --wait \
+  --registry-config "$ANONYMOUS_REGISTRY_CONFIG"
+
+if [ "$PROFILE" = mac ]; then
+  echo ""
+  echo ">>> Mac profile complete: management kind cluster and Flux Operator are ready"
+  echo ">>> No GitOps sync or AWS resources were provisioned"
+  exit 0
+fi
 
 # ── Step 3: Create GitHub PAT credentials secret ─────────────────────────────
 # Basic-auth secret consumed by Flux's source-controller to clone the repo.
@@ -178,7 +185,8 @@ helm upgrade --install flux \
   --set instance.sync.url="${GIT_REPO_URL}" \
   --set instance.sync.ref=refs/heads/main \
   --set instance.sync.path=capi-mgmt \
-  --set instance.sync.pullSecret=flux-github-pat
+  --set instance.sync.pullSecret=flux-github-pat \
+  --registry-config "$ANONYMOUS_REGISTRY_CONFIG"
 
 # ── Post-bootstrap health check ───────────────────────────────────────────────
 # Verify the Flux controllers are running before declaring success.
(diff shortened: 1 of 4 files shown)


bootstrap.sh
```diff
@@ -108,19 +108,26 @@ echo ">>> Waiting for cluster node to be ready..."
...
+cleanup_registry_config() {
+  rm -f "$ANONYMOUS_REGISTRY_CONFIG"
+}
+trap cleanup_registry_config EXIT
 helm install flux-operator oci://ghcr.io/controlplaneio-fluxcd/charts/flux-operator \
   --namespace flux-system \
   --create-namespace \
-  --wait
+  --wait \
+  --registry-config "$ANONYMOUS_REGISTRY_CONFIG"
```
shrinedogg, line 123: Warning: The FluxInstance chart pull in step 4 (line 175, outside this diff hunk) hits the same ghcr registry but does not pass `--registry-config`, so on the aws path a stale cached ghcr credential still breaks there, which is the failure this PR sets out to prevent. The temp config is still alive at that point (the EXIT trap has not fired yet), so add `--registry-config "$ANONYMOUS_REGISTRY_CONFIG"` to that command too.

shrinedogg (changes requested):
## Code Review Summary

**Verdict: Request changes** (1 warning, 1 note)

Verified locally on the PR branch: `bash -n` passes, helm 4.2 (the pinned version) supports `--registry-config`, only one EXIT trap is registered in bootstrap.sh so the temp registry config cleanup cannot clobber another trap, and the README and operations.md descriptions match the new profile behavior. CI is green (the konflate skip is expected for fork PRs).

### Warning
- **bootstrap.sh:175**: the FluxInstance chart pull (step 4) uses the same ghcr registry but does not get the anonymous `--registry-config`, so the cached-credentials failure this PR fixes can still occur on the aws path. See inline comment (anchored at line 123 because line 175 is outside this PR's diff hunk).

### Note
- **PR title and commit message**: "authenticate Flux chart pulls with GitHub PAT" describes the opposite of what the change does (it forces anonymous pulls and keeps the PAT only for Flux Git sync). Since merges land on main as squash commits titled by the PR, suggest retitling, for example "fix: pull Flux charts from GHCR anonymously".
- Minor: the .env.example GITHUB_TOKEN comment lost the fine-grained-vs-classic token guidance (now just "repository read access"). The detail survives in docs/operations.md, so optional.

### Looks good
- The empty-registry-config approach is the right fix for stale cached ghcr credentials, and the temp file is cleaned up via EXIT trap on both the success and failure paths
- Moving the Flux Operator install ahead of the mac exit makes the mac profile more useful while keeping it credential-free
- Docs (README, operations.md, .env.example headers) updated consistently with the new behavior

bootstrap.sh
lmilbaum: Thank you for flashing this out. Fixed.

shrinedogg (commented):
Reviewed and verified locally. The approach is sound: an empty `--registry-config` file forces anonymous chart pulls, so stale or over-scoped ghcr.io credentials in the user's helm/docker config can no longer break bootstrap, and PAT usage is now scoped to Flux git sync only, which matches the actual requirement.

Local verification (`KNR_OPS_PROFILE=mac ./bootstrap.sh`, helm v4.2.2, docker):

- flux-operator 0.58.0 chart pulled anonymously from GHCR and installed (pod Running)
- script exited 0 after step 2 with the documented message
- no `flux-github-pat` secret and no FluxInstance created, as documented
- kind cluster deleted afterwards; no leftover state

Notes before merge:

1. **Title mismatch.** The PR title says "authenticate Flux chart pulls with GitHub PAT", but the change does the opposite: charts are pulled anonymously and the PAT is only required for Flux git sync on the AWS profile. Suggest retitling, e.g. "fix: pull Flux charts anonymously, require PAT only for git sync".
2. **helm v4 dependency.** `--registry-config` on install/upgrade does not exist on helm v3 (checked v3.19.0). The mise-pinned toolchain (helm 4.2) covers this, but a contributor running system helm v3 gets an "unknown flag" failure. Optional hardening: a helm major-version check in the bootstrap.sh prereq loop.
3. **FYI, expected:** as a fork PR, the `konflate` render check skips by design (fork safety). This PR touches no Flux manifests, so nothing was lost; the validate workflow passed. Worth knowing that bootstrap.sh changes in general are covered by neither workflow, so local runs like the above are the only verification for this class of change.

Non-blocking nits: the EXIT trap for the temp registry config is safe (no other traps in the script to clobber), and the `mktemp` usage is portable across macOS and Linux.

shrinedogg (approved):
LGTM!

PR #17, opened 2026-08-14
feat: install FluxInstance for mac profile

lmilbaum:
## Summary
- install FluxInstance in the management kind cluster for both profiles
- keep GitHub/SOPS secrets and Git synchronization AWS-only
- retain anonymous OCI chart pulls for both Flux charts
- document the updated Mac profile behavior

## Validation
- bash -n bootstrap.sh
- git diff --check
- mocked Mac bootstrap with Flux Operator and FluxInstance

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,8 @@ mise run teardown           # full teardown (EKS, AWS resources, kind)
 The shared toolchain is defined in `mise.toml`. AWS-specific tools are layered
 through `mise.aws.toml`; use the `aws` profile when those tools are needed.
 The `mac` profile is still work in progress. It creates the management kind
-cluster and installs the Flux Operator, but does not configure GitOps sync or
-provision AWS resources:
+cluster, installs the Flux Operator and FluxInstance, but does not configure
+GitOps sync or provision AWS resources:
 
 ```sh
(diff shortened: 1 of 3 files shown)


shrinedogg (approved):
## Code Review Summary

**Verdict: Approve**

- `bash -n bootstrap.sh` passes.
- Rendered the FluxInstance chart (helm 4.2, anonymous registry config) with the exact mac-shaped values, no `instance.sync.*`: the chart renders a valid FluxInstance with no sync section, so the mac install shape is schema-valid and pulls anonymously.
- All AWS-only pieces are gated on the profile: the env checks at the top, the GitHub PAT and SOPS age secrets, the `instance.sync.*` settings, and the final message. Nothing AWS-only leaks into the mac path.
- Both chart pulls use the anonymous `--registry-config`, retaining the fix from PR #16.
- README and docs/operations.md describe the new behavior accurately; no stale references to the previous "steps 1 and 2" mac behavior remain.
- CI is green (the konflate skip is expected for fork PRs).

Nice touches: the step 4 header comment dropped "to start GitOps reconciliation" to match the mac behavior, and the args-array construction keeps the shared chart settings in one place.

Approving.

PR #18, opened 2026-08-14
refactor: rename mac profile to local-host

lmilbaum:
## Summary
- rename mise.mac.toml to mise.local.toml
- rename the profile identity from mac to local
- update bootstrap, teardown, README, and operations documentation
- keep macOS host-specific implementation comments unchanged

## Validation
- bash -n bootstrap.sh teardown.sh
- git diff --check
- mise -E local env -s zsh

### DIFF
--- a/mise.local.toml
+++ b/mise.local.toml
@@ -1,0 +1,2 @@
+[env]
+KNR_OPS_PROFILE = "local"
--- a/mise.local.toml
+++ b/mise.local.toml
@@ -1,2 +1,0 @@
-[env]
-KNR_OPS_PROFILE = "local"
--- a/README.md
+++ b/README.md
@@ -70,21 +70,21 @@ mise run teardown           # full teardown (EKS, AWS resources, kind)
 
 The shared toolchain is defined in `mise.toml`. AWS-specific tools are layered
 through `mise.aws.toml`; use the `aws` profile when those tools are needed.
-The `local` profile is still work in progress. It creates the management kind
+The `local-host` profile is still work in progress. It creates the management kind
 cluster, installs the Flux Operator and FluxInstance, but does not configure
 GitOps sync or provision AWS resources:
 
 ```sh
-mise -E local install
-mise -E local run bootstrap
-mise -E local run teardown
+mise -E local-host install
+mise -E local-host run bootstrap
+mise -E local-host run teardown
 ```
 
 The Flux charts are pulled anonymously. The AWS profile requires a
-GitHub PAT so Flux can clone this repository; the local profile does not require
+GitHub PAT so Flux can clone this repository; the local-host profile does not require
 GitHub or AWS credentials.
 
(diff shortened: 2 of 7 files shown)


mise.local.toml
```diff
@@ -0,0 +1,2 @@
+[env]
```
shrinedogg: Warning: mise.local.toml is a reserved filename in mise: it is auto-loaded as a local override on every invocation, with or without -E local. Verified locally with mise 2025.11.10: `mise config` lists mise.local.toml in the default load set, and a bare `mise env` exports KNR_OPS_PROFILE=local. Effect: `mise run bootstrap` and `mise run teardown` with no -E flag now take the local path, while the README quickstart and docs present the bare commands as the full AWS handoff. The aws profile still wins when explicitly selected, so the flip only hits the default path, which the PR's validation did not exercise. Suggest picking a profile name that is not a mise-reserved filename (anything other than "local").

shrinedogg (changes requested):
## Code Review Summary

**Verdict: Request changes** (1 warning)

Verified locally on the PR branch (mise 2025.11.10):

- The rename sweep is complete: every remaining "mac" reference in the repo is unrelated (SOPS MAC fields, MachinePool types) or an intentionally kept macOS-specific podman comment, matching the PR description. No references in .github either.
- `bash -n` passes on both scripts, `./teardown.sh local` takes the kind-only path, and the old `./teardown.sh mac` fails loudly with the unsupported-profile error.
- `mise -E local env` exports KNR_OPS_PROFILE=local and `mise -E aws env` exports aws.
- CI is green (the konflate skip is expected for fork PRs).

### Warning
- **mise.local.toml:1**: "local" collides with mise's reserved, auto-loaded override filename. A bare `mise env` (no -E flag) exports KNR_OPS_PROFILE=local, so the default `mise run bootstrap` and `mise run teardown` invocations silently flip from the documented AWS behavior to the local path. See inline comment. Suggest renaming the profile to a non-reserved name.

### Looks good (once the collision is resolved)
- Complete, consistent rename across scripts and docs
- The old profile name fails closed with a clear error in both scripts

shrinedogg:
Pushed commit 23e03e6 renaming the profile from local to local-host, addressing the review above.

Verified locally before pushing: mise.local-host.toml is not in mise's default load set, so a bare `mise env` no longer exports KNR_OPS_PROFILE and the default bootstrap/teardown keep their documented AWS behavior. `mise -E local-host env` exports local-host, `mise -E aws env` still exports aws, both old names (mac, local) fail closed with the unsupported-profile error, and `mise run validate` passes.

Also retitled the PR so the squash-merge commit message matches the final change.

mise.local.toml
shrinedogg: After discussion with @lmilbaum a decision was made to change the usage of "local" to "local-host" to prevent the mise issue.

PR #24, opened 2026-08-15
refactor: centralize bootstrap preflight checks

lmilbaum:
## Summary
- centralize profile, tool, container-engine, GitHub, and AWS age-key validation in `preflight_checks`
- verify the AWS repository and `main` branch before creating the kind cluster
- keep cluster creation and resource installation outside preflight

## Validation
- bash -n bootstrap.sh
- git diff --check
- invalid repository probe stops before cluster creation


bootstrap.sh
```diff
@@ -4,83 +4,105 @@
...
+    github_branch_status="$(curl -sS -o /dev/null -w '%{http_code}' \
+      -H 'Accept: application/vnd.github+json' \
+      -H "${GITHUB_AUTH}" \
+      "https://api.github.com/repos/${GITHUB_REPO}/branches/${github_branch_path}" || true)"
+    if [ "$github_branch_status" != 200 ]; then
+      echo "ERROR: GitHub repository or branch '${GIT_BRANCH}' is unavailable at '${GIT_REPO_URL}' (HTTP ${github_branch_status})" >&2
+      exit 1
+    fi
+
+    AGE_KEY_FILE="${AGE_KEY_FILE:-age.agekey}"
```
shrinedogg, line 49: Suggestion: this age-key validation now runs in preflight, but the same AGE_PUBKEY and AGE_CONTENT checks still run again in step 3b (lines 165-182). Since preflight executes in the same shell, the step-3b copy can be deleted, keeping only the secret creation. If you do, carry over the more specific error messages from step 3b (missing "# created:" header vs missing "AGE-SECRET-KEY-" line) instead of the merged generic one here.

shrinedogg (approved):
## Code Review Summary

**Verdict: Approve** (2 suggestions, neither blocking)

- `bash -n` passes; CI green (the konflate skip is expected for fork PRs).
- Failure-path probes, all exiting before any cluster mutation: unsupported profile, missing GITHUB_TOKEN, non-HTTPS GIT_REPO_URL (clear validation error), and a nonexistent repository (HTTP 404, the PR's stated validation claim).
- Positive-path probe the PR validation did not cover: the exact curl call against the real repo and branch with a valid token returns 200, so the new guard does not break the happy path.
- Cluster-name consistency: bootstrap creates `mgmt` and teardown deletes `mgmt`.
- Fail-fast ordering holds: every validation now runs before kind cluster creation, and the age-key pubkey and content validation moved from post-creation into preflight, closing the residual gap left from #15.

### Suggestions
- **bootstrap.sh:48**: the age-key validation is now duplicated in preflight and step 3b. See inline comment.
- The move dropped several explanatory comments this repo usually keeps (the podman-docker shim rationale, the ENGINE_SOCK mount rationale in the docker case, the GITHUB_USER note, the SOPS age-key context). Worth restoring the "why" while moving the code.
- Note: the four editor-config commits (vscode/spell-checker) net to zero in the diff and vanish on squash merge, so no history concern.

### Looks good
- The repo/branch existence check fails closed with a clear error, including the HTTP 000 network-error case
- URL normalization handles trailing slashes and .git suffixes; non-HTTPS URLs are rejected with a clear message
- preflight_checks runs in the current shell, so the variables it sets (CONTAINER_ENGINE, ENGINE_SOCK, GITHUB_USER, AGE_PUBKEY) persist correctly for the install steps

bootstrap.sh
lmilbaum: Thank you for the feedback. Fixed.

bootstrap.sh
```diff
@@ -4,83 +4,114 @@
...
+    if [ ! -f "$AGE_KEY_FILE" ]; then
+      echo "ERROR: age key file not found at '$AGE_KEY_FILE'." >&2
+      echo "       Generate one with:  mise run sops-keygen" >&2
+      echo "       and add its PUBLIC key to .sops.yaml. See docs/secrets.md." >&2
+      exit 1
+    fi
+    AGE_PUBKEY="${AGE_PUBLIC_KEY:-$(grep '^# public key:' "${AGE_KEY_FILE}" 2>/dev/null | sed 's/^# public key: //')}"
```
shrinedogg, line 56: Suggestion (pre-existing construct, surfaced while re-testing): under set -euo pipefail, when the age key file has no "# public key:" line and AGE_PUBLIC_KEY is unset, grep exits 1 and pipefail makes this assignment fail, so the script dies silently here with exit 1 and no message. Verified locally: a garbage age key file produces exit 1 with zero output, and the "Cannot determine age public key" error below never fires. Appending `|| true` inside the command substitution lets the intended error path run. Only triggers when AGE_PUBLIC_KEY is unset, which is why .env users never see it.

shrinedogg (approved):
## Re-review Summary

**Verdict: Approve** (1 new minor suggestion, pre-existing and non-blocking)

Both items from the previous review are fixed in `0ea750c`:

- **Step-3b duplicate age-key validation**: removed; exactly one AGE_PUBKEY assignment remains (in preflight), and the secret creation kept a pointer comment.
- **Specific age-key error messages**: restored. Verified with probes: a missing key file prints the sops-keygen guidance, and a file with a header but no AGE-SECRET-KEY line prints "Expected a line starting with 'AGE-SECRET-KEY-'".
- **Explanatory comments**: the GITHUB_USER note, podman shim detection note, and engine-socket mount rationale are all restored (the mount note moved next to kind create, a better home for it).

Regression probes on the new head all pass and exit before any cluster mutation: unsupported profile, missing GITHUB_TOKEN, non-HTTPS GIT_REPO_URL, nonexistent repo (HTTP 404), and the positive API path (200). `bash -n` clean, CI green (the konflate skip is expected for fork PRs).

### Suggestion
- **bootstrap.sh:56**: the AGE_PUBKEY assignment can exit silently for a malformed age key file when AGE_PUBLIC_KEY is unset (pre-existing construct, surfaced while probing this round). See inline comment.

Approving.

PR #25, opened 2026-08-15
ci: authenticate curl request with GitHub token to avoid rate limiting

lmilbaum:
Fixes the rate limiting issue in the validate workflow when downloading kustomize.

The kustomize installation script download was being rate-limited by GitHub's API. Adding the default `github.token` to the curl request avoids unauthenticated rate limits and ensures the workflow completes reliably.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -15,7 +15,7 @@ jobs:
 
       - name: Install kustomize
         run: |
-          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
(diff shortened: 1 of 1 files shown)


shrinedogg:
Pushed commit 773b11c addressing the review finding.

The throttled request was not the script download from raw.githubusercontent.com but the release-detection call inside install_kustomize.sh, which hits api.github.com/repos/kubernetes-sigs/kustomize/releases unauthenticated unless GITHUB_TOKEN is set in the environment (script lines 142-146; it sends its own Authorization header when the variable is present). The header on the outer curl never reaches that inner call.

The commit exposes `GITHUB_TOKEN: ${{ github.token }}` as a step env var, so the installer authenticates its own API call. Verified against the live API: x-ratelimit-limit is 60 unauthenticated vs 5000 authenticated. The header on the script download is kept (harmless, and covers the download itself if raw throttling ever kicks in).

PR #26, opened 2026-08-15
bootstrap: validate age key file format before extracting public key

lmilbaum:
Reorder validation to check file format (# created:, # public key:, AGE-SECRET-KEY-)
before attempting to extract AGE_PUBKEY.

This prevents silent grep failures when the file is malformed and clarifies the
error messages. Addresses feedback on bootstrap preflight validation.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -53,21 +53,24 @@ preflight_checks() {
       echo "       and add its PUBLIC key to .sops.yaml. See docs/secrets.md." >&2
       exit 1
     fi
-    AGE_PUBKEY="${AGE_PUBLIC_KEY:-$(grep '^# public key:' "${AGE_KEY_FILE}" 2>/dev/null | sed 's/^# public key: //')}"
+    # Validate age key file format first (before attempting to extract the public key).
+    # This avoids silent grep failure if the file is malformed.
+    AGE_CONTENT=$(cat "${AGE_KEY_FILE}")
+    missing_fields=""
+    echo "$AGE_CONTENT" | grep -q '^# created:' 2>/dev/null || missing_fields="${missing_fields}# created: header, "
+    echo "$AGE_CONTENT" | grep -q '^# public key:' 2>/dev/null || missing_fields="${missing_fields}# public key: comment, "
+    echo "$AGE_CONTENT" | grep -q '^AGE-SECRET-KEY-' 2>/dev/null || missing_fields="${missing_fields}AGE-SECRET-KEY- line, "
+    
+    if [ -n "$missing_fields" ]; then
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -53,21 +53,24 @@ preflight_checks() {
...
+    # This avoids silent grep failure if the file is malformed.
     AGE_CONTENT=$(cat "${AGE_KEY_FILE}")
-    if ! echo "$AGE_CONTENT" | grep -q '^# created:' 2>/dev/null; then
-      echo "ERROR: '${AGE_KEY_FILE}' does not appear to be a valid age key file." >&2
-      echo "       Expected a file with '# created:' comment header" >&2
+    missing_fields=""
+    echo "$AGE_CONTENT" | grep -q '^# created:' 2>/dev/null || missing_fields="${missing_fields}# created: header, "
+    echo "$AGE_CONTENT" | grep -q '^# public key:' 2>/dev/null || missing_fields="${missing_fields}# public key: comment, "
+    echo "$AGE_CONTENT" | grep -q '^AGE-SECRET-KEY-' 2>/dev/null || missing_fields="${missing_fields}AGE-SECRET-KEY- line, "
+
```
shrinedogg, line 63: Nit: this blank line carries trailing spaces, so `git diff --check` (one of the PR's validation commands) flags it. One-line cleanup.

shrinedogg (approved):
## Code Review Summary

**Verdict: Approve** (1 nit)

Verified locally on the PR branch with the probe matrix from the previous review:

- Garbage key file (the case that exited silently before): now fails with "Missing: # created: header, # public key: comment, AGE-SECRET-KEY- line" and exit 1. The silent path is closed.
- Header but no secret key: "Missing: AGE-SECRET-KEY- line".
- Empty "# public key:" value: the friendly "Cannot determine age public key" error fires instead of a set -e trip.
- AGE_PUBLIC_KEY set plus a malformed file: the file is still format-validated first.
- Valid key file: preflight passes and extraction succeeds (verified with a bogus CONTAINER_ENGINE so the script stops right after preflight; no cluster was created).
- Regression probes unchanged: bad profile, missing GITHUB_TOKEN, non-HTTPS URL, nonexistent repo (404), positive API path (200).
- `bash -n` clean; CI green (the konflate skip is expected for fork PRs).

### Nit
- **bootstrap.sh:63**: trailing whitespace on the blank line trips `git diff --check`, one of the PR's own validation commands. See inline comment.

### Looks good
- Validating the format before extraction is a cleaner fix than guarding the grep itself: the extraction can no longer fail under set -e, and the aggregated "Missing:" message names everything to regenerate in one pass
- Keeping the AGE_PUBKEY empty check as a fallback covers the empty-value edge case

PR #21, opened 2026-08-14
refactor: nest management resources under aws

lmilbaum:
## Summary
- nest the management GitOps tree under mgmt/aws
- keep workload manifests under workload/
- update Flux sync, SOPS, validation, teardown, and documentation paths

## Validation
- mise run validate
- git diff --check

PR #27, opened 2026-08-15
feat: add empty mgmt/docker kustomization placeholder

lmilbaum:
Add minimal kustomization.yaml for local-host profile sync without provisioning resources.

This allows Flux to sync without errors when running on docker/podman backend,
preparing for the local-host GitOps sync feature.


mgmt/local-host/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
shrinedogg, line 1: Question: today the profile maps to an identically named tree (aws profile syncs mgmt/aws). This placeholder points the local-host profile at mgmt/docker. Is the divergence deliberate, with the directory naming the container backend rather than the profile? If this tree will only ever serve the local-host profile, mgmt/local-host would keep the mapping obvious, and the comment notes podman is covered too, which the name docker does not suggest. Non-blocking, just cheaper to decide before the sync PR wires it up.

shrinedogg (approved):
## Code Review Summary

**Verdict: Approve** (1 non-blocking naming question)

Verified locally:

- `kubectl kustomize mgmt/docker` builds cleanly (empty output, valid empty kustomization).
- `mise run validate` passes and now covers 28 overlays including this one.
- CI green: the kustomize build and yamllint jobs both pick up the new file (the workflow covers mgmt since #21). The konflate skip is expected for fork PRs.
- No interaction with SOPS rules or existing Flux paths; nothing references mgmt/docker yet, the sync wiring is the follow-up.

### Question
- **mgmt/docker/kustomization.yaml:1**: directory name vs profile name mapping (aws maps to mgmt/aws today; local-host would map to mgmt/docker here). See inline comment.

### Looks good
- Correct empty-kustomization shape (apiVersion/kind plus resources: []), which Flux reconciles to a no-op
- Landing the placeholder ahead of the sync PR keeps that diff focused

shrinedogg:
Pushed commit a76228b renaming the placeholder tree to mgmt/local-host, resolving the naming question from review: the profile-to-tree mapping now stays symmetric (aws -> mgmt/aws, local-host -> mgmt/local-host), and the name covers podman as the file comment already states.

Verified before pushing: kubectl kustomize mgmt/local-host builds clean, mise run validate covers all 28 overlays with no failures, no dangling mgmt/docker references, git diff --check clean.

PR #28, opened 2026-08-15
feat: bootstrap local container registry for local-host profile

lmilbaum:
When running with PROFILE=local-host, bootstrap a Docker Registry container (registry:2)
to track OCI artifacts for local development.

## Features
- Runs in the kind network for in-cluster access
- Exposes on localhost:5000 (configurable via REGISTRY_PORT env var)
- Idempotent: checks if already running, restarts if stopped
- Creates a configmap with registry URL for cluster discovery

This enables local OCI artifact management for development and testing without
requiring external registries.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -140,6 +140,39 @@ echo ">>> Waiting for cluster node to be ready..."
 kubectl config use-context kind-mgmt
 kubectl wait --for=condition=Ready node --all --timeout=120s
 
+# ── Step 1.5: Bootstrap local container registry (local-host profile only) ────
+if [ "$PROFILE" = local-host ]; then
+  echo ">>> Bootstrapping local container registry..."
+  REGISTRY_NAME="knr-registry"
+  REGISTRY_PORT="${REGISTRY_PORT:-5000}"
+  
+  # Check if registry already exists
+  if ! $CONTAINER_ENGINE ps -a --filter "name=^${REGISTRY_NAME}$" | grep -q "${REGISTRY_NAME}"; then
+    # Create registry container
+    $CONTAINER_ENGINE run -d \
+      --name "$REGISTRY_NAME" \
+      --network kind \
+      -p "127.0.0.1:${REGISTRY_PORT}:5000" \
+      registry:2
+    echo "    Registry started: localhost:${REGISTRY_PORT}"
+  else
+    # Check if registry is running, restart if needed
+    if ! $CONTAINER_ENGINE ps --filter "name=^${REGISTRY_NAME}$" | grep -q "${REGISTRY_NAME}"; then
+      echo "    Restarting stopped registry..."
+      $CONTAINER_ENGINE start "$REGISTRY_NAME"
+    else
+      echo "    Registry already running: localhost:${REGISTRY_PORT}"
+    fi
+  fi
+  
+  # Configure kind nodes to access the registry via the hostname
+  # Add a configmap to tell the cluster about the local registry
+  kubectl create configmap local-registry-config \
+    --from-literal=registry-url="${REGISTRY_NAME}:5000" \
+    --namespace kube-system \
+    --dry-run=client -o yaml | kubectl apply -f -
+fi
+
 # ── Step 2: Install the Flux Operator ────────────────────────────────────────
 echo ">>> Installing Flux Operator..."
 ANONYMOUS_REGISTRY_CONFIG="$(mktemp)"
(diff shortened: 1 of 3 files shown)


mgmt/docker/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
shrinedogg, line 1: Warning: this branch is based before #27 merged, and #27 landed on main with the tree renamed to mgmt/local-host. Merging this PR as-is would add mgmt/docker back alongside it (docs/operations.md:109 in this PR also points Flux at mgmt/docker). Please rebase onto main so the placeholder keeps the mgmt/local-host name.

bootstrap.sh
```diff
@@ -140,6 +140,39 @@ echo ">>> Waiting for cluster node to be ready..."
...
+  REGISTRY_NAME="knr-registry"
+  REGISTRY_PORT="${REGISTRY_PORT:-5000}"
+  
+  # Check if registry already exists
+  if ! $CONTAINER_ENGINE ps -a --filter "name=^${REGISTRY_NAME}$" | grep -q "${REGISTRY_NAME}"; then
+    # Create registry container
+    $CONTAINER_ENGINE run -d \
+      --name "$REGISTRY_NAME" \
+      --network kind \
+      -p "127.0.0.1:${REGISTRY_PORT}:5000" \
```
shrinedogg, line 155: Warning: the default REGISTRY_PORT=5000 collides with macOS AirPlay Receiver. Verified on a stock Mac: ControlCenter listens on *:5000 (IPv4 and IPv6); the registry container starts, but localhost:5000 resolves to ::1 and answers 403 from AirPlay, while 127.0.0.1:5000/v2/ correctly returns 200 from the registry. The documented docker push localhost:5000/... workflow therefore hits AirPlay, not the registry. Suggest defaulting to 5001 (the kind local-registry guide convention) or documenting that AirPlay Receiver must be disabled.

shrinedogg (changes requested):
## Code Review Summary

**Verdict: Request changes** (3 warnings, 1 suggestion, 1 nit)

Verified locally (Docker Desktop on macOS, AirPlay Receiver enabled):

- `bash -n` passes; CI green (the konflate skip is expected for fork PRs).
- Live registry simulation (registry:2 container on a kind network): create, idempotency filters, and the restart path all behave as described.
- Port collision probe: with the default REGISTRY_PORT=5000, `127.0.0.1:5000/v2/` returns 200 from the registry but `localhost:5000/v2/` returns 403 from macOS AirPlay Receiver, which listens on *:5000 including IPv6. The documented `docker push localhost:5000/...` workflow therefore targets AirPlay, not the registry, on a stock Mac. See inline comment.
- `git diff --check` fails on two trailing-whitespace lines (bootstrap.sh:148, 167).
- Base check: this branch predates #27's merge, which renamed the placeholder tree to mgmt/local-host. The PR still adds mgmt/docker/kustomization.yaml and the docs reference it; merging would resurrect the old name alongside the new one.

### Warnings
- **Stale base / mgmt/docker resurrection**: rebase onto main so the placeholder stays mgmt/local-host (inline comment on the file).
- **bootstrap.sh:155**: default port 5000 vs AirPlay Receiver (inline comment, with probe evidence). Suggest defaulting to 5001, the kind local-registry convention, or documenting the AirPlay conflict.
- **teardown.sh** (not in this diff): the local-host teardown path deletes only the kind cluster and leaves the knr-registry container running and published, orphaned from the cluster it served. Either remove the container in the local-host teardown path or document that it persists.

### Suggestion
- In-cluster pulls: Flux OCIRepository sources can set plainHTTP, but the docs workflow shows pushing a container image, and kubelet pulls from knr-registry:5000 over HTTP will fail without a containerdConfigPatches entry in the kind config. Add the patch for the local-host profile or scope the docs to OCI artifacts.

### Nit
- bootstrap.sh:148 and :167 trailing whitespace (git diff --check).

### Looks good
- Step ordering is right: the registry block runs after the kind network exists
- The idempotency logic (create, restart-if-stopped, already-running) verified in a live docker simulation
- The AWS-only prerequisite regrouping in docs/operations.md reads better than before

mgmt/docker/kustomization.yaml
lmilbaum: Rebased

bootstrap.sh
lmilbaum: Fixed

PR #30, opened 2026-08-16
feat: bootstrap local container registry for local-host profile

lmilbaum:
## Summary

- bootstrap an idempotent local OCI registry for the `local-host` profile
- publish registry connection details in the management cluster
- remove the registry container during local-host teardown
- document the local OCI artifact workflow and prerequisites

## Testing

- Not run (not recorded on the branch)

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -5,6 +5,8 @@ set -euo pipefail
 
 PROFILE="${KNR_OPS_PROFILE:-${1:-aws}}"
 GIT_BRANCH="main"
+REGISTRY_NAME="knr-registry"
+REGISTRY_PORT="${REGISTRY_PORT:-5001}"
 
 preflight_checks() {
   case "$PROFILE" in
@@ -60,7 +62,7 @@ preflight_checks() {
     echo "$AGE_CONTENT" | grep -q '^# created:' 2>/dev/null || missing_fields="${missing_fields}# created: header, "
     echo "$AGE_CONTENT" | grep -q '^# public key:' 2>/dev/null || missing_fields="${missing_fields}# public key: comment, "
     echo "$AGE_CONTENT" | grep -q '^AGE-SECRET-KEY-' 2>/dev/null || missing_fields="${missing_fields}AGE-SECRET-KEY- line, "
-    
+
     if [ -n "$missing_fields" ]; then
       echo "ERROR: '${AGE_KEY_FILE}' is not a valid age key file." >&2
       echo "       Missing: ${missing_fields%, }" >&2
@@ -128,9 +130,17 @@ fi
 # socket path. This ensures all in-cluster components can access a Docker-compatible API
 # at /var/run/docker.sock, whether the backend is Docker or Podman (which exposes a
 # Docker-compatible socket). This is essential for building and loading container images.
+KIND_REGISTRY_PATCH=""
+if [ "$PROFILE" = local-host ]; then
+  KIND_REGISTRY_PATCH="containerdConfigPatches:
+  - |-
+    [plugins.\"io.containerd.grpc.v1.cri\".registry.mirrors.\"localhost:${REGISTRY_PORT}\"]
+      endpoint = [\"http://${REGISTRY_NAME}:5000\"]"
+fi
 kind create cluster --name mgmt --config - <<EOF
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
+${KIND_REGISTRY_PATCH}
 nodes:
   - role: control-plane
     extraMounts:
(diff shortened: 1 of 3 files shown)


bootstrap.sh
```diff
@@ -143,6 +143,38 @@ echo ">>> Waiting for cluster node to be ready..."
...
+      $CONTAINER_ENGINE start "$REGISTRY_NAME"
+    else
+      echo "    Registry already running: localhost:${REGISTRY_PORT}"
+    fi
+  fi
+
+  # Configure kind nodes to access the registry via the hostname
+  # Add a configmap to tell the cluster about the local registry
+  kubectl create configmap local-registry-config \
+    --from-literal=registry-url="${REGISTRY_NAME}:5001" \
```
shrinedogg: Warning: this URL is wrong for in-cluster consumers. The registry container listens on 5000 inside the kind network; 5001 is only the host-side published port. Verified live from a container on the kind network: knr-registry:5000/v2/ returns 200 while knr-registry:5001 is unreachable. This should be ${REGISTRY_NAME}:5000 (or localhost:${REGISTRY_PORT} if the configmap is meant for host-side consumers; as written it also ignores a REGISTRY_PORT override either way).

docs/operations.md
```diff
@@ -84,8 +87,32 @@ from Git with no further manual steps.
 
 The local-host profile performs the cluster, Flux Operator, and FluxInstance steps in
 the `mgmt` management cluster, but does not create GitHub or SOPS secrets,
-configure Git sync, or start GitOps reconciliation. The AWS profile adds the
-GitHub/SOPS secrets and configures the FluxInstance to sync `mgmt/aws/`.
+configure Git sync, or start GitOps reconciliation. Instead, it bootstraps a
+local Docker Registry container (`registry:2`) running on the host machine
+(accessible at `localhost:5000`) for managing OCI artifacts.
```
shrinedogg: Warning: this line and line 98 still say localhost:5000 and "defaults to 5000", but the code now defaults REGISTRY_PORT to 5001 and the workflow example below correctly uses 5001. Please update the prose to match.

shrinedogg (changes requested):
## Code Review Summary

**Verdict: Request changes** (2 warnings, 2 suggestions)

All three warnings from the #28 review are addressed, and each fix was verified live:

- Base: the branch is based on current main, so the placeholder stays mgmt/local-host and the docs point there.
- Port: the code default is now 5001, sidestepping the macOS AirPlay collision on 5000 (verified in the #28 round: localhost:5000 answers 403 from AirPlay on a stock Mac).
- Teardown: the local-host path now removes knr-registry. Verified live: with a registry container present and no cluster, teardown removes it and exits 0.
- The trailing-whitespace nit is fixed; git diff --check is clean. bash -n passes on both scripts; CI green (the konflate skip is expected for fork PRs).

Two new issues in this iteration:

### Warnings
- **bootstrap.sh:174**: the configmap publishes knr-registry:5001, a port that is unreachable inside the cluster (5001 is the host-published port; the container listens on 5000 on the kind network). Verified live from a container on the kind network: :5000/v2/ returns 200, :5001 is unreachable. See inline comment.
- **docs/operations.md:92,98**: prose still says localhost:5000 and defaults to 5000 while the code defaults to 5001. See inline comment.

### Suggestions
- docs/operations.md workflow step 3 points readers to containerdConfigPatches "see bootstrap.sh Step 1.5", but Step 1.5 does not configure any; either add the patch to the kind config for local-host or correct the pointer.
- The teardown registry cleanup ignores an explicit CONTAINER_ENGINE override (bootstrap honors it) and only checks for the CLI, not a running daemon. A podman-created registry would be missed on a machine where docker also exists.

### Looks good
- The port default change, the mgmt/local-host reference, and the teardown cleanup all land correctly
- Idempotent registry block unchanged from the previously verified version

lmilbaum:
Addressed the review in 0e344c1:

- publish the registry's reachable in-cluster endpoint as `knr-registry:5000`
- align the documented host endpoint and default with `localhost:5001`
- configure kind containerd to mirror the host endpoint to the registry on the kind network
- honor and validate `CONTAINER_ENGINE` during teardown, including daemon availability

Validation: `git diff --check`, `bash -n bootstrap.sh teardown.sh`, and `mise run validate` all pass.

shrinedogg (approved):
## Re-review Summary

**Verdict: Approve**

All items from the previous review are fixed in `0e344c1`, and the follow-on teardown refactor in `2d98374` checks out:

- **Configmap dead port (knr-registry:5001)**: fixed, now `${REGISTRY_NAME}:5000`, the in-cluster container port (matches my earlier probe: :5000 answers 200 on the kind network, :5001 is host-only).
- **Docs 5000 vs 5001**: fixed, prose now reads localhost:5001 by default and defaults to 5001.
- **containerdConfigPatches**: now real config, not just a pointer. The kind Cluster config injects a containerd mirror for local-host (localhost:${REGISTRY_PORT} to http://knr-registry:5000), the standard kind local-registry pattern. Rendered the exact heredoc and YAML-parsed it for both profiles.
- **teardown CONTAINER_ENGINE handling**: explicit engine is validated (a bogus value now errors), daemon liveness is checked, and an unavailable engine degrades to a warning plus skipped registry cleanup.

Additional verification on the new head: teardown probes pass (no cluster or registry exits 0; a present registry is removed with exit 0; AWS_ONLY=1 with local-host still hits the guard). The teardown preflight refactor correctly avoids the not-yet-defined info/warn helpers, and the added xargs prerequisite is genuinely used at line 696. `bash -n` clean on both scripts, `git diff --check` clean, CI green (the konflate skip is expected for fork PRs).

Approving.

PR #31, opened 2026-08-16
feat: publish local-host configuration as OCI artifact

lmilbaum:
## Summary

- add a local-host-only `oci-push` mise task that publishes `mgmt/docker` to the local registry
- preserve the `mgmt/docker/` directory structure in the pulled artifact
- publish an initial `knr-ops:latest` artifact during local-host bootstrap
- wait for the registry V2 API with curl retries and report registry and artifact endpoints
- document bootstrap, republish, and teardown behavior

## Validation

- `git diff --check`
- `bash -n bootstrap.sh teardown.sh`
- `mise -E local-host tasks info oci-push`
- `mise run validate`
- verified the generated archive contains `mgmt/docker/kustomization.yaml`

## Not run

- end-to-end registry push (the local registry was not running)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,13 +70,15 @@ mise run teardown           # full teardown (EKS, AWS resources, kind)
 
 The shared toolchain is defined in `mise.toml`. AWS-specific tools are layered
 through `mise.aws.toml`; use the `aws` profile when those tools are needed.
-The `local-host` profile is still work in progress. It creates the management kind
-cluster, installs the Flux Operator and FluxInstance, but does not configure
-GitOps sync or provision AWS resources:
+The `local-host` profile is still work in progress. It creates the management
+kind cluster, a local OCI registry, and the Flux Operator and FluxInstance. It
+also publishes the `mgmt/local-host/` folder as the `knr-ops:latest` OCI artifact,
+but does not configure GitOps sync or provision AWS resources:
 
 ```sh
 mise -E local-host install
 mise -E local-host run bootstrap
+mise -E local-host run oci-push  # republish mgmt/local-host after local changes
 mise -E local-host run teardown
 ```
 
@@ -84,8 +86,9 @@ The Flux charts are pulled anonymously. The AWS profile requires a
 GitHub PAT so Flux can clone this repository; the local-host profile does not require
 GitHub or AWS credentials.
 
-The local-host teardown deletes only the `mgmt` kind cluster. The default
-teardown path suspends Flux and removes the AWS-managed infrastructure.
+The local-host teardown deletes the `mgmt` kind cluster and local registry
+container. The default teardown path suspends Flux and removes the AWS-managed
+infrastructure.
 
 ## Documentation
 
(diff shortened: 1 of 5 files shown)


mgmt/docker/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
shrinedogg: Warning: this creates a second local-host placeholder tree while mgmt/local-host (added in #27 for the same role) is still on disk but now referenced nowhere: grep finds zero references to mgmt/local-host in the PR tree, and the README figure lists mgmt/docker but not mgmt/local-host. If the OCI artifact root replaces the git-sync placeholder, delete mgmt/local-host in this PR so there is one local-host tree; if the #27 name should win, point OCI_SOURCE_PATH at it instead. Either way, one directory should win.

shrinedogg (changes requested):
## Code Review Summary

**Verdict: Request changes** (1 warning; the feature itself verifies end-to-end)

Verified locally, including the end-to-end path the PR lists as not run:

- Started a registry:2 container on 127.0.0.1:5001 and ran `mise -E local-host run oci-push`: the artifact pushed successfully.
- Registry state confirmed via the V2 API (`knr-ops` has tag `latest`).
- Pulled the artifact back with `flux pull artifact`: it contains exactly `mgmt/docker/kustomization.yaml`, so the directory structure is preserved as described.
- `mise run validate` passes over 29 overlays; `bash -n` and `git diff --check` clean; CI green (the konflate skip is expected for fork PRs).
- The bootstrap additions read correctly: the mise prereq for local-host, the registry readiness wait with connrefused retries against the /v2/ endpoint, and the endpoint reporting.

### Warning
- **Two local-host placeholder trees**: this PR adds mgmt/docker as the OCI artifact root, but mgmt/local-host (from #27, for the same role) is still on disk and now referenced nowhere (grep finds zero references in the PR tree; the README figure lists mgmt/docker but not mgmt/local-host). See inline comment. One directory should win; deleting the orphan is a one-line `git rm`.

### Looks good
- The narrow source scope (only mgmt/docker is packaged) keeps .env and age.agekey out of the artifact, a thoughtful default
- The readiness wait and the /v2/ probe match how the registry actually behaves (verified in this round's simulation)
- Temp-dir cleanup via trap in the task, reproducible artifact flags, and honest "Not run" testing notes

mgmt/docker/kustomization.yaml
lmilbaum: Fixed

shrinedogg (approved):
## Re-review Summary

**Verdict: Approve**

The placeholder-tree warning from the previous review is resolved in `dccd4a6`, via the cleaner of the two options: mgmt/docker is deleted, OCI_SOURCE_PATH now points at mgmt/local-host, and the #27 naming consistency holds (aws maps to mgmt/aws, local-host maps to mgmt/local-host). Grep finds zero mgmt/docker references in the tree.

Re-verified end-to-end on the new head: started a local registry, ran `mise -E local-host run oci-push` (push succeeded), and pulled the artifact back; it contains exactly `mgmt/local-host/kustomization.yaml`. `mise run validate` passes over 28 overlays, `bash -n` and `git diff --check` are clean, and CI is green (the konflate skip is expected for fork PRs).

Approving.

PR #33, opened 2026-08-16
feat: sync local-host from OCI registry

lmilbaum:
## Summary

- configure the local-host FluxInstance to sync from an OCIRepository
- consume `knr-registry:5000/knr-ops:latest` at `mgmt/local-host`
- mark the generated OCIRepository as insecure for the local HTTP registry
- update bootstrap completion output and documentation for active OCI reconciliation

## Validation

- `git diff --check`
- `bash -n bootstrap.sh teardown.sh`
- `mise run validate`
- rendered the official FluxInstance Helm chart and verified the generated sync URL, tag, path, and `spec.insecure: true` patch

## Not run

- end-to-end bootstrap against a running local registry

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,8 @@ through `mise.aws.toml`; use the `aws` profile when those tools are needed.
 The `local-host` profile is still work in progress. It creates the management
 kind cluster, a local OCI registry, and the Flux Operator and FluxInstance. It
 also publishes the `mgmt/local-host/` folder as the `knr-ops:latest` OCI artifact,
-but does not configure GitOps sync or provision AWS resources:
+then configures Flux to reconcile from that artifact. It does not provision AWS
+resources:
 
 ```sh
 mise -E local-host install
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -260,6 +260,14 @@ if [ "$PROFILE" = aws ]; then
     --set instance.sync.path=mgmt/aws
     --set instance.sync.pullSecret=flux-github-pat
   )
+else
+  FLUX_INSTANCE_ARGS+=(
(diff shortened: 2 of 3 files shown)


shrinedogg (approved):
## Code Review Summary

**Verdict: Approve**

- Rendered the flux-instance chart (helm 4.2, anonymous registry config) with the exact local-host values: the FluxInstance gets `sync.kind: OCIRepository`, `url: oci://knr-registry:5000/knr-ops`, `ref: latest`, `path: mgmt/local-host`, and the insecure patch lands under `spec.kustomize.patches` as a clean YAML block scalar targeting kind: OCIRepository.
- Confirmed against the upstream flux-operator docs that this is the documented pattern: the operator generates the OCIRepository from spec.sync, and `.spec.kustomize.patches` with a kind target is applied to that generated source (the sync guide uses the same target form for OCIRepository fields). `spec.insecure` is the right field for a plain-HTTP registry.
- Ordering is sound: the initial artifact is published in step 1.5, before the FluxInstance and its source exist (step 4), and the 1m default sync interval picks up subsequent `oci-push` republishes of :latest.
- The else branch is exactly local-host (the profile case statement admits only local-host and aws), and OCI_REPOSITORY/OCI_TAG defaults match the oci-push task, so overrides stay consistent across publish and sync.
- Docs (README, operations.md) updated consistently; `flux get sources oci --watch` is a valid command.
- `bash -n` and `git diff --check` clean; `mise run validate` passes over 28 overlays; CI green (the konflate skip is expected for fork PRs).

Not exercised here or in the PR's testing: a live end-to-end bootstrap. Everything renderable is verified and the runtime behavior is upstream-documented, so the residual risk is low.

Approving.

PR #35, opened 2026-08-17
feat: install Flux on local workload cluster

lmilbaum:
## Summary

- install the CAPI Add-on Helm provider for the local-host profile
- deliver Kindnet, Flux Operator, and a FluxInstance to the CAPD workload cluster
- publish and reconcile the `workload/local-host` path from the local OCI registry
- update bootstrap progress and local-host documentation

## Testing

- `mise run validate`
- `bash -n bootstrap.sh teardown.sh`
- `git diff --check`
- completed a clean `mise -E local-host run bootstrap`
- verified both workload nodes became Ready
- verified workload FluxInstance, OCIRepository, and Kustomization became Ready
- completed `mise -E local-host run teardown`

PR #36, opened 2026-08-18
feat: deploy Podinfo to local workload cluster

lmilbaum:
## Summary

- deploy Podinfo to the local CAPD workload cluster through Flux using its OCI Helm chart
- add a mise task and documentation for forwarding Podinfo to http://localhost:9898
- align the local bootstrap reconciliation watcher with the readiness timeout

## Testing

- mise run validate

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,19 +77,27 @@ registry, and the Flux Operator and FluxInstance. It publishes the
 `mgmt/local-host/` and `workload/local-host/` folders as the `knr-ops:latest`
 OCI artifact. Flux installs CAPI with its Docker infrastructure provider (CAPD),
 provisions a local one-control-plane/one-worker workload cluster, and installs a
-separate Flux instance there. It does not provision AWS resources:
+separate Flux instance there. That workload Flux instance reconciles Podinfo,
+providing an end-to-end local path from management-cluster bootstrap through
+workload delivery and application access. This covers the complete GitOps and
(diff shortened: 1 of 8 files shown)


shrinedogg (approved):
## Code Review Summary

**Verdict: Approve**

- `mise run validate` passes over 38 overlays; `bash -n` on both scripts and `git diff --check` clean; CI green (the konflate skip is expected for fork PRs).
- The podinfo OCI chart tag 6.14.0 exists on ghcr.io (queried the registry API; the default tag listing caps at 100 entries, and the full listing confirms 6.14.0, with 6.14.1 the newest). The HelmRelease uses the same chartRef pattern as the rest of the repo, and the chart is public, so no credentials or insecure flag are needed.
- The workload bootstrap block is careful with state: kubeconfig in a mktemp file (0600), the log stream and file are cleaned by an EXIT trap on both success and failure paths, and the port discovery validates the parsed port.
- The watch-alignment fix is real: the progress watcher now shares LOCAL_RECONCILE_TIMEOUT instead of the flux watch default, so the display cannot time out ahead of the authoritative wait.
- Waiting on the workload flux-system Kustomization closes the gap noted in the #35 review: bootstrap now returns only once the workload-side Flux root is Ready, not just when the addons are delivered.
- Docs (README, operations.md) match the new behavior, and the port-forward task targets the correct podinfo service port (9898).

Approving.

PR #37, opened 2026-08-18
fix: wait explicitly for FluxInstance readiness

lmilbaum:
## Summary

- use Helm 4 legacy resource waiting to avoid treating the FluxInstance transient InProgress condition as a terminal failure
- wait explicitly for the FluxInstance Ready condition after chart installation

## Context

Follow-up to #36; the fix was added to its source branch after that PR had already merged.

## Testing

- bash -n bootstrap.sh
- git diff --check
- mise run validate

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -271,12 +271,21 @@ else
     --set-json 'instance.kustomize.patches=[{"patch":"- op: add\n  path: /spec/insecure\n  value: true","target":{"kind":"OCIRepository"}}]'
   )
 fi
+# Helm 4's watcher strategy treats the FluxInstance's transient InProgress
+# condition as a terminal failure. Use the legacy chart-resource wait here,
+# then wait explicitly for the operator-owned Ready condition below.
 helm upgrade --install flux \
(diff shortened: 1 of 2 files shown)


shrinedogg (approved):
## Code Review Summary

**Verdict: Approve**

- The new wait strategy flag exists in the pinned helm (4.2.2): `--wait` accepts watcher, hookOnly, and legacy, with bare --wait mapping to watcher, so the old invocation did use the watcher strategy and `--wait=legacy` is a valid fix for the transient InProgress condition being treated as terminal.
- The explicit `kubectl wait fluxinstance/flux` target matches the chart's instance name and namespace, and the CR exists by that point (it is a chart resource), so the wait cannot trip NotFound; a not-yet-present Ready condition simply polls until the timeout.
- The change applies to both profiles: the aws path now also blocks on the operator-owned Ready condition, a small strictness improvement.
- The validate task now runs bash -n on both scripts before the overlay builds; passes locally (38 overlays, no failures). `git diff --check` clean; CI green (the konflate skip is expected for fork PRs).

Approving.

PR #38, opened 2026-08-18
fix: clean up bootstrap reconciliation output

lmilbaum:
## Summary

- limit workload Flux log streaming to actual errors
- remove transient management Flux status rows while retaining authoritative readiness waits
- replace layered EXIT traps with one stable, idempotent cleanup handler

## Testing

- mise run validate
- git diff --check

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,8 +96,9 @@ GitHub PAT so Flux can clone this repository; the local-host profile does not re
 GitHub or AWS credentials.
 
 `mise -E local-host run bootstrap` waits for both the management and workload
-Flux reconciliation chains and streams their progress. A successful bootstrap,
-followed by the Podinfo port-forward, verifies the end-to-end local-host flow.
+Flux reconciliation chains, streams management progress, and surfaces workload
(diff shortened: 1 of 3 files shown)


shrinedogg (approved):
## Code Review Summary

**Verdict: Approve**

- The trap consolidation is behavior-verified, not just read: extracted cleanup_bootstrap and cleanup_workload_reconciliation and exercised them under set -euo pipefail. A live background pid is reaped, the temp kubeconfig is removed, an empty-state invocation is a no-op, and the explicit-call-plus-EXIT-trap double invocation is safe. The script now has exactly one trap statement.
- The mgmt-side change drops the transient-status stream and keeps the authoritative kubectl wait; the docs describe the new behavior accurately, including why the workload stream is error-filtered.
- The new workload pod wait runs before flux logs, so the log streamer starts against existing controller pods (the flux-system Kustomization discovery loop already gates on the operator having created the sync objects).
- `flux logs --level=error` is a valid filter for the CLI.
- `bash -n` and `git diff --check` clean; `mise run validate` passes; CI green (the konflate skip is expected for fork PRs).

Approving.

PR #43, opened 2026-08-22
fix: select native Zarf asset in mise

lmilbaum:
## Summary

- select the native Linux or macOS arm64 Zarf release asset in mise
- update the air-gap prerequisites to match

## Testing

- `mise run validate`

### DIFF
--- a/docs/airgap.md
+++ b/docs/airgap.md
@@ -14,13 +14,10 @@ re-verify.
 
 ## Prerequisites
 
-The kit is single-architecture (arm64), including the pinned Zarf CLI: the
-mise pin in `mise.toml` (`asset_pattern = "zarf_v0.83.0_Darwin_arm64"`)
-resolves only on macOS/arm64, so `mise install` fails on other platforms
-with a confusing asset-not-found error. On Linux, adjust the pin to the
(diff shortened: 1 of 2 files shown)


shrinedogg:
Pushed two commits from my review; the second supersedes the first:

- `f3ac8e5` inlined the pinned version in `asset_pattern` after I hit "Variable `version` not found in context" locally. That finding was an artifact of my own stale mise (2025.11.10), which predates `{{ version }}` in the asset_pattern Tera context. The original template in this PR is correct for current mise: the docs list `{{ version }}` as a supported asset_pattern variable alongside the `os()`/`arch()` remap functions.
- `9c80b9b` restores the original `zarf_v{{ version }}_{{ os(linux="Linux", macos="Darwin") }}_{{ arch() }}` form so the pinned version is not duplicated. Verified on mise 2026.8.10 (Darwin/arm64): the pattern renders and resolves, `zarf version` prints v0.83.0, and `mise run validate` exits 0.

Caveat for anyone on an older mise: this template needs a recent mise; if the "Variable version not found" error appears, `mise self-update` first. Apologies for the noise from my first commit.

PR #42, opened 2026-08-20
ci: add air-gapped workflow

lmilbaum:
## Summary

- add an ARM64 air-gap package workflow triggered on updates to its fork branch or by manual dispatch
- install pinned build tools with mise and authenticate to GHCR with `docker/login-action`
- publish the Flux configuration as an OCI artifact in GHCR while retaining the local-registry default for local builds
- make the pinned Zarf installation portable across Linux and macOS ARM64
- build the complete Zarf air-gap package on GitHub-hosted Ubuntu ARM64 and store its `.tar.zst` as a workflow artifact

## Testing

- fork workflow passed: https://github.com/lmilbaum/knr-ops/actions/runs/32382702623
- OCI artifact published to `ghcr.io/lmilbaum/knr-ops-airgap:latest`
- Zarf package uploaded as workflow artifact `knr-ops-airgap-arm64`
- PR contains one squashed commit

PR #45, opened 2026-08-23
fix: make air-gap sed rewrites portable

lmilbaum:
## Summary

- replace BSD-only `sed -i ''` calls with portable backup-suffix edits
- remove the temporary backup files after each rewrite
- isolate the smallest Linux-runner prerequisite from #42

## Testing

- `bash -n airgap/scripts/build-config-artifact.sh`
- exercised both replacements against a temporary file on macOS
- `git diff --check`


shrinedogg (approved):
## Code Review Summary

**Verdict: Approve**

Verified locally:

- `sed -i.bak` is the portable in-place form: GNU requires the suffix attached, BSD accepts it attached (probed on macOS with the script's exact commands: replacements applied, backup removed, no .bak left). The pipe delimiters are safe for these values (hostnames and cluster names contain no pipes).
- Sweep confirms no remaining BSD-only `sed -i ''` calls under airgap/.
- `bash -n` and `git diff --check` clean; CI green (the konflate skip is expected for fork PRs).

Carving this out of #42 for a fast merge is the right call; it is the only piece the Linux runner strictly needs from that script.

Approving.

PR #49, opened 2026-08-23
fix: make air-gap Docker socket configurable

lmilbaum:
## Summary

- allow the gap-side staging script to use `DOCKER_SOCKET_PATH`
- retain `/var/run/docker.sock` as the default
- fail early with a clear error when the configured socket does not exist
- mount the configured host socket into the kind management node

Extracted as the next small, independent change from #42.

## Testing

- `bash -n airgap/scripts/stage-and-create-cluster.sh`
- verified a nonexistent `DOCKER_SOCKET_PATH` fails with the expected error
- `git diff --check`

### DIFF
--- a/airgap/scripts/stage-and-create-cluster.sh
+++ b/airgap/scripts/stage-and-create-cluster.sh
@@ -24,6 +24,12 @@ ARCHIVES="$AIRGAP_DIR/archives"
 
 CLUSTER_NAME="${CLUSTER_NAME:-mgmt}"
 KIND_NODE_IMAGE="${KIND_NODE_IMAGE:-kindest/node:v1.36.1}"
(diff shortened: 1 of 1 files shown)


shrinedogg (approved):
## Code Review Summary

**Verdict: Approve**

Verified locally:

- The new guard fires before any docker or kind action: `DOCKER_SOCKET_PATH=/nonexistent-sock` exits 1 with the expected error (live probe), and the default `/var/run/docker.sock` passes the `-S` socket test on a Docker Desktop host, so a normal setup is unaffected.
- `-S` is the right test for a socket, and the mount keeps the standard in-node path, so the change only affects which host socket is mapped into the kind node; CAPD's expectations are unchanged.
- `bash -n` and `git diff --check` clean; CI green (the konflate skip is expected for fork PRs).
- A clean, independent extraction from #42.

Approving.

PR #50, opened 2026-08-23
feat: make airgap OCI registry configurable

lmilbaum:
## Summary

- allow the air-gap config artifact destination to be selected with `OCI_REGISTRY`
- add optional registry credentials through `OCI_USERNAME` and `OCI_PASSWORD`
- preserve the existing localhost registry default and insecure-registry behavior

## Testing

- `bash -n airgap/scripts/build-config-artifact.sh`
- `git diff --check`
- underlying `mise run validate` commands: shell checks and every kustomize overlay passed

`mise run validate` could not start directly because the installed mise is 2026.8.9 while the repository requires 2026.8.10.

### DIFF
--- a/airgap/scripts/build-config-artifact.sh
+++ b/airgap/scripts/build-config-artifact.sh
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 # build-config-artifact.sh — build the airgap-trimmed knr-ops config tree and
-# push it to the connected-side local registry, ready to be baked into the
+# push it to the connected-side OCI registry, ready to be baked into the
 # Zarf package (zarf.yaml lists it under the knr-ops-config component images).
 #
-# Connected-side only. Requires: flux CLI, local registry on localhost:5001
-# (the knr-registry container from `mise -E local-host run bootstrap`).
+# Connected-side only. Requires: flux CLI and an OCI registry. By default it
+# uses the knr-registry container from `mise -E local-host run bootstrap`.
 #
 # Why a trimmed tree: in the gap the substrate (cert-manager, CAPI, CAAPH,
 # flux-operator) is deployed by Zarf, and the capi-operator HelmRelease /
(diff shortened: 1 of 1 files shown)


airgap/scripts/build-config-artifact.sh
```diff
@@ -314,12 +315,20 @@ SOURCE_URL=$(git config --get remote.origin.url || true)
...
-  --revision="${GIT_REF}@sha1:${GIT_SHA}" \
-  --insecure-registry \
+PUSH_ARGS=(
+  "$OCI_URL"
+  --path="$ARTIFACT_ROOT"
+  --source="$SOURCE_URL"
+  --revision="${GIT_REF}@sha1:${GIT_SHA}"
   --reproducible
+)
+if [[ "$OCI_REGISTRY" == localhost:* ]]; then
```
shrinedogg, line 325: Suggestion: this glob matches only the `localhost` host form; `OCI_REGISTRY=127.0.0.1:5001` would not receive `--insecure-registry`, and there is no env var to force it, so a plain-HTTP registry addressed via 127.0.0.1 fails on TLS. Default and `localhost:*` both work, so this is narrow: worth either documenting the localhost form or matching both host spellings.

shrinedogg (approved):
## Code Review Summary

**Verdict: Approve** (2 suggestions, neither blocking)

Verified locally:

- Queried the installed flux CLI: `--creds <username>[:<password>]` (generic provider), `--insecure-registry`, and `--reproducible` are all valid `flux push artifact` flags, so the new argument list is sound.
- The default path is unchanged: `OCI_REGISTRY` defaults to `localhost:${REGISTRY_PORT}`, producing the identical `oci://localhost:5001/knr-ops-airgap:latest` URL and still receiving `--insecure-registry`; credentials engage only when both OCI_USERNAME and OCI_PASSWORD are set.
- Replicated the argument assembly under `set -euo pipefail`: 6 args on the default path, a single correctly-quoted `--creds user:pass` token when creds are set.
- Ran `mise run validate` on this head (40 overlays, no failures), which the PR could not start directly because the runner's mise is 2026.8.9 against the required 2026.8.10. `bash -n` and `git diff --check` clean; CI green (the konflate skip is expected for fork PRs).

### Suggestions
- The `localhost:*` gate on line 325 misses the `127.0.0.1` spelling; see the inline comment.
- `zarf.yaml`'s `knr-ops-config` component still references the artifact at the static `localhost:5001/knr-ops-airgap:latest`, so a non-default `OCI_REGISTRY` pushes to a destination the package build would not find. A one-line note that the zarf.yaml reference follows the registry destination would keep the two in sync.

### Looks good
- Keeping the push behind an argument array with conditional flags is cleaner than inline quoting for the optional creds
- The header comment now describes the general contract (any OCI registry) instead of the localhost-only one

Approving.

lmilbaum:
Addressed both follow-up suggestions in #51: loopback registry handling now covers `127.0.0.1`, and package creation uses the same configurable OCI destination while restoring `zarf.yaml` afterward.

airgap/scripts/build-config-artifact.sh
lmilbaum: Addressed in #51: `127.0.0.1:*` now receives `--insecure-registry` alongside `localhost:*`.

PR #51, opened 2026-08-23
fix: align airgap package registry settings

lmilbaum:
## Summary

- treat both `localhost` and `127.0.0.1` OCI registries as insecure local endpoints
- make Zarf package creation consume the same registry, repository, and tag used to publish the config artifact
- restore the tracked `zarf.yaml` after package creation, including failure exits

Addresses the review suggestions from #50.

## Testing

- `bash -n airgap/scripts/build-config-artifact.sh airgap/scripts/build-package.sh`
- `git diff --check`
- verified the Zarf image reference replacement against a temporary manifest copy
- ran the underlying repository validation commands; all shell checks and kustomize builds passed

### DIFF
--- a/airgap/scripts/build-package.sh
+++ b/airgap/scripts/build-package.sh
@@ -2,7 +2,7 @@
 # build-package.sh — connected-side package build.
 #
 #   1. mise run validate (all overlays still build)
-#   2. build-config-artifact.sh (trimmed airgap tree -> localhost:5001 OCI)
+#   2. build-config-artifact.sh (trimmed airgap tree -> configured OCI registry)
 #   3. stage workload-node pod images into archives/ (host-daemon tarball for
 #      CAPD preLoadImages; distinct from the mgmt substrate images the Zarf
 #      package pushes into the internal registry)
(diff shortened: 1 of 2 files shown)


airgap/scripts/build-package.sh
```diff
@@ -21,6 +21,33 @@ mise run validate
...
+  OCI_REPOSITORY="${OCI_REPOSITORY:-knr-ops-airgap}"
+  OCI_TAG="${OCI_TAG:-latest}"
+  ZARF_CONFIG="$REPO_ROOT/airgap/zarf.yaml"
+  ZARF_CONFIG_BACKUP=$(mktemp "${TMPDIR:-/tmp}/knr-ops-zarf.XXXXXX")
+  cp "$ZARF_CONFIG" "$ZARF_CONFIG_BACKUP"
+  restore_zarf_config() {
+    cp "$ZARF_CONFIG_BACKUP" "$ZARF_CONFIG"
+    rm -f "$ZARF_CONFIG_BACKUP"
+  }
+  trap restore_zarf_config EXIT
```
shrinedogg, line 37: Suggestion: the restore is a `trap … EXIT`, which fires on normal and error exits and even SIGTERM (verified), but bash skips the EXIT trap on SIGINT, so a Ctrl-C during the package build leaves zarf.yaml with the rewritten reference. It's self-detecting — the next run's `MATCH_COUNT -ne 1` guard fails with the "expected exactly one config artifact reference" error — and `git checkout -- airgap/zarf.yaml` restores it, but a `trap restore_zarf_config EXIT INT TERM` (or a startup check that resets an already-rewritten ref) would make it airtight.

shrinedogg (approved):
## Code Review Summary

**Verdict: Approve** (1 suggestion, non-blocking)

Verified locally:

- The zarf-rewrite block is gated on `OCI_REGISTRY` being explicitly set, so an ordinary build leaves zarf.yaml byte-identical and the push path is unchanged — the default (`localhost:5001/knr-ops-airgap:latest`) still works end to end.
- The `127.0.0.1` gate resolves the #50 suggestion: both `localhost:*` and `127.0.0.1:*` now receive `--insecure-registry`.
- Exactly-one-occurrence guard is sound: `grep -Fc` returns 1 on current zarf.yaml, the rewrite preserves the `oci://` prefix, and zero `localhost:5001` remain afterward.
- Restore semantics probed on a scratch copy: zarf.yaml returns to its original bytes on normal exit, on error exit (the "including failure exits" claim holds), and on SIGTERM. `bash -n` on both scripts and `git diff --check` clean; `mise run validate` passes (40 overlays, no failures); CI green (the konflate skip is expected for fork PRs).

### Suggestion
- The `trap … EXIT` restore does not fire on SIGINT (Ctrl-C), so a Ctrl-C mid-build can leave zarf.yaml rewritten. See the inline comment on build-package.sh:37 — it's self-detecting and recoverable, so this is not blocking, but trapping SIGINT too (or a startup reset) would make it airtight.

### Looks good
- Resolves both #50 suggestions cleanly
- Backing up to a mktemp file and restoring from it (rather than an in-place undo) is the robust shape, and the count guard doubles as the self-detection for the one edge case
- The header comment now reflects the configurable-registry behavior instead of the localhost-only one

Approving.

airgap/scripts/build-package.sh
lmilbaum: Addressed in the updated single commit (`cc1adba`): explicit INT and TERM handlers exit with 130/143, which runs the existing EXIT restoration exactly once. Verified with a SIGINT test that the original manifest is restored and the backup removed.

PR #53, opened 2026-08-24
fix: stop airgap run after deployment failure

lmilbaum:
## Summary

- stop the offline validation immediately when cluster staging fails
- stop when Zarf initialization or package deployment fails
- avoid misleading downstream verification errors after a prerequisite failure

## Testing

- `git diff --check`

### DIFF
--- a/airgap/scripts/offline-run.sh
+++ b/airgap/scripts/offline-run.sh
@@ -62,6 +62,7 @@ if CLUSTER_NAME=$CLUSTER_NAME REGISTRY_NAME=$REGISTRY_NAME REGISTRY_PORT=$REGIST
   pass "stage (docker load, kind create, registry seed)"
 else
   fail "stage-and-create-cluster.sh"
+  exit 1
 fi
 
 step "2. zarf init"
@@ -70,6 +71,7 @@ if ( cd "$AIRGAP_DIR" && "$ZARF" init "$ARCHIVES/zarf-init-arm64-v0.83.0.tar.zst
   pass "zarf init"
 else
   fail "zarf init"
+  exit 1
 fi
 
 step "3. zarf package deploy"
@@ -77,6 +79,7 @@ if ( cd "$AIRGAP_DIR" && "$ZARF" package deploy "$ARCHIVES/zarf-package-knr-ops-
   pass "zarf package deploy"
 else
   fail "zarf package deploy"
+  exit 1
 fi
 
 step "4. verify mgmt substrate (all non-baked images from 127.0.0.1:31999)"


airgap/scripts/offline-run.sh
```diff
@@ -62,6 +62,7 @@ if CLUSTER_NAME=$CLUSTER_NAME REGISTRY_NAME=$REGISTRY_NAME REGISTRY_PORT=$REGIST
   pass "stage (docker load, kind create, registry seed)"
 else
   fail "stage-and-create-cluster.sh"
+  exit 1
```
shrinedogg, line 65: Note on why this works: the `exit 1` runs in the top-level `{ ... } > "$LOG" 2>&1` grouping (current shell), so it exits the whole script. The `( cd ... && zarf ... )` subshell wraps only the `if` condition and does not affect it. This is also why the script keeps `set -uo pipefail` without `-e`: adding `-e` would abort the verification steps that the final PASS/FAIL gate aggregates.

shrinedogg (approved):
Reviewed with verification. The change is correct and minimal: it adds `exit 1` to the failure branches of the three prerequisite steps (stage, zarf init, package deploy), so a failure stops the script instead of falling through to the read-only verification steps (4-8), which would otherwise report confusing downstream errors against a half-provisioned cluster.

Verified locally:
- `bash -n airgap/scripts/offline-run.sh` passes
- `git diff --check` clean
- CI: Build kustomize overlays and Lint YAML pass; Rendered Flux diff skipped (expected: this PR touches no Flux-managed tree)
- Confirmed the `exit 1` lines run in the top-level `{ ... } > "$LOG"` grouping (current shell), so they exit the whole script with code 1; the `( cd ... && zarf ... )` subshell wraps only the `if` condition and does not swallow the else-branch exit
- Reproduced the semantics with a harness: on a failing prerequisite the script stops before the next step and exits 1, and the downstream steps do not run

Why `exit 1` and not `set -e`: the script deliberately uses `set -uo pipefail` (without `-e`) so all the read-only verification steps (4-8) run and are aggregated by the final PASS/FAIL gate. Fail-fast is scoped to the three mutating/prerequisite steps, which is the right boundary.

Looks good:
- Minimal 3-line change, one concern per line, matches the PR title
- Fail-fast ordering is correct: the exits fire before the downstream verification, not after
- No doc update needed: docs/airgap.md documents the sequence and verification checklist and does not promise continue-on-failure behavior

airgap/scripts/offline-run.sh
lmilbaum: Thanks—that’s intentional. The script omits `set -e` so the read-only verification steps can all run and contribute to the final PASS/FAIL summary. The explicit `exit 1` statements scope fail-fast behavior to the three prerequisite steps—staging, Zarf initialization, and package deployment—because later checks are not meaningful if any of those fail.

PR #54, opened 2026-08-24
feat: accept airgap package input path

lmilbaum:
## Summary

- accept an optional Zarf package path in `offline-run.sh`
- normalize the selected package to an absolute path before changing directories
- preserve the existing archive path when no argument is supplied

## Testing

- `git diff --check`
- verified normalization of a package path containing spaces

### DIFF
--- a/airgap/scripts/offline-run.sh
+++ b/airgap/scripts/offline-run.sh
@@ -8,6 +8,7 @@
 # Usage (operator):
 #   1. an agent (or you) starts this script in the background while online:
 #        nohup airgap/scripts/offline-run.sh >/dev/null 2>&1 &
+#        nohup airgap/scripts/offline-run.sh /path/to/package.tar.zst >/dev/null 2>&1 &
 #   2. toggle Wi-Fi OFF within 6 minutes.
 #   3. watch:  tail -f /tmp/airgap-offline-run.log
 #   4. when the log shows "OFFLINE RUN COMPLETE", toggle Wi-Fi back ON and
@@ -18,6 +19,13 @@ LOG=/tmp/airgap-offline-run.log
 SUMMARY=/tmp/airgap-offline-summary.txt
 AIRGAP_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
 ARCHIVES="$AIRGAP_DIR/archives"
+PACKAGE_INPUT="${1:-$ARCHIVES/zarf-package-knr-ops-airgap-arm64-0.1.0.tar.zst}"
+PACKAGE_DIR=$(cd "$(dirname "$PACKAGE_INPUT")" && pwd)
+PACKAGE="$PACKAGE_DIR/$(basename "$PACKAGE_INPUT")"
+if [ ! -f "$PACKAGE" ]; then
+  echo "offline-run.sh: package not found: $PACKAGE" >&2
+  exit 1
+fi
 
 # Direct binary paths (offline-safe; no mise/network resolution).
 ZARF="$HOME/.local/share/mise/installs/github-zarf-dev-zarf/v0.83.0/zarf"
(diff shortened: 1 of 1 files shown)


airgap/scripts/offline-run.sh
```diff
@@ -18,6 +18,13 @@ LOG=/tmp/airgap-offline-run.log
 SUMMARY=/tmp/airgap-offline-summary.txt
 AIRGAP_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
 ARCHIVES="$AIRGAP_DIR/archives"
+PACKAGE_INPUT="${1:-$ARCHIVES/zarf-package-knr-ops-airgap-arm64-0.1.0.tar.zst}"
+PACKAGE_DIR=$(cd "$(dirname "$PACKAGE_INPUT")" && pwd)
+PACKAGE="$PACKAGE_DIR/$(basename "$PACKAGE_INPUT")"
+if [ ! -f "$PACKAGE" ]; then
```
shrinedogg, line 25: Verification for the guard pushed in 1de9645: run with a nonexistent directory the script now exits rc=1 before the offline wait loop (stderr: "offline-run.sh: package not found: ..."); a trailing-slash input is caught the same way; a valid package path containing spaces passes the guard and reaches step 0. This also covers the case where the default archive is absent, and keeps the fail-fast intent of #53 without set -e.

airgap/scripts/offline-run.sh
```diff
@@ -18,6 +18,13 @@ LOG=/tmp/airgap-offline-run.log
 SUMMARY=/tmp/airgap-offline-summary.txt
 AIRGAP_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
 ARCHIVES="$AIRGAP_DIR/archives"
+PACKAGE_INPUT="${1:-$ARCHIVES/zarf-package-knr-ops-airgap-arm64-0.1.0.tar.zst}"
```
shrinedogg, line 22: Small follow-up (non-blocking): the usage header above (lines 8-14) still shows only the no-argument invocation. One extra line there, e.g. `nohup airgap/scripts/offline-run.sh /path/to/package.tar.zst ...`, would document the new input where operators look first.

shrinedogg (approved):
Approved with one follow-up commit pushed to this branch (1de9645: fail-fast guard for the package input).

Verified locally, on 95a694d (original) and 1de9645 (guard commit):

- bash -n airgap/scripts/offline-run.sh and git diff --check clean on both
- CI on 95a694d: Build kustomize overlays and Lint YAML pass; Rendered Flux diff skipped as expected for a scripts-only change
- Harness exercising the normalization block (same set -uo pipefail semantics): the no-argument default resolves to the previous hardcoded absolute path; relative, bare-filename, and space-containing inputs normalize correctly; the quoted expansion is safe at the deploy call
- Guard commit verified live against the script: a bad package path exits rc=1 immediately with a message on stderr, before the offline wait loop or any staging; a trailing-slash input is caught; a valid space-containing package passes the guard and proceeds to step 0
- Sweep of other references to the package name: the zarf deploy here is the only programmatic consumer; docs/airgap.md and the stage script echo describe the default path and stay accurate

One small non-blocking follow-up for the author: the usage header (lines 8-14) still documents only the no-argument invocation.

airgap/scripts/offline-run.sh
lmilbaum: Kept the fail-fast package guard and folded it into the squashed single commit (`221a90d`). Verified that a missing package exits with status 1 before the offline wait.

airgap/scripts/offline-run.sh
lmilbaum: Addressed in the squashed single commit (`221a90d`) by adding the explicit package-path invocation to the usage header.

shrinedogg (approved):
LGTM!

PR #55, opened 2026-08-24
feat: stream airgap run logs

lmilbaum:
## Summary

- stream offline-run progress to stdout while retaining the complete log file
- preserve failures through the `tee` pipeline via the script’s existing `pipefail` setting

## Testing

- `git diff --check`
- verified output reaches stdout and the log file
- verified a simulated exit status 7 remains 7 through the `tee` pipeline

### DIFF
--- a/airgap/scripts/offline-run.sh
+++ b/airgap/scripts/offline-run.sh
@@ -167,4 +167,4 @@ else
   cat "$SUMMARY"
   exit 0
 fi
-} > "$LOG" 2>&1
+} 2>&1 | tee "$LOG"


airgap/scripts/offline-run.sh
```diff
@@ -167,4 +167,4 @@ else
   cat "$SUMMARY"
   exit 0
 fi
-} > "$LOG" 2>&1
+} 2>&1 | tee "$LOG"
```
shrinedogg, line 170: Verified: exit status passes through the tee pipeline intact (block exits 0 and 7 -> overall 0 and 7, thanks to set -o pipefail and tee succeeding), stdout gets the live stream, and $LOG keeps the full transcript.

Context for the record, non-blocking: if stdout's reader closes early (only possible in attached usage; the documented nohup >/dev/null background usage never closes it), the block dies on SIGPIPE (rc 141) mid-deploy instead of reporting its real status. If that ever matters, `trap '' PIPE` before the block restores the true status through an early-closed reader. Not needed for this PR.

shrinedogg (approved):
Approved. Verified locally:

- bash -n airgap/scripts/offline-run.sh and git diff --check clean
- CI: Build kustomize overlays and Lint YAML pass; Rendered Flux diff skipped as expected for a scripts-only change
- Exit-status preservation: exercised the pipeline shape with block exits 0 and 7; with the script's set -o pipefail and tee succeeding, PIPESTATUS[0] and the overall status match the block's exit, so the final PASS/FAIL gate's exit 0/1 and the per-step exit 1s survive the tee pipeline
- Output routing: stdout receives the live stream and $LOG still receives the full transcript
- Failure-path probes: an early-closed stdout reader turns into SIGPIPE (rc 141) and an unwritable $LOG makes tee die while the deploy continues (rc 1); both are pre-existing shapes for this script's documented usage (nohup ... >/dev/null 2>&1 &, /tmp writable) and are not introduced by this PR

Looks good:
- Minimal one-line change scoped to output routing, no logic touched
- Streaming makes tail -f $LOG optional; the Wi-Fi-off wait loop is now observable live
- Reuses the script's existing pipefail rather than new machinery

airgap/scripts/offline-run.sh
lmilbaum: Thanks for verifying the status propagation and output routing. Agreed that no change is needed here: the documented `nohup ... >/dev/null 2>&1 &` usage keeps the output sink open. If we later support attached consumers that may close stdout early, we can handle SIGPIPE explicitly as a separate behavior change.

PR #56, opened 2026-08-24
fix: discover installed airgap tools

lmilbaum:
## Summary

- resolve required air-gap tools from `PATH` before entering offline mode
- support both macOS operator installs and Linux CI without hardcoded executable paths
- avoid invoking mise after the environment is disconnected

## Testing

- `git diff --check`
- verified all six tools resolve correctly from a controlled `PATH`


shrinedogg (approved):
Appreciated, Liora! LGTM

PR #57, opened 2026-08-24
feat: allow monitored airgap runs

lmilbaum:
## Summary

- allow callers to skip the interactive offline-connectivity wait with `SKIP_OFFLINE_CHECK=1`
- retain the existing Wi-Fi-off confirmation loop by default
- make the override explicit for callers that independently enforce and monitor network isolation

## Testing

- `git diff --check`
- verified the override bypasses connectivity probes
- verified the default branch still confirms unreachable registries

### DIFF
--- a/airgap/scripts/offline-run.sh
+++ b/airgap/scripts/offline-run.sh
@@ -13,6 +13,8 @@
 #   3. watch:  tail -f /tmp/airgap-offline-run.log
 #   4. when the log shows "OFFLINE RUN COMPLETE", toggle Wi-Fi back ON and
 #      tell the agent to collect the results.
+# Set SKIP_OFFLINE_CHECK=1 only when the caller enforces network isolation and
+# monitors external traffic; by default this script waits to confirm isolation.
 set -uo pipefail
 
 LOG=/tmp/airgap-offline-run.log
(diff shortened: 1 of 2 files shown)


shrinedogg (approved):
## Code Review Summary

Verified locally, in a sandbox with stubbed tools and a mode-switchable curl:

- All four paths behave as claimed: with SKIP_OFFLINE_CHECK=1 the script logs the skip and proceeds past step 0 while the internet is up; unset with internet down it confirms OFFLINE and proceeds; unset with the internet up and a zero wait it times out with exit 1 and the expected FAIL summary line; SKIP_OFFLINE_CHECK=0 is not a skip (only the literal 1 is), so the conservative default holds.
- The default branch is byte-identical to main modulo indentation; the wrap adds only the gate, the log line, else, and fi.
- The sandbox runs also re-verified the #53 fail-fast on this base: a failing stage script stops the run immediately with exit 1.
- bash -n and git diff --check clean; mise run validate passes on this head (40 overlays); CI green (the konflate skip is expected for fork PRs).

### Suggestions
- The final result line (offline-run.sh:171) unconditionally prints "RESULT: PASS - full airgap deploy verified with no internet". When the check was skipped, this script did not verify that; the caller's monitoring did. Appending a "(connectivity check skipped; caller-monitored)" suffix in skip mode would keep the log honest.
- The new mode is undocumented: neither the header's operator workflow (lines 8-15) nor docs/airgap.md mentions SKIP_OFFLINE_CHECK. A header line stating the contract (default waits for isolation; 1 means the caller enforces and monitors it) would keep the flag from being flipped casually.

### Looks good
- Gating on the exact value 1 rather than a nonempty check is the conservative shape; unset, 0, and garbage all keep the wait
- The skip log line records the rationale ("external traffic is monitored by the caller"), which keeps the audit trail self-explanatory

Approving.

PR #58, opened 2026-08-24
docs: clarify airgap isolation mode logging

lmilbaum:
## Summary

- make the isolation step heading accurate in both self-checked and caller-monitored modes
- log the active isolation mode explicitly at the start of an offline run
- document the caller-monitored invocation with its operator responsibility

## Testing

- `git diff --check main...HEAD`


shrinedogg (approved):
LGTM

PR #59, opened 2026-08-24
feat: stage airgap host assets

lmilbaum:
## Summary

- download the pinned ARM64 Zarf init package during the connected-side build
- pull and archive the management node, workload node, CAPD load-balancer, and local-registry images
- make `build-package.sh` produce every host-side asset documented for transfer across the gap

## Testing

- all `mgmt/` and `workload/` kustomize overlays built successfully with `kubectl kustomize`
- `git diff --check` passed
- `mise run validate` could not start locally because the installed mise is 2026.8.9 while the repository requires 2026.8.10

### DIFF
--- a/airgap/scripts/build-package.sh
+++ b/airgap/scripts/build-package.sh
@@ -3,9 +3,8 @@
 #
 #   1. mise run validate (all overlays still build)
 #   2. build-config-artifact.sh (trimmed airgap tree -> configured OCI registry)
-#   3. stage workload-node pod images into archives/ (host-daemon tarball for
-#      CAPD preLoadImages; distinct from the mgmt substrate images the Zarf
-#      package pushes into the internal registry)
+#   3. stage the Zarf init package, host-daemon images, workload-node pod
+#      images, and OCI charts into archives/
 #   4. zarf package create
 #
 # Output: zarf-package-knr-ops-airgap-arm64-0.1.0.tar.zst next to airgap/.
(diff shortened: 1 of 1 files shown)


shrinedogg (approved):
## Code Review Summary

**Verdict: Approve** (1 note, non-blocking)

Verified locally:

- Ran `mise x -- zarf tools download-init --architecture arm64` for real (zarf 0.83.0, the mise.toml pin): it produced exactly `zarf-init-arm64-v0.83.0.tar.zst`, the filename offline-run.sh:85 hardcodes for the gap-side deploy. The `-o`/`-a` flags were checked against the installed CLI.
- The four docker save targets match the `*.tar` glob that stage-and-create-cluster.sh loads, and match the docs/airgap.md transfer table row for row; the archives left over from the validated radio-off run carry the identical filenames.
- Pulled, saved, reloaded, and inspected `registry:2` with `--platform linux/arm64`: the platform survives the round trip (arch=arm64, os=linux).
- Image pins are consistent: mgmt node v1.36.1 is the stage script's KIND_NODE_IMAGE default, workload node v1.35.0 is cluster-class.yaml's customImage, and the haproxy tag matches the image the validated offline run actually used (the CAPD CRD doc-string default v20210715-a6da3463 is stale relative to what the controller pulls).
- bash -n and git diff --check clean; mise run validate passes on this head (40 overlays), which the PR could not run due to the stale local mise 2026.8.9; CI green (the konflate skip is expected for fork PRs).

### Note
- The version pins now span three files: HOST_IMAGES here (plus the mise zarf pin for the init package), stage-and-create-cluster.sh's KIND_NODE_IMAGE default, and cluster-class.yaml's customImage (kindest/node:v1.35.0). A one-line header comment noting the coupling would help the next version bump land in all three places together.

### Looks good
- The script now produces every host-side asset documented for transfer, so the connected-side build is complete end to end
- Relative output paths (airgap/archives) are consistent with the rest of the script

Approving.

PR #61, opened 2026-08-24
ci: build airgap transfer bundle

lmilbaum:
## Summary

- add an ARM64 GitHub Actions job for the connected-side airgap bundle build
- start an ephemeral local OCI registry and run the existing package builder
- upload the complete transfer bundle as a one-day workflow artifact
- document the new build workflow in the airgap guide and agent repository map

Deployment and network-isolation verification remain separate follow-up slices.

## Testing

- all `mgmt/` and `workload/` overlays built successfully with `kubectl kustomize`
- `git diff --check` passed

### DIFF
--- a/.github/workflows/air-gapped.yml
+++ b/.github/workflows/air-gapped.yml
@@ -1,0 +1,60 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+name: air-gapped
+
+on:
+  # Intentionally run for every PR: package inputs span the repository. A path
+  # filter could miss new inputs while still producing a successful but
+  # incomplete bundle.
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: air-gapped-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build air-gap bundle
+    runs-on: ubuntu-24.04-arm
+    timeout-minutes: 15
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5001:5000
+        options: >-
+          --health-cmd "wget --spider --quiet http://localhost:5000/v2/"
+          --health-interval 2s
+          --health-timeout 5s
+          --health-retries 15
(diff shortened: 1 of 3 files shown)


shrinedogg (approved):
## Code Review Summary

Verified locally and against the real workflow run:

- The check itself is the proof: "Build air-gap bundle" passed on this head in 3m24s and produced a 1.8 GB, 25-file artifact. The CI log shows the 408M zarf package tarball matching the upload glob.
- Traced the env vars through build-config-artifact.sh: AIRGAP_CLUSTER_NAME=airgap-wl and WORKLOAD_REGISTRY_HOST=knr-registry-airgap drive the rehearsal renames that stage-and-create-cluster.sh and offline-run.sh expect gap-side; OCI_REGISTRY=localhost:5001 matches the registry service's port mapping and the localhost:* insecure-registry gate, and the zarf.yaml rewrite/restore path from #51 handles the override.
- Workflow hygiene checks out: least-privilege permissions, cancel-in-progress concurrency, a 15-minute timeout, a registry health check, if-no-files-found: error, and retention-days: 1. The mise-action pin (2026.8.10) is exactly the version the repo requires.

### Suggestions
- The docs transfer table lists the zarf CLI binary as the first item crossing the gap, but the artifact omits it (and a linux/arm64 runner could not produce the macOS binary the operator flow uses anyway). A one-line note in docs/airgap.md that the CI bundle excludes the CLI (fetch via mise or the zarf release for the target OS) would keep the "complete transfer bundle" claim accurate.
- The workflow runs on every PR at roughly 3.5 minutes of ARM runner time plus a 1.8 GB artifact. A paths: filter would trim that, though the build also exercises the full-repo validation, so the broad trigger has real value; up to you.

Approving.

lmilbaum:
I see the point about using a `paths` filter, and I considered it. I am keeping the every-PR trigger intentionally because the package inputs are spread across several directories. A filter could miss a newly added file or directory that becomes part of the package; the workflow could still pass while producing a bundle that is incomplete or differs from what we expect. Running on every PR gives us full-repository integration coverage and avoids that silent gap.

PR #63, opened 2026-08-24
ci: monitor airgap deployment traffic

lmilbaum:
## Summary

- consume the ARM64 bundle produced by the existing air-gapped build job
- deploy it on a fresh ARM64 runner while capturing public traffic from the kind bridge
- fail when the deployment fails or any public packet is observed
- retain deployment logs, summaries, and packet-capture evidence for one day
- document the monitored deployment stage in the airgap guide and agent repository map

Enforced network blocking remains a separate follow-up slice.

## Testing

- all `mgmt/` and `workload/` overlays built successfully with `kubectl kustomize`
- `git diff --check` passed

### DIFF
--- a/.github/workflows/air-gapped.yml
+++ b/.github/workflows/air-gapped.yml
@@ -2,22 +2,21 @@
 name: air-gapped
 
 on:
-  # Intentionally run for every PR: package inputs span the repository. A path
-  # filter could miss new inputs while still producing a successful but
-  # incomplete bundle.
-  pull_request:
+  schedule:
+    - cron: "17 2 * * *"
   workflow_dispatch:
 
 permissions:
   contents: read
 
 concurrency:
-  group: air-gapped-${{ github.event.pull_request.number || github.ref }}
+  group: air-gapped-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
   build:
     name: Build air-gap bundle
+    if: github.repository == 'polarsquad/knr-ops' && github.ref == 'refs/heads/main'
     runs-on: ubuntu-24.04-arm
     timeout-minutes: 15
     services:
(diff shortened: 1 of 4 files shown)


shrinedogg (approved):
## Code Review Summary

Verified locally:

- Replayed crafted pcaps through the exact tcpdump filter (tcpdump -r, no root): 14 of 14 cases correct. Public IPv4 (8.8.8.8, 1.1.1.1) and public IPv6 destinations are captured; RFC1918 (10/8, 172.16/12, 192.168/16), loopback, link-local, IPv6 ULA, and genuine multicast frames (01:00:5e and 33:33 MACs) are all excluded. The bare `multicast` primitive matches Ethernet multicast, which is what real IP-multicast traffic uses, so the exclusions hold for on-the-wire traffic.
- The capture fails closed: a tcpdump that cannot start aborts the step before any deployment happens, and any observed public packet fails the job with the decoded evidence printed.
- The new offline-run.sh retry loop (20 x 15s for the workload Flux OCIRepository) addresses the real race where the workload API is reachable before its Flux source has reconciled, and it sits correctly inside the LB-port branch.
- Triggers and docs are consistent: schedule + workflow_dispatch matches what AGENTS.md and docs/airgap.md describe. bash -n and git diff --check clean.

### Suggestions
- This monitored-deploy job has never run: it is new, and the trigger change means it cannot execute on this PR, so its first run would be tonight's cron on main. Unproven pieces include actions/download-artifact@v8.0.0 (new in this PR; #61 only proved checkout@v7.0.1 and upload-artifact@v7.0.1), tcpdump availability on the ubuntu-24.04-arm image, and the 45-minute budget. A manual workflow_dispatch from this branch before merging would prove all of it in one shot.
- Dropping the pull_request trigger removes the per-PR bundle-build gate from #61, so a build-package.sh regression now surfaces only in the nightly after merge. If that matters, keep pull_request for the ~3.5 minute build job alone and leave the monitored deploy on schedule/dispatch.

### Looks good
- Monitoring via the kind bridge with an explicit public-destination filter is a clean way to assert isolation without blocking the network, and the skip-gate handoff (SKIP_OFFLINE_CHECK=1 with caller-monitored traffic) lines up exactly with the #57 contract
- Evidence upload uses if: always(), so failure runs keep their logs and pcap

Approving.

lmilbaum:
The workflow has been simplified and restricted to the upstream default branch.

What changed:

- It runs nightly at 02:00 UTC and can also be started with `workflow_dispatch`.
- The build job is guarded to run only when `github.repository == 'polarsquad/knr-ops'` and `github.ref == 'refs/heads/main'`. This prevents expensive runs from forks, PR branches, or manual dispatches from non-default branches.\n- Each eligible run builds a fresh ARM64 bundle, uploads it as a one-day artifact, and the dependent monitored-deploy job downloads and deploys that bundle from the same workflow run.\n- Workload Flux readiness now waits for asynchronous reconciliation instead of checking only once.\n- The full path—including `actions/download-artifact@v8.0.0`, tcpdump on `ubuntu-24.04-arm`, and the deployment timeout—was proven by successful run https://github.com/polarsquad/knr-ops/actions/runs/32744400763.\n\nThere is no per-PR bundle build now. Restoring one would provide earlier detection of `build-package.sh` regressions, but it would build and upload a bundle that no deployment job consumes, adding roughly three minutes of ARM64 runner time to every PR. Given that the nightly/manual run validates the complete build-and-deploy path, keeping the current setup is reasonable. We can reassess if nightly failures show that earlier build-only feedback would have materially helped.

PR #64, opened 2026-08-24
ci: detect upstream default branch dynamically

lmilbaum:
## Summary

- replace the hardcoded upstream repository and `main` ref guard with fork/default-branch metadata
- keep ARM64 air-gap jobs restricted to a non-fork repository on its declared default branch
- update the air-gap documentation to describe the portable guard

## Testing

- `git diff --check`
- workflow delta matches the post-merge change previously reviewed on PR #63

### DIFF
--- a/.github/workflows/air-gapped.yml
+++ b/.github/workflows/air-gapped.yml
@@ -16,7 +16,9 @@ concurrency:
 jobs:
   build:
     name: Build air-gap bundle
-    if: github.repository == 'polarsquad/knr-ops' && github.ref == 'refs/heads/main'
+    if: >-
+      github.event.repository.fork == false &&
+      github.ref_name == github.event.repository.default_branch
     runs-on: ubuntu-24.04-arm
     timeout-minutes: 15
     services:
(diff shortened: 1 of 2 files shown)


shrinedogg (approved):
## Code Review Summary

**Verdict: Approve**

Verified locally:

- Walked the guard equivalence case by case. Upstream main schedule: both forms pass. Fork: the old form skipped on the hardcoded repo name, the new form skips on fork=true. Manual dispatch from a non-default branch on upstream: both skip. Every trigger this workflow has (schedule, workflow_dispatch) populates github.event.repository with the fork and default_branch fields, so the context is always available.
- The new form is strictly more portable: a repo rename, transfer, or default-branch rename breaks the hardcoded guard but not the metadata-driven one.
- monitored-deploy inherits the guard through needs: build, so no second guard is required; a skipped build skips the dependent job.
- The remaining polarsquad reference (KONFLATE_REPO in konflate.yml) is konflate instance config, not a runner guard, and is correctly untouched. The docs sentence matches the new behavior; git diff --check clean; CI green (the konflate skip is expected for fork PRs).

Approving.

PR #65, opened 2026-08-24
fix: diagnose airgap registry readiness failures

lmilbaum:
## Summary

- configure Docker-native health checks for newly created gap-side registries
- wait for Docker to report the registry healthy and expose container status/logs on failure
- verify the published host port separately through the Registry `/v2/` API
- preserve compatibility with existing registry containers that have no health-check configuration

## Testing

- `bash -n airgap/scripts/stage-and-create-cluster.sh`
- `git diff --check`
- `shellcheck airgap/scripts/stage-and-create-cluster.sh` when ShellCheck is available

### DIFF
--- a/airgap/scripts/stage-and-create-cluster.sh
+++ b/airgap/scripts/stage-and-create-cluster.sh
@@ -59,13 +59,46 @@ kubectl wait --for=condition=Ready node --all --timeout=180s
 # the Zarf internal registry is only reachable inside the mgmt cluster).
 REGISTRY_NAME="${REGISTRY_NAME:-knr-registry}"
 REGISTRY_PORT="${REGISTRY_PORT:-5001}"
+registry_failure() {
+  echo "ERROR: registry '${REGISTRY_NAME}' $*" >&2
+  docker ps -a --filter "name=^${REGISTRY_NAME}$" >&2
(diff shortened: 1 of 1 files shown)


shrinedogg (approved):
## Code Review Summary

**Verdict: Approve**

Verified locally, against live containers on this host:

- The health-cmd premise holds: registry:2 ships busybox wget at /usr/bin/wget, so the in-container check can run. Without it, every newly created registry would report unhealthy.
- Healthy path: a probe registry created with the PR's exact flags went healthy in about 4 seconds, and the new curl verification (with --retry-all-errors and --retry-max-time 30) passes against it.
- Unhealthy path: a probe with a deliberately broken health-cmd flipped to unhealthy; the wait loop breaks on that status early, and registry_failure then surfaces docker ps -a and docker logs, the diagnostics this PR is for.
- Compatibility path: a pre-existing registry container with no health config reports none via the inspect template, so the script skips the health wait and falls through to the curl check, preserving the old behavior for containers created before this change.
- An exited registry container is still recreated correctly (docker ps only matches running containers), and the recreated container picks up the new health config.
- bash -n and git diff --check clean; CI green (the konflate skip is expected for fork PRs). shellcheck is not installed on this machine, so I could not reproduce that one.

Approving.

PR #66, opened 2026-08-24
ci: verify deployment with external egress blocked

lmilbaum:
## Summary

- add a second air-gap deployment job that blocks new external connections from the kind Docker network
- preserve established reply traffic so host-to-container registry access continues to work
- capture attempted public traffic and retain isolated-run evidence
- run monitored and isolated deployments in parallel from the same bundle

## Evaluation intent

Running both jobs is temporary. Parallel runs provide directly comparable evidence for:

- which method identifies offline violations more accurately
- whether either method produces false positives or misses attempted egress
- runtime and performance differences between observation and enforced isolation

After enough nightly and manual runs have been assessed, the less effective job will be removed.

## Testing

- `git diff --check`
- `bash -n airgap/scripts/offline-run.sh airgap/scripts/stage-and-create-cluster.sh`
- workflow YAML parsed locally
- the isolated deployment approach previously passed in https://github.com/lmilbaum/knr-ops/actions/runs/32743099384

## Runtime impact

- adds approximately eight GitHub-hosted ARM64 runner-minutes per nightly/manual run
- does not intentionally extend wall-clock duration because both deployments run concurrently after the build

### DIFF
--- a/.github/workflows/air-gapped.yml
+++ b/.github/workflows/air-gapped.yml
@@ -61,6 +61,120 @@ jobs:
           retention-days: 1
           compression-level: 0
 
+  isolated-deploy:
+    name: Deploy without external egress
+    needs: build
+    runs-on: ubuntu-24.04-arm
+    timeout-minutes: 45
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7.0.1
+
+      - name: Install mise
+        uses: jdx/mise-action@v4.2.3
+        with:
+          version: 2026.8.10
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+
+      - name: Download air-gap bundle
+        uses: actions/download-artifact@v8.0.0
+        with:
+          name: knr-ops-airgap-arm64
+          path: airgap
+
+      - name: Deploy with external egress blocked
+        env:
+          SKIP_OFFLINE_CHECK: "1"
+        run: |
+          set -uo pipefail
+
+          capture=/tmp/airgap-public-egress.pcap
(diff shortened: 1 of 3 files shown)


shrinedogg (approved):
## Code Review Summary

**Verdict: Approve**

Verified locally:

- Replicated the exact DOCKER-USER rule on a scratch bridge in a Docker VM and exercised all three traffic classes: public egress from the bridge is dropped (a wget to github.com hangs and dies), intra-bridge container-to-container traffic passes (a registry:2 /v2/ round trip), and host-to-published-port access passes, so the conntrack NEW qualifier preserves established reply traffic exactly as claimed. The comment-matched cleanup deletes the rule and confirms it is gone.
- The prior fork run referenced in the PR (lmilbaum/knr-ops run 32743099384) is real and green across all three jobs, including Deploy without external egress, so the enforced-isolation deployment has already passed end to end on the real runner.
- The tcpdump filter is byte-identical to the one from #63 (verified there against crafted pcaps: public IPv4/IPv6 matched, RFC1918/loopback/link-local/multicast excluded). Attempted egress is captured at the bridge before the DROP, so attempt evidence survives the blocking.
- Structure checks out: pure addition with monitored-deploy untouched, both deploy jobs run in parallel off needs: build alone, the fork/default-branch guard on build is inherited by both deploy jobs, the evidence artifact names are distinct, the cron (17 2) matches the 02:17 UTC in the docs, and the temporary-evaluation framing is documented consistently in AGENTS.md and docs/airgap.md. git diff --check clean; CI green (the konflate skip is expected for fork PRs).

Approving.

Issue #62, opened 2026-08-24
Centralize project dependency version pins


lmilbaum:
## Problem

Dependency versions are duplicated across the repository, not only in the airgap bundle. The same tool, controller, chart, container image, or Kubernetes node version can appear in several independent consumers, including:

- `mise.toml` and environment-specific mise files
- GitHub Actions workflows and action versions
- management-cluster manifests for the AWS and local-host profiles
- workload manifests and per-cluster overlays
- HelmRelease, HelmChartProxy, and other chart references
- container image references in Kubernetes resources
- CAPI/CAPD node images and cluster classes
- airgap image inventories, Zarf configuration, build scripts, staging scripts, and documentation

For example, the current airgap path couples the Zarf pin, `HOST_IMAGES`, `KIND_NODE_IMAGE`, CAPD `customImage` values, and workflow tool versions. Similar duplication exists outside airgap. Partial upgrades can therefore produce mismatched build, bootstrap, reconciliation, and deployment inputs. Comments can document these relationships but cannot enforce them.

## Goal

Manage project dependency versions from one authoritative catalog and derive or validate every consumer from it. A dependency bump should require changing its version in exactly one place, regardless of whether it is consumed by AWS, local-host, workload, CI, or airgap paths.

## Acceptance criteria

- Inventory repository-wide version pins and identify which occurrences represent the same logical dependency.
- Define one authoritative, human-reviewable dependency catalog.
- Make mise configuration, CI workflows, Kubernetes manifests, Helm references, cluster images, scripts, and airgap packaging consume values derived from that catalog where practical.
- Eliminate duplicated literal pins, or add validation that fails when a consumer cannot be generated and drifts from the catalog.
- Support intentional profile-specific or architecture-specific version differences explicitly rather than treating them as accidental drift.
- Keep generated output deterministic and suitable for rendered GitOps review.
- Preserve existing AWS, local-host, workload, and airgap behavior during migration.
- Document the dependency update procedure.
- Extend `mise run validate` or CI to verify that generated/derived dependency references are current.

## Suggested delivery

Treat this as an incremental migration: first inventory and choose the catalog format, then move one dependency family at a time while adding drift checks. Avoid a single repository-wide generated rewrite that is difficult to review.

shrinedogg:
Merged once CI was able to run. https://github.com/polarsquad/knr-ops/pull/69#pullrequestreview-5032633766

PR #69, opened 2026-08-26
feat: dependency version catalog with drift checks

shrinedogg:
Part of #62.

## What changed

- `deps/versions.toml`: new authoritative dependency version catalog. Full
  `[versions]` inventory (mise tools, k8s nodes/clusters, cluster-api family,
  flux family, platform components, ACK controllers, EKS addons, GitHub
  actions, and explicitly recorded "latest" decisions) plus composed
  `[images]` entries. Phase 1 adds `[[slot]]` entries for the mise tool
  family: go, kubectl, kind, helm, clusterctl, zarf, clusterawsadm (in
  mise.aws.toml), and mise (min_version).
- `deps/scripts/check.py`: slot drift checker (stdlib only, Python 3.11+
  tomllib). Validates each slot pattern and count against its file;
  `--forbid` phase 1 fails if the mise-family literals appear outside the
  catalog, their slotted files, docs/, or a documented exclusion list of
  consumers slotted in later PRs; `--markdown` emits the docs inventory table.
- `deps/scripts/catalog.env.sh`: bash + awk emitter exporting the catalog as
  `CATALOG_VERSIONS_*` / `CATALOG_IMAGES_*` env vars with `{key}`
  interpolation; no Python needed on gap-side hosts.
- `docs/dependencies.md`: purpose, generated inventory table, update
  procedure, intentional-difference list, exclusions list, validation vs
  generation decision per consumer class.
- `mise.toml`: adds `python = "3.13"` to `[tools]`; new `deps-check` and
  `deps-render` tasks; `validate` now runs deps-check first.
- `.github/workflows/validate.yml`: new "Dependency drift" step running the
  same check on every PR and push.
- `AGENTS.md`, `README.md`: point contributors at the catalog and doc page.
- Negative test: with kind bumped to 0.33 in mise.toml, check.py fails with
  `FAIL slot: file=mise.toml key=kind value=0.32 expected=1 actual=0`.
  (Note: running it through `mise run deps-check` in that state fails even
  earlier at mise's kind@0.33 install step, since no such release exists.)

## Verification

- `bash -n deps/scripts/catalog.env.sh`: pass.
- `source deps/scripts/catalog.env.sh && echo "$CATALOG_VERSIONS_ZARF"`:
  prints `0.83.0` (also verified `CATALOG_IMAGES_WORKLOAD_NODE` =
  `kindest/node:v1.35.0`).
- `mise run deps-check`: passes ("all slots match the catalog").
- Negative test (kind 0.33): fails naming the kind slot, then reverted.
- `mise run validate`: passes end to end (deps-check + bash -n + all
  kustomize overlays).


shrinedogg (commented):
## Code Review Summary

**Verdict: Comment - catalog design and checks verified; two warnings to address before undrafting**

Verified locally (head 6ce54d4):

- `python3 deps/scripts/check.py` and `--forbid` both pass; `bash -n deps/scripts/catalog.env.sh` passes; sourcing it exports the expected values including interpolated images (CATALOG_IMAGES_MGMT_NODE=kindest/node:v1.36.1, CATALOG_IMAGES_WORKLOAD_NODE=kindest/node:v1.35.0).
- Negative slot test reproduced exactly: bumping kind to 0.33 in mise.toml fails with `FAIL slot: file=mise.toml key=kind value=0.32 expected=1 actual=0`, exit 1.
- Forbid pass probed with planted literals: a bare `0.83.0` in a non-excluded path fails naming the zarf key; the same literal in docs/ is allowed; `10.83.0` is correctly ignored (segment boundary works).
- The docs inventory rows are content-identical to the `--markdown` output (verified by sorting both sides).
- `mise run deps-check` and full `mise run validate` pass end to end (deps-check, bash -n, all kustomize overlays).
- Environment note: macOS system python3 is 3.9 (no tomllib), so the `python = "3.13"` tool pin is load-bearing for the bare `python3` task invocations; under `mise run` the shim resolves correctly.

Warning: `mise run deps-render` ships as a task but its script does not exist in this PR. Running it fails with `python3: can't open file '.../deps/scripts/render.py': [Errno 2] No such file or directory`. docs/dependencies.md step 2 of the update procedure also sends contributors to `mise run deps-render` for generated consumers. Suggest dropping the task (and softening the docs step) until render.py lands, or landing render.py here.

Warning: docs/dependencies.md says the inventory table is "Generated by `python3 deps/scripts/check.py --markdown`", but the committed table is alphabetical while the generator emits catalog order. The rows are content-identical (verified), but anyone regenerating per the docs gets a reordered table, so the table will churn on every update. One-line fix: sort in run_markdown.

Note: the forbid matcher flags hyphen-suffixed variants: `0.83.0-rc1` matches zarf's `0.83.0` (segment boundaries work; `10.83.0` is ignored). Fail-loud is the safe direction, but a one-line mention in docs/dependencies.md would set expectations for prerelease pins.

Note: the validate workflow run for this head (32984185273) has been QUEUED for over 3 hours with zero jobs started, while earlier runs of the same workflow completed normally (Aug 24) and konflate passed in 18s. The Dependency drift step is wired correctly (`on: pull_request`, first step after checkout), but its first live run has not demonstrated a pass. Suggest cancelling and re-running before undrafting.

Suggestion (robustness, non-blocking): catalog.env.sh's awk parses any in-section line containing `=`, so a future comment containing `=` inside [versions]/[images] would emit a bogus variable; a `/^[ \t]*#/ { next }` guard fixes it. And check.py tracebacks on a missing slot file instead of failing cleanly.

Looks good: the catalog covers the full inventory including the recorded-latest decisions, the slot patterns fail loud on count mismatches in both directions, the forbid boundary matching avoids the obvious partial-collision traps, and AGENTS.md, README, and docs all describe the new workflow accurately.

shrinedogg (commented):
Review responses on commit 8da9faa:

Warning 1 (deps-render): Task removed from mise.toml; update procedure in docs/dependencies.md now lists slots only and notes generated consumers land in follow-up PRs.

Warning 2 (markdown churn): run_markdown now sorts keys; the committed inventory table was regenerated from `--markdown` output, so regeneration is a no-op.

Suggestion (awk comment guard): catalog.env.sh skips lines starting with # inside [versions]/[images].

Suggestion (missing slot file): check.py reports `FAIL slot: file=... key=... (file not found)` instead of tracebacking.

Note (prerelease matching): documented under "Forbid matcher scope" in docs/dependencies.md: segment-based, hyphen-suffixed variants match the base key and fail loudly; give prereleases their own key.

Note (queued CI run): will cancel and re-run before undrafting.

Gates re-run after fixes: mise run deps-check pass, source catalog.env.sh exports expected values, mise run validate exit 0.

shrinedogg (commented):
## Re-review Summary (commits 8da9faa..4b5b491)

**Verdict: Comment - all review items verifiably closed; the CI gate remains undemonstrated for infrastructure reasons**

Delta scope checked first: b68be0c plus the two chore commits touch only .gitignore, catalog.env.sh, check.py, docs/dependencies.md, and mise.toml. Nothing unrelated crept in; 8da9faa removes the stray .hermes/ plans and gitignores the directory.

| Item from the previous review | Status | Evidence |
|---|---|---|
| Warning: deps-render dead task | Fixed | Task removed from mise.toml (`mise tasks` shows only deps-check); the update procedure in docs/dependencies.md no longer references it and defers generated consumers to follow-up PRs |
| Warning: --markdown order churn | Fixed | run_markdown now sorts keys; regenerating and diffing against the committed inventory table is an exact no-op, not just modulo order |
| Suggestion: awk comment guard | Fixed | Planted `# note with = sign` inside [versions]: no bogus variable is emitted and CATALOG_VERSIONS_GO still resolves correctly |
| Suggestion: missing slot file traceback | Fixed | Pointed a slot at a nonexistent file in a scratch catalog: clean `FAIL slot: file=... key=... (file not found)`, exit 1, no traceback |
| Note: prerelease matching undocumented | Fixed | New "Forbid matcher scope" section documents segment matching and advises giving prereleases their own catalog key |
| Note: queued CI run | Open | See below |

Gates re-run at head 4b5b491: `check.py` and `--forbid` pass, `bash -n deps/scripts/catalog.env.sh` passes, `mise run deps-check` passes, full `mise run validate` exits 0.

CI caveat: the original validate run (32984185273) has been queued for over 3.5 hours as GHA is currently down. Will merge when the service returns.

Issue #74, opened 2026-08-27
Automate dependency updates with Renovate and retire the central version catalog


shrinedogg:
## Problem

knr-ops carries a growing set of dependency references across mise, CI, Kubernetes manifests, Flux resources, Helm charts, scripts, and the air-gap image inventory. Updates are found through manual monitoring, routine version edits are repetitive, and there is no audit trail of which updates passed, failed, merged, or were postponed. The central version catalog introduced in #62 was intended to make one location authoritative, but native consumers cannot read it directly, so their original declarations remain and must be kept in sync through custom checks, parsers, and rules.

## Goal

Automate dependency update discovery and pull-request preparation. Keep each version declaration in the native file consumed by its tool or controller, and use Renovate to discover updates and prepare changes across those files. Automation proposes changes for review rather than committing directly to main.

## Proposed scope

- Automate dependency update discovery and pull-request generation with Renovate.
- Retire the central catalog and literal-slot synchronization machinery.
- Keep infrastructure update approval and merge decisions manual.
- Recognize cross-file occurrences of the same logical dependency and update them together in one pull request:
 - rely on built-in Renovate managers and shared datasources/package identities where possible,
 - add explicit annotations or narrowly scoped custom managers only for stable, nonstandard references (e.g. specific Zarf or shell-script values),
 - add targeted validation for relationships that must remain equal or complete, such as an updated runtime image also appearing in the air-gap inventory,
 - keep intentionally independent versions as separate dependency identities; matching version strings alone must not couple unrelated consumers.
- Pin exact releases or immutable digests and remove `latest` where reproducibility matters.

## Operating model

- Run update discovery on a weekly schedule.
- Open pull requests rather than committing directly to main.
- Group only dependencies with a genuine compatibility or atomic-delivery relationship.
- Keep management-cluster, CAPD workload, and EKS Kubernetes versions independent unless a documented constraint requires coordination.
- Run `mise run validate` and rendered GitOps review on every update pull request.
- Regenerate and verify the air-gap image inventory whenever packaged charts or images change.
- Avoid broad auto-merge; consider it only later for demonstrably low-risk patch updates with reliable validation.

## Acceptance criteria

- Renovate is configured and discovers updates across mise, Flux, Helm, Kubernetes and container images, GitHub Actions, Kustomize, and scripts.
- A logical dependency referenced in multiple native files is updated in a single grouped pull request.
- The central catalog and its synchronization checks are removed with no loss of coverage.
[...]

PR #76, opened 2026-08-27
feat: add Renovate dependency update automation

shrinedogg:
## Summary

Implements the first phase of #74: automated dependency update discovery and pull-request preparation via Renovate, running as a scheduled GitHub Actions workflow.

- `renovate.json5`: flux / kubernetes / helm-values managers scoped to `mgmt/` and `workload/`, GitHub Actions action pins, explicit per-tool mise custom managers (native mise manager disabled for deterministic dep resolution), clusterctl provider manifests (CAPI, CAPD, CAPA, CAAPH), kindest/node image pins, airgap `images.txt` and `zarf.yaml` image refs.
- `.github/workflows/renovate.yml`: weekly Monday 05:00 UTC run plus `workflow_dispatch` with dry-run default; `renovatebot/github-action@v42`; `permissions: {}`; concurrency group.
- Global `automerge: false`; merges stay manual per #74. Flux and cluster-api families grouped; kindest/node versions kept independent. EKS addon versions left unmanaged (no public datasource), noted in config comments.

Config validated with `renovate-config-validator`.

Stacked PRs to follow: air-gap inventory consistency check in CI, then catalog retirement.

Refs #74

## Required follow-up (repo admin)

- Add repo secret `RENOVATE_TOKEN` (fine-grained PAT, contents and pull request read/write on this repo only).
- Trigger the workflow with dry-run enabled to verify dependency extraction.

### DIFF
--- a/.github/workflows/renovate.yml
+++ b/.github/workflows/renovate.yml
@@ -1,0 +1,46 @@
+name: renovate
+
+on:
+  schedule:
+    - cron: "0 5 * * 1" # Mondays 05:00 UTC = 08:00 EEST
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "Dry run (no PRs created)"
+        type: boolean
+        default: true
+
+concurrency:
+  group: renovate
+  cancel-in-progress: false
+
+permissions: {}
+
+jobs:
+  renovate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Renovate dry-run config check
+        if: github.event_name == 'workflow_dispatch'
+        uses: renovatebot/github-action@v42
+        with:
+          configurationFile: renovate.json5
+          token: ${{ secrets.RENOVATE_TOKEN }}
+        env:
+          RENOVATE_DRY_RUN: ${{ inputs.dry_run && 'full' || 'false' }}
+          RENOVATE_REPO: ${{ github.repository }}
+          LOG_LEVEL: debug
+
+      - name: Renovate scheduled run
+        if: github.event_name == 'schedule'
+        uses: renovatebot/github-action@v42
+        with:
+          configurationFile: renovate.json5
+          token: ${{ secrets.RENOVATE_TOKEN }}
+        env:
+          RENOVATE_REPO: ${{ github.repository }}
+          RENOVATE_ONBOARDING: "false"
+          LOG_LEVEL: info
--- a/.github/workflows/renovate.yml
+++ b/.github/workflows/renovate.yml
@@ -25,7 +25,7 @@ jobs:
 
       - name: Renovate dry-run config check
         if: github.event_name == 'workflow_dispatch'
-        uses: renovatebot/github-action@v42
+        uses: renovatebot/github-action@v46.2.4
         with:
           configurationFile: renovate.json5
           token: ${{ secrets.RENOVATE_TOKEN }}
@@ -36,7 +36,7 @@ jobs:
 
       - name: Renovate scheduled run
         if: github.event_name == 'schedule'
-        uses: renovatebot/github-action@v42
+        uses: renovatebot/github-action@v46.2.4
         with:
           configurationFile: renovate.json5
           token: ${{ secrets.RENOVATE_TOKEN }}
--- a/mise.aws.toml
+++ b/mise.aws.toml
@@ -2,7 +2,7 @@
 KNR_OPS_PROFILE = "aws"
 
 [tools]
-aws-cli = "latest"
+aws-cli = "2.36.32"
 clusterawsadm = { version = "2.13.0", binary = "clusterawsadm" }
 
 [tasks.aws-bootstrap]
--- a/mise.toml
+++ b/mise.toml
@@ -7,9 +7,9 @@ kubectl = "1.35"
 kind = "0.32"
 helm = "4.2"
 clusterctl = "1.14.0"
-flux2 = "latest"
-sops = "latest"
-age = "latest"
+flux2 = "2.9.4"
+sops = "3.13.3"
+age = "1.3.1"
 
 # Zarf CLI (airgap prototype). The mise github backend downloads the raw
 # release asset and keeps its original filename, so we select the native
--- a/renovate.json5
+++ b/renovate.json5
@@ -275,9 +275,10 @@
       "separateMajorMinor": true,
       "groupName": null
     },
-    // Floating "latest" pins (flux2, sops, age, aws-cli in mise; the zarf
-    // local-registry image) are intentionally untracked: Renovate skips a
-    // "latest" currentValue as invalid before any rule can pin it, so they
-    // stay floating by decision until pinned by hand once.
+    // flux2, sops, age, and aws-cli were previously floating "latest" pins;
+    // they are now pinned to concrete versions in mise.toml / mise.aws.toml
+    // so Renovate tracks them normally. The zarf local-registry image keeps
+    // its bare "latest" tag: it is built locally per run and has no
+    // comparable version.
   ]
 }
--- a/renovate.json5
+++ b/renovate.json5
@@ -51,6 +51,7 @@
       "matchStrings": ["go = \"(?<currentValue>[^\"]+)\""],
       "depNameTemplate": "golang",
       "datasourceTemplate": "golang-version",
+      "versioningTemplate": "loose",
       "autoReplaceStringTemplate": "go = \"{{newValue}}\""
     },
     {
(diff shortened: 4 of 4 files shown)


shrinedogg (commented):
## Code Review Summary

**Verdict: Comment - intent and coverage are right, but a local dry-run shows four functional problems; the workflow cannot run as written**

Verified locally (head 312d820). I ran `renovate-config-validator` (passes, matching the PR) and a full local dry-run (`renovate --platform=local --dry-run=full`, CLI 42.99.0), plus per-pattern probes against the real repo files.

Verified working:

- Native managers extract cleanly: flux 8 files/11 deps, kubernetes 69/92, helm-values 1/2, github-actions 4/26.
- Schedule coherence: cron 05:00 UTC Monday lands at/after 06:00 Europe/Helsinki year-round, matching the config's own schedule gate.
- Workflow hardening as described: `permissions: {}`, dry-run default true on dispatch, concurrency group, onboarding disabled for scheduled runs.
- Every custom-manager pattern matches its intended content once line semantics are assumed (probed in Python): 12/12 mise pins exactly one match each, provider manifests (capi x3, capa, capd, caaph on both profiles), images.txt 45/45 entries, zarf.yaml 15/15 image entries with zero unmatched.
- EKS addons left unmanaged with the no-datasource rationale recorded, consistent with the catalog note.
- konflate rendered-diff check green (no mgmt/workload changes, as expected).

Warning: `renovatebot/github-action@v42` does not resolve. The repo publishes exact-version tags only (v42.0.0 up to v46.2.4); there is no floating `v42` tag or branch (`commits/v42` returns 422). Both steps fail at action resolution, so the workflow cannot run at all. Pin an exact tag (v46.2.4 is current; exact pins also fit the repo's catalog philosophy, and the github-actions manager will bump it).

Warning: 17 of 18 custom managers extract zero dependencies. Renovate compiles matchStrings with only the `g` flag (verified in the shipped source: `regEx(matchString, "g")` in the regex manager's strategies.js), so every `^`/`$`-anchored pattern matches nothing in a multi-line file. The dry-run extraction stats prove it: the regex manager found 8 deps in 4 files, all kindest/node, the only unanchored pattern, and logged "No dependencies found in file" for mise.toml and mise.aws.toml. The mise pins, clusterctl provider manifests, images.txt, and zarf.yaml managers are all silently dead. Fix: drop the `^`/`$` anchors and keep enough literal context (e.g. `go = "(?<currentValue>[^"]+)"`), or add explicit `\n` boundaries.

Warning: the clusterctl manager's depName `fluxcd/clusterctl` returns 404; no such repo exists. clusterctl ships from kubernetes-sigs/cluster-api (mise's aqua backend resolves clusterctl to exactly that repo; latest release v1.14.0).

Warning: the go and python managers use github-releases against repos that publish zero GitHub releases (verified: golang/go and python/cpython release lists are empty; tags only). Once the anchors are fixed these two managers still produce nothing. For go use the golang-version datasource; for python/cpython use github-tags with the existing `^v` extractVersion.

Note (sequencing with #78): once extraction works, Renovate's mise/zarf/clusterctl bump PRs will fail the Dependency drift step from #69 until the catalog retirement lands. Consider merging #78 before the first scheduled run, or expect red Renovate PRs in the meantime.

Note: renovate-config-validator passes this config (verified) but catches none of the above; the PR's own "trigger dry-run first" follow-up is the right gate, and this review's dry-run is a preview of what it will report. Also note the local run warned "GitHub token is required for some dependencies" for the github-actions manager; in the workflow RENOVATE_TOKEN covers this, so no action needed.

Looks good (once runnable): the manager inventory matches the repo's real version surface, the grouping rules (flux family, cluster-api family, kindest/node independence) match the catalog's intentional differences, and global automerge:false keeps merges manual per #74.

shrinedogg (commented):
## Re-review Summary (commit ce19041)

**Verdict: Comment - three of four warnings cleanly fixed, but the rewritten airgap patterns have two new defects**

Re-verified with a full local dry-run (renovate CLI 42.99.0, `--platform=local --dry-run=full`) plus byte-level pattern checks.

Fixed, verified:

- Action ref: both steps now pin renovatebot/github-action@v46.2.4, which exists and is the current release tag. The workflow can start.
- Anchors: regex-manager extraction went from 8 deps in 4 files to 52 deps in 24 files. The unanchored mise patterns extract 12/12 pins, the provider patterns 10/10 version fields (capi x3 on both profiles, capa, capd, caaph on both profiles), kindest/node 8/8 across cluster-class and the three airgap scripts. Zero "Failed to look up" and zero 404s in the run.
- clusterctl: now kubernetes-sigs/cluster-api, matching the mise aqua backend.
- go now uses the golang-version datasource and python/cpython uses github-tags; both are the right datasources (see the notes for two residual issues).

New Warning: the images.txt and zarf.yaml patterns are under-escaped. In renovate.json5 their character classes contain a single `\s`; JSON5 treats `\s` as an unrecognized escape and drops the backslash (verified by parsing the file with the same json5 package renovate ships: the loaded pattern is `[^s@]+`, not `[^\s@]+`). The resulting class excludes the literal letter "s" and allows whitespace and newlines. Two visible consequences in the dry-run: the zarf.yaml manager captured a multi-line currentValue for ghcr.io/fluxcd/kustomize-controller that stops exactly before the "s" in "source-controller" (that dep is skipped as invalid-value, and the two entries swallowed inside the value never become deps: 12 of 15 entries extracted), and the images.txt manager extracted only 10 of 39 unique image refs. The provider patterns escaped `\s{2}` correctly and extract 10/10, so the fix is to double the backslashes the same way in both airgap patterns.

New Warning: even with correct classes the images.txt pattern skips every second line. The pattern starts with `(?:^|\n)` and ends with a consumed `\n`, so each match eats the newline the next match needs as its anchor (probed: 24 of 45 lines match). Fix: drop the trailing `\n` and let `(?:#.*)?` end the match; `.*` stops at end of line naturally, and this stays RE2-compatible (no lookahead).

Note: the golang-version datasource rejects the current pin `go = "1.26"` as invalid-value (default semver versioning wants three version parts). Either pin `go = "1.26.0"` in mise.toml or set a lenient versioningTemplate on that manager, or the go manager produces nothing despite the correct datasource.

Note: the four `ubuntu:latest` image refs in the airgap files skip with invalid-version (unversioned tags are not comparable). If tracking them matters they need pinDigests/digest handling; if not, a config comment would make the choice visible. The floating mise `latest` pins (flux2, sops, age, aws-cli) are likely in the same boat once lookups run with a token; locally they were skipped as github-token-required, which is a local-mode artifact only, the workflow's RENOVATE_TOKEN covers it.

Everything else from the first review stands: native managers extract cleanly (flux 11, kubernetes 92, helm-values 2, github-actions 26), schedule coherence holds, and the workflow hardening is unchanged.

shrinedogg (commented):
## Re-review Summary (commit 32f03ae)

**Verdict: Comment - all four previous items verified fixed; two new repo-name lookups fail, one dead rule**

This time the dry-run (renovate CLI 42.99.0, `--platform=local --dry-run=full`) ran with a GitHub token active, so every datasource lookup executed for real rather than skipping.

Fixed, verified:

- Escaping: byte-verified by parsing renovate.json5 with the same json5 package renovate ships; the images.txt and zarf.yaml classes now load as `[^\s:@#/]` and `[^\s@]` correctly. Regex-manager extraction went from 52 to 90 deps: images.txt 45/45 lines, zarf.yaml 15/15 entries, on top of mise 12/12, providers 10/10, kindest/node 8/8.
- Alternation: the trailing `\n` is dropped from both patterns; no lookahead, RE2-compatible.
- golang: `versioningTemplate: "loose"` works; the dry-run computed a real update, go 1.26 -> 1.27.0.
- The config comment documenting the intentionally untracked localhost `latest` tag is accurate.

New Warning: the clusterawsadm manager's depName `kubernetes-sigs/clusterawsadm` returns 404 ("Could not resolve to a Repository"). clusterawsadm ships from `kubernetes-sigs/cluster-api-provider-aws` (confirmed via the GitHub API and via `mise registry clusterawsadm`, which resolves to that exact repo). Same bug class as the clusterctl depName from the first review.

New Warning: the CAAPH manager's depName `kubernetes-sigs/cluster-api-addon-helm` returns 404. The real repo is `kubernetes-sigs/cluster-api-addon-provider-helm` (verified; its latest release is v0.6.4, matching the catalog value).

New Note: the "Pin floating latest mise values to concrete release versions" packageRule can never fire as configured. With the token active the dry-run shows all four floating pins (fluxcd/flux2, getsops/sops, FiloSottile/age, aws/aws-cli) skipping as invalid-value because currentValue `latest` is not comparable; the skip happens before the pin rule can act. Either pin those four to concrete versions once by hand (after which Renovate tracks them normally), or drop the rule and document that they intentionally stay floating. Same applies to the `ubuntu:latest` image refs, which skip as invalid-version.

Minor, pre-existing, not from this PR: the native kubernetes manager reads four ACK `Role` manifests' apiVersion strings as versions and marks them invalid-value; harmless noise in the log.

The two lookup 404s were invisible in earlier runs because the github-token-required skip fired first; they will fail every scheduled run until the depNames are corrected. Everything else stands: full extraction (regex 90 deps, total 221), zero config errors, workflow hardening unchanged.

shrinedogg (commented):
Re-review of `58f7179` + `5d7e932`:

- Both depName corrections verified against the GitHub API: cluster-api-addon-provider-helm latest v0.6.4 and cluster-api-provider-aws latest v2.13.0, matching the pinned versions. The clusterawsadm regex still extracts the inline-table pin exactly once.
- clusterawsadm and the CAPA provider now intentionally share the depName kubernetes-sigs/cluster-api-provider-aws, so their updates arrive as one grouped cluster-api PR; desirable since clusterawsadm must stay version-aligned with CAPA.
- Floating-latest pins resolved by pinning to concrete versions (flux2 2.9.4, sops 3.13.3, age 1.3.1, aws-cli 2.36.32) rather than by rule; all four now extract as tracked deps with correct currentValues. Local dry-run skips them as github-token-required only, which the workflow token covers.
- Remaining invalid-value skips are the known noise: four ACK Role apiVersion strings (native kubernetes manager) and the zarf local-registry latest image, documented in config comments.
- Validator passes, mise run validate passes, CI green.

No further changes requested.

PR #77, opened 2026-08-27
feat: airgap inventory consistency check in CI

shrinedogg:
## Summary

Second phase of #74 (stacked on #76): guarantees the air-gap image inventory stays complete when Renovate proposes updates.

- `airgap/scripts/check-inventory.py`: builds every `mgmt/local-host` overlay, extracts container image refs, verifies each repository appears in `airgap/images.txt`. Honors the inventory's `[M]/[H]/[W]` category conventions: `[H]` host-daemon images are skipped as not agent-rewriteable. Exits 1 with a missing-repos table on drift; `--list` mode for inspection.
- `airgap/scripts/test_check_inventory.py`: 16 unit tests for the matching and parsing logic, stdlib unittest only.
- `.github/workflows/validate.yml`: new `airgap-inventory` job runs the check in CI on every PR, so a Renovate update that bumps a rendered image without updating the inventory fails validation.

Note: most local-host components (Flux, cert-manager, CAPI controllers) install via Helm charts and do not surface in kustomize render output; the check covers everything that actually renders inline. Chart-pulled images remain covered by the `airgap/images.txt` and `zarf.yaml` Renovate managers in #76, which update the inventory itself.

Refs #74

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -36,6 +36,24 @@ jobs:
           done
           exit $fail
 
+  airgap-inventory:
+    name: Airgap inventory check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Install kustomize
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          curl -s -H "Authorization: token *** github.token }}" "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
+          sudo mv kustomize /usr/local/bin/
+          kustomize version
+
+      - name: Check airgap image inventory
+        run: python3 airgap/scripts/check-inventory.py
+
   yaml-lint:
     name: Lint YAML
     runs-on: ubuntu-latest
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -47,9 +47,12 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
         run: |
-          curl -s -H "Authorization: token *** github.token }}" "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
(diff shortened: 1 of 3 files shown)


shrinedogg (commented):
## Code Review Summary

**Verdict: Comment - the check works and is well-scoped; the red CI is a transient install failure, not the check**

Verified locally (head 2650666, diffed against the feat/renovate stacking base: exactly 3 files, +299):

- 16/16 unit tests pass (stdlib unittest, no deps).
- The check passes locally: 2 rendered image repositories under mgmt/local-host, both present in images.txt.
- Negative test reproduced: removing kindnetd from images.txt fails with exit 1 and a table naming docker.io/kindest/kindnetd; restoring turns it green.
- Inventory parsing verified against the real file: 25 repos under [M], 3 under [H] (docker.io/library/registry, kindest/haproxy, kindest/node), and every [W] line duplicates an [M] entry, so the first-wins setdefault is safe under the current file ordering. The [H] skip is correct for this profile: CAPD workload nodes are host-docker containers too, so the rendered kindest/node refs in cluster-class.yaml are rightly not required to be agent-rewriteable.

Note (the failing CI job): the "Airgap inventory check" job never reached the check. Its kustomize-install step piped a 404 body into bash ("404:: command not found", exit 127) at 05:10; the identical step failed the same way on #78's run at 05:13, and all five validate runs since (including main at 10:16) are green. This was a transient raw.githubusercontent failure in a shared pre-existing step, not a problem with this PR. Re-run the failed job before merge; the flakiness itself (curl|bash with no retry) is worth hardening separately.

Note (measured coverage): at this head the check enforces exactly one repository (docker.io/kindest/kindnetd); the only other rendered repo, kindest/node, is [H]-skipped by design. The PR's note about chart-pulled images not rendering inline is accurate, and the check still earns its place as the guard for kindnetd plus anything future overlays render inline. Reviewers just should not read "inventory consistency" as broader than that today.

Suggestion: the PR ships 16 unit tests but no CI job runs them. Adding `python3 -m unittest airgap.scripts.test_check_inventory` as a step in the new job (stdlib, no pytest needed) would gate them alongside the check.

Looks good: stdlib-only, fail-loud on kustomize build errors, category conventions implemented faithfully, honest coverage note in the PR body.

PR #83, opened 2026-08-27
fix: point clusterawsadm manager at cluster-api-provider-aws (refs #74)

shrinedogg:
## Problem

The clusterawsadm custom manager in renovate.json5 uses depName `kubernetes-sigs/clusterawsadm`, which does not exist (GitHub API 404). clusterawsadm releases ship from `kubernetes-sigs/cluster-api-provider-aws` — the same repo mise resolves the tool to (`mise registry clusterawsadm` -> `github:kubernetes-sigs/cluster-api-provider-aws`). Commit 58f7179's title said it corrected both CAAPH and clusterawsadm depNames, but its diff only changed CAAPH, so every Renovate run would log a failed lookup for clusterawsadm.

## Fix

One line: depNameTemplate now `kubernetes-sigs/cluster-api-provider-aws`. The dep then also joins the existing cluster-api package group (clusterawsadm tracks CAPA per the catalog note), which already lists that repo.

## Verification

- `gh api repos/kubernetes-sigs/cluster-api-provider-aws/releases/latest` -> v2.13.0 (matches the current mise.aws.toml pin; `^v` extractVersion fits the tag format).
- Full local dry-run (renovate CLI 42.99.0, `--platform=local --dry-run=full`, token active): **zero failed lookups** across the whole run (previously 8, driven by the two 404 repos), and the dep resolves with currentValue 2.13.0 against latest v2.13.0 producing no spurious update PR.

Refs #74

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -159,7 +159,7 @@
       "description": "mise: clusterawsadm pin (inline table form)",
       "managerFilePatterns": ["/(^|/)mise\\.aws\\.toml$/"],
       "matchStrings": ["clusterawsadm = \\{ version = \"(?<currentValue>[^\"]+)\""],
-      "depNameTemplate": "kubernetes-sigs/clusterawsadm",
+      "depNameTemplate": "kubernetes-sigs/cluster-api-provider-aws",
       "datasourceTemplate": "github-releases",
       "extractVersionTemplate": "^v(?<version>.+)",
       "autoReplaceStringTemplate": "clusterawsadm = { version = \"{{newValue}}\""


lmilbaum:
Would it make sense to use the mise manager instead of regex?

shrinedogg:
Good question! Tested it before landing on regex. I re-ran the comparison today to be sure (native mise manager enabled, the 12 mise regex managers removed, full dry-run on renovate CLI 42.99.0):

- The native manager resolves 10 of the 12 tools correctly, including the backend resolution we got wrong by hand: clusterctl -> kubernetes-sigs/cluster-api and clusterawsadm -> kubernetes-sigs/cluster-api-provider-aws come out right with no manual depName. That is a real advantage and the main reason the migration is attractive.
- Two tools break silently: zarf is extracted but skipped as unsupported-datasource (the [tools."github:zarf-dev/zarf"] backend section form), and age (aqua backend) does not appear in the extraction at all. No error either way; those tools would just never get updates.

So: yes, worth migrating, but as a hybrid. Enable the native mise manager for the 10 tools it verifiably covers and keep the two regex managers for zarf and age until the mise manager learns those backends. I would rather land that as a follow-up PR with a before/after dry-run comparison than grow this one; #83 is a one-line lookup fix that stands on its own.

PR #84, opened 2026-08-28
feat: management cluster definitions for the CAPI pivot (aws + local-host)

shrinedogg:
## What

Declares the management clusters that issue #79 pivots into, one per profile, following each profile's existing cluster pattern exactly:

- **aws**: `mgmt/aws/clusters/eu-north-1/management/` — EKS-based (`AWSManagedControlPlane` + `AWSManagedCluster` + `AWSManagedMachinePool`), 2x t4g.medium (min 2 / max 3, AZs a/b/c), public endpoint, k8s 1.34.6. The `AWSManagedCluster` infra resource joins the central list in `mgmt/aws/infrastructure/aws-clusters-resources/` with the literal name `eu-north-1-management`.
- **local-host**: `mgmt/local-host/clusters/management/` — CAPD/Docker via a new ClusterClass `docker-management` (DevClusterTemplate + KubeadmControlPlaneTemplate + DevMachineTemplates, `kindest/node:v1.35.0`), Cluster `local-management`, 1 CP + 1 worker, wired with a `docker-management-cluster` Flux Kustomization.

## Why

First PR of the #79 stack: the definitions exist in Git so the providers running in the kind bootstrap cluster can create these clusters, and `clusterctl move` (pivot tooling, next PR) can adopt them.

## Details worth review

- **Label wiring (the CRS trap)**: the aws management Cluster carries `region: eu-north-1` but deliberately **no** `fluxcd: enabled`; the local-host one carries `profile: local-host` (so the kindnet CRS supplies the CNI — CAPD clusters ship none) but also no `fluxcd: enabled`. Either label would trigger a ClusterResourceSet that deploys a workload FluxInstance into the management cluster.
- CAPD places node/lb containers on the docker `kind` network by default (no API field), which is what makes `knr-registry` reachable from the management cluster nodes; the network survives `kind delete cluster` because the registry stays attached. Commented in the ClusterClass.
- No new images; airgap inventory check stays green (`check-inventory.py` verified locally).

## Merge-timing note (aws)

Merging this starts provisioning the EKS management cluster from the current kind management cluster (~15-25 min, billing starts). The pivot runbook lands in the next PR; run it promptly after this merges.

## Validation

- `mise run validate` green (deps-check, bash -n, every kustomize overlay incl. the two new ones)
- namePrefix/nameReference rewriting verified in the rendered output (`infrastructureRef` stays the literal `eu-north-1-management`, matching the central infra list)
- `python3 airgap/scripts/check-inventory.py` — ok

[...]

Issue #79, opened 2026-08-27
Management cluster self-management via CAPI pivot from kind bootstrap


shrinedogg:
## Problem

The management cluster is a local kind cluster created by `bootstrap.sh` and it stays on an operator machine for the lifetime of the deployment. For the aws environment this means the control plane that owns the CAPA provider, the EKS workload cluster definitions under `mgmt/aws/clusters/`, the ACK controllers, and the Flux instance reconciling `mgmt/aws/` runs on a laptop. It depends on that machine being up, has no HA story, and is the one piece of the stack that is not itself managed through the GitOps flow it drives.

## Goal

Reach management cluster self-management: provision it as a real cluster on AWS through CAPI, carry out the CAPI pivot (move the CAPI control plane into it with `clusterctl move`), and reduce the local kind cluster to a disposable one-time bootstrap step.

## Proposed scope

Follow the standard CAPI pivot procedure, mapped onto knr-ops conventions:

1. Create the local bootstrap cluster using kind (the existing `mgmt` kind cluster from `bootstrap.sh`).
2. Initialize CAPI with the AWS infrastructure provider in the bootstrap cluster (`clusterctl init -i aws`, or the repo-consistent equivalent: install the CAPI operator HelmRelease from `mgmt/aws/infrastructure/capi-operator/` and let the provider CRs in `mgmt/aws/capi-providers/` (capi-system, capa-system, caaph-system) install core, AWS infrastructure, and addon providers, since the operator resolves well-known provider names from the same built-in registry clusterctl uses).
3. Ensure CAPA credentials and identity exist in the bootstrap cluster before cluster creation: `aws-credentials.sops.yaml` (`AWS_B64ENCODED_CREDENTIALS` from `mise run aws-credentials`) and the `AWSClusterControllerIdentity` from `mgmt/aws/infrastructure/aws-identity/`.
4. Add the management cluster definition following the existing pattern in `mgmt/aws/clusters/<region>/<env>/` (Cluster + AWSManagedControlPlane + AWSManagedCluster), deciding explicitly whether the management control plane is EKS-based like the workload clusters or self-managed on EC2.
5. Apply the management cluster definition and wait for it to be ready (Cluster conditions Available=True, `clusterctl describe cluster`).
6. Install the CAPI components in the new management cluster (same operator/provider CR path as step 2, or `clusterctl init` against the new cluster; required before the move).
[...]

PR #86, opened 2026-08-28
feat: CAPD management cluster definition for self-management (local-host)

shrinedogg:
## What

Declares the CAPD management cluster that issue #79 pivots into for the `local-host` environment: `mgmt/local-host/clusters/management/` with a new ClusterClass `docker-management` (DevClusterTemplate + KubeadmControlPlaneTemplate + DevMachineTemplates, `kindest/node:v1.35.0`), Cluster `local-management`, 1 CP + 1 worker, wired with a `docker-management-cluster` Flux Kustomization.

## Why

Definitions stage of the #79 self-management stack: with this in Git, the providers running in the kind bootstrap cluster can create the cluster, and `clusterctl move` (the CAPI pivot step, a later PR in the stack) can adopt it. After the pivot the management cluster reconciles its own definition from the OCI artifact.

## Details worth review

- Label wiring (the CRS trap): the management Cluster carries `profile: local-host` (the literal manifest label) so the kindnet ClusterResourceSet supplies the CNI (CAPD clusters ship none), but deliberately no `fluxcd: enabled`: that label would make the local-workload-flux-instance ClusterResourceSet deploy a workload FluxInstance into the management cluster.
- CAPD places node/lb containers on the docker `kind` network by default (no API field), which is what makes `knr-registry` reachable from the management cluster nodes; the network survives `kind delete cluster` because the registry stays attached. Commented in the ClusterClass.
- No new images (`kindest/node:v1.35.0` already in `airgap/images.txt`); inventory check stays green.

## Validation

- `mise run validate` green (deps-check, bash -n, every kustomize overlay incl. `mgmt/local-host/clusters/management`)
- `python3 airgap/scripts/check-inventory.py` ok

Supersedes the local-host half of #84 (split by environment per review). Refs #79.

### DIFF
--- a/mgmt/local-host/clusters/flux-ks.yaml
+++ b/mgmt/local-host/clusters/flux-ks.yaml
@@ -20,3 +20,29 @@ spec:
       kind: Cluster
       name: local-workload
       namespace: default
+---
+# The CAPD management cluster: created by CAPD in the kind bootstrap cluster,
+# moved into itself by `clusterctl move` (pivot.sh), and from then on
+# reconciled from the OCI artifact by the Flux instance running inside it.
+apiVersion: kustomize.toolkit.fluxcd.io/v1
+kind: Kustomization
+metadata:
+  name: docker-management-cluster
+  namespace: flux-system
+spec:
+  interval: 5m
+  retryInterval: 2m
+  timeout: 5m
+  prune: true
+  wait: false
+  sourceRef:
+    kind: OCIRepository
+    name: flux-system
+  path: ./mgmt/local-host/clusters/management
+  dependsOn:
+    - name: capd-system
+  healthChecks:
+    - apiVersion: cluster.x-k8s.io/v1beta2
+      kind: Cluster
+      name: local-management
+      namespace: default
--- a/mgmt/local-host/clusters/management/cluster-class.yaml
+++ b/mgmt/local-host/clusters/management/cluster-class.yaml
@@ -1,0 +1,121 @@
+apiVersion: cluster.x-k8s.io/v1beta2
+kind: ClusterClass
+metadata:
+  name: docker-management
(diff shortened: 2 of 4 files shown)


lmilbaum (approved):
Consistent environment terminology should eventually extend to functional metadata so that the architecture does not retain two names for the same concept. I noticed that `profile: local-host` is an actual label consumed by `ClusterResourceSet` selectors. Could we capture a follow-up change to migrate the label and all corresponding selectors to a qualified environment label, such as `knr-ops.polarsquad.com/environment: local-host`? I don’t think that repository-wide migration needs to be part of this PR.

shrinedogg:
Replying to @lmilbaum's review (https://github.com/polarsquad/knr-ops/pull/86#pullrequestreview-5048476863):

Agreed, and worth doing while the label surface is still small. I counted the migration scope on main: `profile: local-host` is a consumed label in four manifests (cluster.yaml Cluster labels, kindnet CRS selector, flux-operator and flux-instance clusterSelectors) plus two generated clusterSelectors in `airgap/scripts/build-config-artifact.sh`, so the render script and the manifests need to move in the same PR. The `profile: default` occurrences in `mgmt/aws/infrastructure/ack-controllers/helm.yaml` are an unrelated concept (the AWS SDK credentials profile key inside HelmRelease values), which is itself a small argument for the qualified name: a bare `profile` key already means something else nearby in this repo.

One sequencing note for the eventual PR: the selectors match live Cluster objects, so label and selector changes have to land atomically per environment. The docker Cluster and CAPD Clusters carry the label at creation, and after the #79 pivot the management Cluster reconciles its own definition, so a mismatched window would break CRS targeting on the cluster doing the reconciling.

Follow-up filed as #87 with this scope; this PR stays as-is.

PR #89, opened 2026-08-28
feat: pivot tooling for management-cluster self-management (aws + local-host)

shrinedogg:
Refs #79. Follow-up to #85/#86 (definitions merged): the tooling that moves the CAPI management inventory from the kind bootstrap cluster into the self-managed management cluster, on both environments.

## What

- **`pivot.sh`** — re-runnable pivot runbook, one script branching on environment like `bootstrap.sh`:
  1. preflight (context guard: kind must be the move source; aws: GitHub/age inputs; local-host: registry alive)
  2. wait for the management cluster (`eu-north-1-management` / `local-management`) — kubeconfig-secret poll, endpoint-agnostic across CAPA/CAPD
  3. export the target kubeconfig (`~/.kube/knr-ops-mgmt.yaml`, context `knr-ops-mgmt`; local-host: rewrite server to the lb container's published port, same as the workload kubeconfigs)
  4. install cert-manager + CAPI operator in the target imperatively at **catalog versions** (slots validated by deps-check), apply the provider CRs verbatim from Git (plain manifests only — capa-system's SOPS secret is Flux-only), create the CAPA `aws-credentials` secret directly (aws only; the identity is NOT pre-applied — it carries a move hook)
  5. suspend all Flux Kustomizations in kind (blocks the mid-move unpause race), `clusterctl move`, then **unpause the moved Clusters** on the target (the move pauses them; Git carries no `spec.paused`, so this must happen before Flux is seeded)
  6. seed Flux on the target (local-host republishes the OCI artifact first so it contains the management manifests), verify Kustomizations
  7. delete kind, guarded (clusters present on the target; local-host: registry serving)
- **`bootstrap-common.sh`** — shared preflight (`require_flux_env`, `detect_container_engine`) and `seed_flux` (operator, secrets, FluxInstance, Ready wait) extracted verbatim from `bootstrap.sh`; `pivot.sh` seeds the target exactly as bootstrap seeds kind. No behavior change to bootstrap.
- **`bootstrap.sh`** — execs `pivot.sh` by default on both environments (`BOOTSTRAP_PIVOT=0` opts out, pre-#79 behavior). Fresh installs pivot automatically once Flux creates the management cluster.
- **`mise.toml`** — `pivot` and `mgmt-kubeconfig` tasks; `validate` bash-n's the new scripts.
- **`deps/versions.toml`** — slots pinning `pivot.sh`'s `CERT_MANAGER_VERSION` / `CAPI_OPERATOR_VERSION` literals to the catalog.

## Why imperative provider install in the target

HelmReleases need Flux. The target has none pre-move, so pivot installs the same charts (same versions, same values) helm-wise, and applies the same provider CRs kubectl-wise; once Flux is seeded it adopts both without drift (identical Git content).

## Safety

- Move failure guidance built into the script: re-run, never delete moved infra objects on the target (providers would deprovision real infrastructure).
- `PIVOT_SKIP_DELETE=1` keeps kind for inspection.
- kind deletion only after all checks pass.

## Validation

- `bash -n` (plus macOS bash 3.2) on all scripts; `mise run validate` green (45 checks); deps-check green with the new slots.
- Preflight smoke-tested against the live kind (`local-management` not yet reconciled, clean error with next-step hints).
- CLI contracts verified against the pinned tools: `clusterctl move` flags, helm 4 `--kubeconfig`, multi-resource `kubectl wait`.

## Not in this PR

- `teardown.sh` self-hosted-mgmt awareness — next PR.
- Docs (operations/architecture) and AGENTS.md update — final PR. The AGENTS.md touch (pivot.sh/bootstrap-common.sh layout bullets, new mise tasks) is required by repo policy for workflow changes; it belongs with the end-to-end documentation.

## Live-run notes (for the merge-to-pivot window)

- aws: definitions are on main; an operator runs `mise run pivot` on the machine hosting the kind cluster for the aws environment. EKS control-plane and 2x t4g.medium billing for `eu-north-1-management` starts when its Cluster object is created.
- local-host: `mise -E local-host run pivot` (the script republishes the OCI artifact first).

Single logical commit; per the #84 review convention this PR covers both environments because the tooling does not separate cleanly (shared bootstrap-common.sh, shared skeleton, shared mise wiring).

### DIFF
--- a/bootstrap-common.sh
+++ b/bootstrap-common.sh
@@ -1,0 +1,249 @@
+#!/usr/bin/env bash
+# bootstrap-common.sh – Shared preflight and Flux-seeding helpers used by
+# bootstrap.sh and pivot.sh. Sourced, never executed directly.
+#
+# Environment variables set by these helpers are plain globals in the
+# sourcing shell (bash 3.2 compatible: no namerefs, no associative arrays).
+
+# ── Preflight: environment (aws) ──────────────────────────────────────────────
+# Validates the GitHub + age inputs the Flux seed needs on the aws environment.
+# Sets: GITHUB_USER, GITHUB_REPO, AGE_KEY_FILE, AGE_PUBKEY.
+require_flux_env() {
+  : "${GITHUB_TOKEN:?GITHUB_TOKEN must be set (a PAT with read access to the repo)}"
+  : "${GIT_REPO_URL:?GIT_REPO_URL must be set}"
+  # GITHUB_USER is used in the Flux GitHub secret for repo clone authentication
+  GITHUB_USER="${GITHUB_USER:-git}"
+
+  case "$GIT_REPO_URL" in
+    https://github.com/*/*) ;;
+    *)
+      echo "ERROR: GIT_REPO_URL must be an HTTPS GitHub repository URL" >&2
+      exit 1
+      ;;
+  esac
+  GITHUB_REPO="${GIT_REPO_URL#https://github.com/}"
+  GITHUB_REPO="${GITHUB_REPO%/}"
+  GITHUB_REPO="${GITHUB_REPO%.git}"
+  GITHUB_AUTH="Authorization: Bearer ${GITHUB_TOKEN}"
+  github_branch_path="${GIT_BRANCH//\//%2F}"
+  github_branch_status="$(curl -sS -o /dev/null -w '%{http_code}' \
+    -H 'Accept: application/vnd.github+json' \
+    -H "${GITHUB_AUTH}" \
+    "https://api.github.com/repos/${GITHUB_REPO}/branches/${github_branch_path}" || true)"
+  if [ "$github_branch_status" != 200 ]; then
+    echo "ERROR: GitHub repository or branch '${GIT_BRANCH}' is unavailable at '${GIT_REPO_URL}' (HTTP ${github_branch_status})" >&2
+    exit 1
+  fi
+
+  AGE_KEY_FILE="${AGE_KEY_FILE:-age.agekey}"
+  if [ ! -f "$AGE_KEY_FILE" ]; then
+    echo "ERROR: age key file not found at '$AGE_KEY_FILE'." >&2
+    echo "       Generate one with:  mise run sops-keygen" >&2
+    echo "       and add its PUBLIC key to .sops.yaml. See docs/secrets.md." >&2
+    exit 1
+  fi
+  # Validate age key file format first (before attempting to extract the public key).
+  # This avoids silent grep failure if the file is malformed.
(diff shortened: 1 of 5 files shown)


shrinedogg (commented):
## Code Review Summary

**Verdict: Comment - verified; safety ordering and CLI contracts hold**

Verified locally (head e60b28b):

- `mise run validate` exits 0: deps-check green with the two new pivot.sh slots, bash -n on all four scripts, all overlays. macOS bash 3.2 `/bin/bash -n` passes all three scripts.
- CLI contracts checked against the pinned binaries, not the docs: clusterctl 1.14.0 has `move --to-kubeconfig` and `-n/--namespace`; helm 4.2.2 has `--kubeconfig`, `--registry-config`, and `--wait=legacy` as a valid wait strategy.
- The mise task banner goes to stderr, so `mise -E aws run aws-credentials` captures only the credential string on stdout; the secret is not banner-polluted.
- The bootstrap-common.sh extraction is verbatim against the removed inline code, and `cleanup_bootstrap` wraps `seed_cleanup` and is called explicitly before `exec` (EXIT traps do not fire across exec; handled).
- The provider manifest kinds match the kubectl wait targets, the SOPS exclusion is real (explicit plain-file lists, never -k on directories containing .sops.yaml), the aws-identity fallback file exists, and the oci-push task exists for the local-host reseed.
- All four CI checks green; konflate shows no rendered changes, as expected for a scripts/catalog-only PR.

Note: I did not execute pivot.sh itself, not even the preflight (the environment blocked running a script whose tail end deletes a kind cluster). The context guard was verified by inspection: an empty or wrong current-context exits 1 before any wait, install, move, or delete. The author's live-kind smoke covers the same guard; I would still run `mise run pivot` for the first live pivot with PIVOT_SKIP_DELETE=1.

Note: the unpause loop after `clusterctl move` patches each moved Cluster with `|| true`, so a failed unpause is silently ignored. The subsequent Flux seed would surface a broken target anyway, so this is acceptable, but a log line per skipped failure would make a partially-paused target visible at a glance.

Note: AGENTS.md changes are deferred to the final docs PR per the PR description. Repo policy requires AGENTS.md updates in the same PR as workflow changes; since the deferral is declared up front, flagging it here so the follow-up is not lost: pivot.sh/bootstrap-common.sh layout, the pivot and mgmt-kubeconfig tasks, and the changed bootstrap default all belong in that update.

Looks good: the suspend-before-move ordering closes the mid-move unpause race, the move failure guidance is the right advice (never delete moved infra objects on the target), the kind-deletion guard requires clusters on the target plus registry health for local-host, and bootstrap's exec default makes fresh installs self-managing without changing the seeded Flux content.

Issue #80, opened 2026-08-27
Air-gap supply-chain hardening: digest pinning, signed SBOMs, offline verification, transactional updates


shrinedogg:
## Problem

The air-gap bundle (Zarf package, `airgap/images.txt`, rendered manifests) is the delivery mechanism for the local-host profile offline, but its supply-chain posture is unverified once deployed. Image references are tag-pinned rather than digest-pinned, the bundle carries no signed SBOM or signature evidence, verification today assumes network reachability, and the exact dependency graph behind each build is not recorded with the artifact. Recovery paths are untested: an air-gapped environment cannot assume a remote registry or a support channel will be available during recovery, and there is no transactional update or rollback strategy for partial artifact delivery.

## Goal

Make the air-gap bundle independently verifiable and safely recoverable offline: every image pinned by digest, SBOMs and signatures shipped inside the package and verifiable without network access, the exact dependency graph recorded per build, and updates applied transactionally with tested rollback.

## Proposed scope

- Pin every image in the air-gap package by immutable digest (tags retained only where needed for human readability).
- Generate an SBOM per build and ship it inside the Zarf package alongside signatures; verify both during deployment with no external network access.
- Record the exact dependency graph used for each build (image refs with digests, chart versions, tool versions) as a build artifact retained with the package.
- Design a transactional update flow: stage the new bundle, verify it completely, then atomically cut over; a failed or partial update must leave the previous known-good state intact.
- Document and test a rollback procedure that works with no registry and no support channel reachable.
- Extend the air-gapped CI workflow's failure-mode matrix:
 - stale system clocks (certificate and token validity),
 - unavailable external DNS,
 - certificate rotation mid-lifecycle,
 - storage pressure during staging and cutover,
 - partial artifact updates (interrupted transfer, truncated package).

## Acceptance criteria

- The Zarf package contains digest-pinned image references, an SBOM, and signatures verifiable fully offline.
- CI proves offline verification: the egress-blocked deployment job validates signatures and SBOMs before Flux reconciliation, not just absence of public traffic.
- Each build produces a recorded dependency graph artifact retained alongside the package.
- An interrupted or corrupted update demonstrably rolls back to the prior working state without external resources.
- The five failure modes above are covered by automated tests in the air-gapped workflow.

## Non-goals

- Replacing Zarf with another packaging tool.
- Full binary provenance (SLSA level 3+) for upstream third-party images we consume but do not build.
- Online revocation checking, which contradicts the offline requirement.

[...]

PR #91, opened 2026-08-28
feat: pin airgap images by digest

lmilbaum:
## Summary

- enable Renovate digest pinning for the air-gap image inventory and Zarf package definition
- retain readable image tags while teaching the Zarf matcher to preserve SHA-256 digests
- add a registry-backed CI integration test proving both files receive pinDigest updates
- pin and catalog the Renovate and GitHub Action versions used by the new validation job

This is the first incremental step toward #80. Renovate will create the follow-up image digest changes; this PR configures and verifies that behavior.

## Testing

- python3.13 deps/scripts/check.py --forbid
- Python syntax compilation for airgap/tests/test-renovate-digest-pinning.py
- Renovate 44.50.1 config validation
- Renovate registry-backed lookup: 45 proposed pins for airgap/images.txt and 14 for airgap/zarf.yaml
- git diff --check

Refs: #80


shrinedogg (approved):
## Code Review Summary

**Verdict: Approve - verified end to end, including the exact pin counts**

Verified locally (head ce66539):

- Integration test run here with Renovate 44.50.1 installed into a prefix: pin counts are `airgap/images.txt: 45`, `airgap/zarf.yaml: 14`, matching the PR description exactly. The only local failure was the Renovate engine requirement (`node: ^24.11.0`; this machine runs 22), which makes the workflow's `setup-node` with node 24 load-bearing rather than incidental; the CI job is green at 52s.
- The counts reconcile: zarf.yaml's 14 is the 15 image entries minus `localhost:5001/knr-ops-airgap:latest` (no public registry to pin against); images.txt's 45 is every line entry, and `pinDigest` works on floating tags like `latest` because it needs no version comparison.
- Matcher probe with renovate's own json5 parser and `g`-flag semantics: the new zarf matchString extracts all 15 entries today (none digest-pinned yet), and a simulated `repo:tag@sha256:...` ref parses into depName/currentValue/currentDigest correctly. The images.txt matcher still covers all 45 lines (39 unique dep:value pairs).
- `deps/scripts/check.py --forbid` passes with the new `renovate` slot; the docs inventory table regenerates as an exact no-op; `actions/checkout@v7.0.1` and `actions/setup-node@v7.0.0` both resolve to real tags.
- Scope is tight: pinDigests applies only to docker deps in the two airgap files, the digest group is optional so unpinned refs keep working, and automerge stays false so every pin lands as a reviewed PR.
- Downstream compatibility: the inventory check's repo-name parsing already strips `@sha256:` suffixes, and images.txt/zarf.yaml sit outside the forbid gate's scanned paths, so the first digest-pinning PR will not trip existing checks.

Looks good: the registry-backed test is the right kind of proof for this config (it would have caught the pattern-level bugs this config has seen before), and pinning the test's own toolchain (renovate 44.50.1, checkout, setup-node) in the catalog keeps the validation deterministic.

Issue #92, opened 2026-08-28
Rewrite the imperative bootstrap as a Rust CLI (bootstrap-rs)


shrinedogg:
## Problem

The one-time imperative bootstrap is a 415-line bash script (`bootstrap.sh`) that mixes argument handling, environment validation, container engine detection, HTTP checks via `curl`, heredoc-generated kind config, background process management (`flux logs --follow` plus trap cleanup), and polling loops. It is hard to test, error handling depends on `set -euo pipefail` semantics, and every behavior change carries shell quoting and portability risk. The script is also the entrypoint for both profiles (`local-host` and `aws`), so regressions affect all environments.

## Goal

Replace `bootstrap.sh` with a single statically-linked Rust binary, `knr-bootstrap`, that preserves the exact bootstrap behavior, step order, and output while gaining typed argument parsing, testable units, and structured error handling. An initial port exists on the `feat/bootstrap-rs` branch under `bootstrap-rs/`.

## Proposed scope

1. Port `bootstrap.sh` to a Rust CLI under `bootstrap-rs/`:
   - `clap` derive CLI: profile as a positional value enum (`local-host`, `aws`), every script env knob exposed as a flag with the same env var name (`KNR_OPS_PROFILE`, `REGISTRY_PORT`, `REGISTRY_READY_RETRIES`, `LOCAL_RECONCILE_TIMEOUT`, `CONTAINER_ENGINE`, `GIT_REPO_URL`, `GITHUB_TOKEN`, `GITHUB_USER`, `AGE_KEY_FILE`, `AGE_PUBLIC_KEY`, `OCI_REPOSITORY`, `OCI_TAG`), so existing `.env` workflows keep working unchanged.
   - `reqwest` (`default-features = false`, rustls) replaces the two `curl` calls: the GitHub branch availability check and the local registry `/v2/` readiness poll.
   - `tokio` process execution for all tool invocations (`kind`, `kubectl`, `helm`, `mise`, `clusterctl`, `flux`, docker/podman), preserving podman-via-docker-shim detection, the kind config passed on stdin, the `sops-age` dry-run-and-apply pipe, the background `flux logs` follower with kill-on-drop cleanup, and the 60x2s discovery polls.
2. Add tests: unit tests for age key file validation, GIT_REPO_URL parsing, and workload port extraction; a smoke test that exercises preflight failure paths without a cluster.
3. Integrate with knr-ops conventions where applicable:
[...]

PR #93, opened 2026-08-28
feat: Rust port of the imperative bootstrap (bootstrap-rs)

shrinedogg:
## What

Adds `bootstrap-rs/`, a Rust CLI (`knr-bootstrap`) that reimplements the one-time imperative bootstrap as a single binary. Step order and terminal output match `bootstrap.sh`, which stays in place until the binary has proven itself on both environments (decommission path in #92).

- `clap` derive CLI limited to the script's own surface: a positional profile value enum (`local-host`, `aws`) and `--recreate`. Every environment knob the script reads (`KNR_OPS_PROFILE`, `REGISTRY_PORT`, `REGISTRY_READY_RETRIES`, `LOCAL_RECONCILE_TIMEOUT`, `CONTAINER_ENGINE`, `GIT_REPO_URL`, `GITHUB_TOKEN`, `GITHUB_USER`, `AGE_KEY_FILE`, `AGE_PUBLIC_KEY`, `OCI_REPOSITORY`, `OCI_TAG`) is read from the environment with the script's `${VAR:-default}` semantics (unset and empty both fall back), so existing `.env` workflows keep working unchanged. The expanded flag interface is deferred to a follow-up PR for separate review.
- `reqwest` (default-features off, rustls) replaces the two direct `curl` invocations in the script: the GitHub branch availability check and the local registry `/v2/` readiness poll, with explicit connect (10s) and request (30s) timeouts. `curl` is still required for the `local-host` profile because the existing `mise oci-push` task shells out to it for its own registry check; it stays in the local-host preflight list until #92 rewires that task. The flow therefore still depends on `curl` being installed for local-host; only the script's own two `curl` calls are replaced.
- The registry readiness poll matches the script's `curl --fail --retry "$REGISTRY_READY_RETRIES" --retry-connrefused --retry-delay 1` behavior: one initial attempt plus N retries 1s apart (default 120), any HTTP response below 400 succeeds, connection errors and 5xx are retried, and a definitive 4xx (except 408) fails immediately without consuming the retry budget, exactly as curl does.
- `tokio` processes for every tool invocation (`kind`, `kubectl`, `helm`, `mise`, `clusterctl`, `flux`, docker/podman), preserving podman-via-docker-shim detection, the kind config piped via stdin, and the 60x2s discovery polls. The background `flux logs --follow` child is killed by a drop guard on both the success and error paths, mirroring the script's EXIT trap; the temp kubeconfig is a NamedTempFile cleaned the same way. Complete signal-driven cancellation (terminating and reaping every active child on SIGINT/SIGTERM) is intentionally out of scope here and tracked for a focused follow-up, so no partial lifecycle guarantees ship in this PR.

## Restart safety (intentional divergence from the script)

The script deletes and recreates the `mgmt` cluster on every run. The binary instead:

- Reuses an existing `mgmt` cluster by default after validating it (context reachable — recovered via `kind export kubeconfig` when missing — and all nodes Ready within the script's hardcoded 120s wait); replacement requires the explicit `--recreate` flag, and an unhealthy existing cluster fails with a message pointing at it.
- Makes every step rerun-safe so a partial failure resumes by rerunning: the registry configmap and both flux-system secrets are `kubectl apply`ed as JSON manifests on stdin, and both helm installs use `helm upgrade --install` with pinned chart versions and an explicit `--timeout 10m`. The sops-age secret keeps delete-then-apply to avoid merging stale key entries. An existing `knr-registry` container is validated for host-port binding and kind-network attachment and reconciled (recreated) when misconfigured.
- Keeps secret material off argv everywhere (secrets travel in stdin manifests), so command failure errors can echo the full command line safely. The GitHub PAT is read from `GITHUB_TOKEN` only; there is no token flag.
- Preflight reports all missing tools at once and covers every binary the run invokes (`flux` and `clusterctl` required for local-host; `curl` required for local-host for as long as the mise oci-push task uses it).

## Tests and CI

- 17 unit tests: CLI contract (defaults matching the script, `--recreate`, profile rejection, rejected deferred flags), env-knob resolution with `${VAR:-default}` and empty-string semantics, `GIT_REPO_URL` parsing, age key validation and pubkey extraction, workload port extraction, per-profile kind config rendering, required-tools coverage, secret JSON round-trip.
- `.github/workflows/bootstrap-rs.yml`: fmt check, clippy `--locked -D warnings` across all targets, locked build and test; path-filtered to `bootstrap-rs/**`, cargo caching.
- `rust-toolchain.toml` pins 1.96.1 with clippy and rustfmt for identical local and CI toolchains.

Not in this PR, tracked in #92: mise task wiring, docs and AGENTS.md updates, airgap bundle integration, the expanded CLI flag interface, and complete signal-driven child cancellation.

## Validation

- CI green on this PR, including the new Build, lint, and test job.
- Preflight failure paths exercised against the built binary: invalid profile, missing `GITHUB_TOKEN`, non-HTTPS `GIT_REPO_URL`, GitHub 401 on the branch check (live reqwest call), malformed age key file. Same messages and nonzero exits as the script.
- Profile precedence probed on the built binary: non-empty `KNR_OPS_PROFILE` wins over the positional (aws env + local-host positional runs aws, matching the script's `${KNR_OPS_PROFILE:-${1:-aws}}`), empty env values fall through to the positional, and the deferred flags are rejected with clap's unexpected-argument error.
- Full bootstraps deliberately not run here: they would touch the running `mgmt` kind cluster. Both environments are part of the #92 acceptance criteria before `bootstrap.sh` retires.

Refs #92.


lmilbaum:
Should the new bootstrap be restart-safe and idempotent? Currently a rerun deletes and recreates the management cluster, which is repeatable but destructive. I suggest reusing and validating existing state by default, supporting resume after partial failure, and requiring an explicit `--recreate` flag for replacement. I would also expect tests and Rust CI in this PR, secret-safe command errors, complete tool preflight checks, explicit timeouts, and a pinned Rust toolchain before integration.

bootstrap-rs/src/main.rs
```diff
@@ -0,0 +1,1239 @@
...
+
+/// One-time imperative bootstrap for the knr-ops management cluster.
+///
+/// Reruns are safe: an existing healthy 'mgmt' cluster is reused and all
+/// steps are idempotent, so a partial failure can be resumed by rerunning.
+#[derive(Parser, Debug)]
+#[command(name = "knr-bootstrap", version, about)]
+struct Cli {
+    /// Deployment profile
+    #[arg(value_enum, env = "KNR_OPS_PROFILE", default_value = "aws")]
```
shrinedogg: Parity: profile precedence is inverted against the script. bootstrap.sh resolves a non-empty KNR_OPS_PROFILE first and the positional second; clap resolves the positional first. Probed on the built binary: `KNR_OPS_PROFILE=aws knr-bootstrap local-host` runs the local-host profile where the script runs aws. Empty-string env values also diverge: `KNR_OPS_PROFILE=` exits with a parse error where the script falls back to the positional or aws. Suggest resolving the profile manually (non-empty env over positional, aws last) and treating empty env values as unset for every env-backed knob that the script reads with `${VAR:-default}`.

bootstrap-rs/src/main.rs
```diff
@@ -0,0 +1,1239 @@
...
+        "metadata": {
+            "name": "local-registry-config",
+            "namespace": "kube-system",
+        },
+        "data": { "registry-url": format!("{REGISTRY_NAME}:5000") },
+    }))
+    .await?;
+
+    println!(">>> Publishing initial OCI artifact from the local Git checkout...");
+    run("mise", &["-E", "local-host", "run", "oci-push"]).await?;
```
shrinedogg: Parity: the resolved CLI values never reach oci-push. The mise child inherits the process environment, so `--registry-port 5500` creates and polls 5500 while the task curls 5001 (mise.local-host.toml reads REGISTRY_PORT itself). Same for --oci-repository and --oci-tag. Suggest setting REGISTRY_PORT, OCI_REPOSITORY, and OCI_TAG explicitly on the mise child from the parsed Cli values. Note the overrides also cannot complete local-host end to end today: the workload FluxInstance hardcodes knr-ops:latest (mgmt/local-host/addons/flux-apps/flux-instance.yaml), worth a flag-help caveat until #92 templates it.

bootstrap-rs/src/main.rs
```diff
@@ -0,0 +1,1239 @@
...
+        "kind: Cluster\napiVersion: kind.x-k8s.io/v1alpha4\n{registry_patch}nodes:\n  - role: control-plane\n    extraMounts:\n      - hostPath: {engine_sock}\n        containerPath: /var/run/docker.sock\n"
+    )
+}
+
+/// Tools required on PATH for the given profile.
+fn required_tools(profile: Profile) -> Vec<&'static str> {
+    // The binary owns the HTTP checks the script used curl for, so curl is
+    // no longer required. flux and clusterctl are exercised by the
+    // local-host reconciliation watch (Step 5).
+    let mut tools = vec!["kind", "helm", "kubectl"];
```
shrinedogg, line 246: Parity: curl is dropped from the local-host list, but oci-push still invokes curl itself (the registry availability check in mise.local-host.toml). On a host without curl, preflight passes and the run fails only after the cluster and registry mutations. Suggest keeping curl in the local-host list for as long as the task shells out to it.

bootstrap-rs/src/main.rs
```diff
@@ -0,0 +1,1239 @@
...
+    if !status.success() {
+        bail!("'{cmd} {}' failed with {status}", args.join(" "));
+    }
+    Ok(())
+}
+
+/// Apply a Kubernetes manifest (JSON) via `kubectl apply -f -`. Idempotent,
+/// and keeps secret values off argv.
+async fn kubectl_apply(manifest: &serde_json::Value) -> Result<()> {
+    run_with_stdin("kubectl", &["apply", "-f", "-"], &manifest.to_string()).await
```
shrinedogg, line 357: Suggestion: client-side apply persists the submitted manifest in the kubectl.kubernetes.io/last-applied-configuration annotation, and with stringData the values remain plaintext there (kubernetes issue 89938 documents the stringData variant). The argv goal is met; this is the residual exposure. Applying via pre-encoded data with create-or-replace semantics (the sops-age path already deletes first) avoids both the annotation and the stringData apply quirks.

bootstrap-rs/src/main.rs
```diff
@@ -0,0 +1,1239 @@
...
+    /// Container engine to use (docker or podman); auto-detected when omitted
+    #[arg(long, env = "CONTAINER_ENGINE")]
+    container_engine: Option<String>,
+
+    /// HTTPS GitHub repository URL (required for the aws profile)
+    #[arg(long, env = "GIT_REPO_URL")]
+    git_repo_url: Option<String>,
+
+    /// GitHub PAT with read access to the repo (required for the aws profile)
+    #[arg(long, env = "GITHUB_TOKEN", hide_env_values = true)]
```
shrinedogg: Suggestion: hide_env_values only affects help rendering; the flag itself puts the PAT in shell history and process listings, which cuts against the no-secrets-on-argv invariant the PR sets for itself. Consider dropping the flag and reading GITHUB_TOKEN from the environment only.

bootstrap-rs/src/main.rs
```diff
@@ -0,0 +1,1239 @@
...
+/// cluster exists) the cluster is (re)built from the rendered config.
+async fn ensure_kind_cluster(cli: &Cli, engine_sock: &str) -> Result<()> {
+    let clusters = capture_lossy("kind", &["get", "clusters"]).await;
+    let exists = clusters.lines().any(|l| l.trim() == "mgmt");
+    let node_ready_timeout = format!("--timeout={}", cli.node_ready_timeout);
+
+    if exists && !cli.recreate {
+        println!(">>> Reusing existing kind cluster 'mgmt' (pass --recreate to replace it)...");
+        println!(">>> Validating existing cluster health...");
+        if run_quiet("kubectl", &["config", "use-context", "kind-mgmt"]).await
```
shrinedogg, line 563: Restart-safety gap: kind get clusters can report mgmt while the kubeconfig lacks the kind-mgmt context (interrupted first create, pruned kubeconfig). The reuse path fails here and the only recovery is destructive --recreate. Suggest attempting `kind export kubeconfig --name mgmt` before giving up, then retrying the health check.

shrinedogg (commented):
Reviewing against current main (post-#79), which changed the bootstrap contract after this branch's original base; thanks for merging main in.

The quality bar from the earlier comment is met: reuse by default with --recreate, rerun-safe applies, secrets off argv with safe error text, complete preflight reporting, explicit timeouts, pinned toolchain, 13 unit tests, and a green fmt/clippy/locked build and test job.

Scope: the pivot (BOOTSTRAP_PIVOT, bootstrap-common.sh, pivot.sh) is the one structural gap, and #92 already gates bootstrap.sh retirement on full-parity runs against current main, so these comments are the checklist for that pass plus three parity bugs to fix here (profile precedence, oci-push env forwarding, curl preflight).

Verified locally: cargo fmt, clippy --locked, build --locked, and test --locked all clean with 13/13 tests; profile precedence and empty-env behavior probed against the built binary; helm template --set-json with the exact Rust literal renders identically to the script form; deps-check reproduces the bootstrap-rs/target false positives after a build.

Warnings: parity bugs (profile precedence inverted vs KNR_OPS_PROFILE, mise oci-push sees neither flags nor env so --registry-port breaks the task, curl still transitively required after mutations); restart-safety gaps on the reuse paths (missing kind-mgmt context, stale registry port or network); deprecated containerd registry.mirrors on containerd 2.x; unpinned chart installs vs the 0.58.0 Git pin; clippy running unlocked before the locked checks; catalog and deps-check integration; signal-time kubeconfig cleanup.

Suggestions: last-applied-configuration stores stringData plaintext (kubernetes.io #89938); --github-token puts the PAT on argv.

bootstrap-rs/src/main.rs
```diff
@@ -0,0 +1,1239 @@
...
+                "kind",
+                "-p",
+                &publish,
+                "registry:2",
+            ],
+        )
+        .await?;
+        println!("    Registry created and running: localhost:{port}");
+    } else {
+        let running = capture_lossy(
```
shrinedogg: Restart-safety gap: an existing knr-registry is reused on name and Running alone; the host-port binding and kind-network attachment are never checked, and --recreate rebuilds the cluster without reconciling the registry. A container bound to a different port fails readiness late with a misleading message; one detached from the kind network passes host readiness and fails later in-cluster at knr-registry:5000. Suggest inspecting and reconciling port and network on the existing path.

bootstrap-rs/src/main.rs
```diff
@@ -0,0 +1,1239 @@
...
+        .map(str::trim)
+        .filter(|p| !p.is_empty() && p.chars().all(|c| c.is_ascii_digit()))
+        .map(str::to_string)
+}
+
+/// Render the kind cluster config, mirroring the script's heredoc.
+fn render_kind_config(profile: Profile, registry_port: u16, engine_sock: &str) -> String {
+    let registry_patch = if profile == Profile::LocalHost {
+        format!(
+            "containerdConfigPatches:\n  - |-\n    [plugins.\"io.containerd.grpc.v1.cri\".registry.mirrors.\"localhost:{registry_port}\"]\n      endpoint = [\"http://{REGISTRY_NAME}:5000\"]\n"
```
shrinedogg, line 230: The mirrors stanza is deprecated on containerd 2.x (removed per containerd PR 9766; config_path with certs.d hosts.toml is the supported mechanism, see the kind local-registry guide). The pinned kindest/node v1.36.1 runs containerd 2.x. Verified on a live mgmt node: the generated stanza lands via the compat shim, so it works today and silently breaks when the shim is removed. A faithful port of the script inherits that deprecation debt; worth fixing here rather than in the script.

bootstrap-rs/src/main.rs
```diff
@@ -0,0 +1,1239 @@
...
+            "flux-system",
+            "--for=condition=Ready",
+            "--timeout=10m",
+        ],
+    )
+    .await?;
+
+    // Verify the Flux controllers are running before declaring success.
+    println!(">>> Waiting for Flux controllers to be ready...");
+    let _ = run(
```
shrinedogg: Silent success: the `let _ =` mirrors the script's `|| true`, but for the aws profile this is the last check before the completion message, so ImagePullBackOff or missing pods still exit 0. Suggest distinguishing a wait timeout (print pod status, tolerate) from a hard failure, or suppressing the completion message when the wait failed.

bootstrap-rs/src/main.rs
```diff
@@ -0,0 +1,1239 @@
...
+    println!(">>> Installing Flux Operator...");
+    let cfg = registry_config.to_string_lossy();
+    // upgrade --install (script used install): rerun-safe after a partial failure.
+    run(
+        "helm",
+        &[
+            "upgrade",
+            "--install",
+            "flux-operator",
+            "oci://ghcr.io/controlplaneio-fluxcd/charts/flux-operator",
```
shrinedogg, line 795: Repo convention: both helm installs use bare OCI refs (here and the flux-instance chart), so reruns can resolve different privileged cluster software with no repository change. Git pins flux-operator 0.58.0 (mgmt/*/addons/flux-apps/flux-operator.yaml) and deps/versions.toml tracks it. Suggest --version with catalog-validated constants for both charts.

bootstrap-rs/src/main.rs
```diff
@@ -0,0 +1,1239 @@
...
+            cli.local_reconcile_timeout
+        );
+        let _ = run("flux", &["get", "kustomizations"]).await;
+        bail!("local-host reconciliation timed out");
+    }
+
+    println!();
+    println!(">>> Workload cluster Flux reconciliation errors");
+    let workload_kubeconfig =
+        tempfile::NamedTempFile::new().context("failed to create temp kubeconfig")?;
```
shrinedogg, line 1037: Signal hygiene: destructors do not run on default SIGINT/SIGTERM termination, so Ctrl-C during the final workload wait leaves this kubeconfig (live cluster credentials) on disk in TMPDIR; the script's EXIT trap removes it (verified in a sandboxed harness). Suggest tokio signal handling that cancels into cleanup, or unlink housekeeping for the tempfile prefix on startup.

.github/workflows/bootstrap-rs.yml
```diff
@@ -0,0 +1,58 @@
...
+            ~/.cargo/registry
+            ~/.cargo/git
+            bootstrap-rs/target
+          key: bootstrap-rs-${{ runner.os }}-${{ hashFiles('bootstrap-rs/Cargo.lock', 'bootstrap-rs/rust-toolchain.toml') }}
+
+      - name: Format check
+        run: cargo fmt --check
+
+      - name: Clippy
+        run: cargo clippy --all-targets -- -D warnings
```
shrinedogg: CI gap: clippy is the first dependency-resolving cargo command and runs unlocked; if Cargo.toml and Cargo.lock drift, clippy updates the lockfile and the later locked build and test validate the modified tree. One-word fix: `cargo clippy --locked --all-targets -- -D warnings`.

bootstrap-rs/rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.96.1"
```
shrinedogg, line 2: Repo convention: the new pinned literals (1.96.1 here, actions/cache@v4 and the workflow's checkout@v5, registry:2) are absent from deps/versions.toml, which main now extends with slots for imperative pins (pivot.sh chart versions). Related and reproducible: deps-check --forbid scans every file, so a normal cargo build leaves bootstrap-rs/target artifacts that fail mise run deps-check (4 false positives on the helm literal). Suggest catalog keys plus slots for the toolchain and workflow actions, and excluding bootstrap-rs/target in the scanner, alongside the #92 mise wiring.

shrinedogg (commented):
## Code Review Summary

**Verdict: Comment - verified; best practices hold throughout**

Verified locally (head 662e640):

- Toolchain gates re-run here: `cargo fmt --check` clean, `cargo clippy --locked --all-targets -- -D warnings` zero warnings, `cargo test --locked` 15/15 pass (the PR says 13; the hardening commit added two more tests, so the description is one commit stale).
- All claimed preflight failure paths exercised against the built binary: invalid profile (clap error with possible values), missing GITHUB_TOKEN (the script's exact message), non-HTTPS GIT_REPO_URL (exact message, exit 1), a live reqwest GitHub check with a fake token (HTTP 401, exact message), and a malformed age key (all three missing fields reported, after the valid-token branch check passed end to end).
- Best-practice scan: no unwrap/expect/panic in non-test code beyond the two signal-handler expects (unrecoverable on failure, idiomatic). Secrets travel only in stdin JSON manifests; error paths echo argv safely; the token binding uses hide_env_values. ChildGuard kills the flux logs follower on drop, SIGINT/SIGTERM cancel the future so Drop cleanup runs, HTTP calls carry explicit connect (10s) and request (30s) timeouts, and temp files are NamedTempFile auto-cleaned.
- Project hygiene: Cargo.lock committed (correct for a binary), target/ gitignored, rust-toolchain.toml pins 1.96.1 with clippy and rustfmt, rust-version declared in Cargo.toml, and the workflow is path-filtered with contents:read, concurrency cancellation, locked builds, and cache keyed on lockfile plus toolchain.
- lmilbaum's review points are all covered: restart-safe reuse with health validation and --recreate, idempotent steps (apply, upgrade --install), tests, Rust CI, secret-safe errors, complete preflight reporting all missing tools at once, explicit timeouts, pinned toolchain. The reuse path's extra care shows: it recovers a missing context via kind export kubeconfig instead of demanding destruction, and it validates the existing registry's port binding and kind-network attachment before trusting it.

Suggestion: the new workflow adds `actions/cache@v4` with no catalog entry. #91 established the convention of cataloging new workflow action versions (gh_checkout_renovate_test, gh_setup_node); a gh_cache key would keep that consistent. No gate enforces it today, so this is convention-level, and the checkout@v5 reuse of the existing key is already consistent.

Note: the binary reads process env only; .env workflows arrive via the mise task wiring tracked in #92, which the PR correctly scopes out. The local-host `curl` requirement is documented as belonging to the mise oci-push task, not the binary itself.

Looks good: the intentional divergence from the script (reuse instead of delete-recreate) is the right default for a tool meant to resume partial failures, the step order and messages match bootstrap.sh, and bootstrap.sh stays in place until the #92 acceptance runs prove both environments.

shrinedogg:
Suggestion from my review (https://github.com/polarsquad/knr-ops/pull/93#pullrequestreview-5054121617) addressed in 7db68e3: actions/cache@v4 is now cataloged as `gh_cache` in deps/versions.toml per the #91 convention (recorded, no slot), and the docs inventory was regenerated from the catalog (which also picks up the (validated in pivot.sh) annotations this branch's table predated). All checks green on the new head.

lmilbaum:
### CLI requirements and choice of Rust

Based on our discussion, I think it would be useful to make the high-level requirements for the CLI explicit. This also gives us something to evaluate the implementation against, beyond achieving 1:1 parity with the current Bash script.

At a high level, I see the requirements as:

* **Functional parity:** Preserve the capabilities of the existing bootstrap script.
* **Maintainability:** Keep the implementation easy to understand, modify, review, and test as the project grows. This is particularly important since maintainability is one of the reasons for moving away from increasingly complex Bash scripts.
* **Reliability:** Provide predictable behavior, clear error handling, and safe recovery/retry where applicable.
* **Extensibility:** Make it straightforward to add new providers and deployment targets without significantly increasing the complexity of the core CLI.
* **Testability:** Allow the orchestration logic to be tested independently where possible.
* **Portability and distribution:** Make the CLI easy to run across the environments we want to support, with minimal runtime dependencies.
* **Security:** Handle credentials and sensitive information safely.

With those requirements in mind, I think Rust has some interesting advantages for this use case:

* Single-binary distribution with no language runtime dependency
* Strong type and memory safety
* Strong error-handling model
* Good tooling and libraries for building CLIs
* Good cross-platform support
* Well suited for building a CLI that may grow beyond what is comfortable to maintain as shell scripts

There are also tradeoffs worth acknowledging:

* Steeper learning curve than languages such as Go or Python
* Potentially higher implementation complexity
* Longer compile times
* Smaller pool of contributors already familiar with Rust compared with some alternatives

I don't think this means we need a formal requirements process at this stage. But having a small set of agreed requirements gives us a useful way to evaluate both the CLI design and the technology choice as knr-ops grows.

The important question for me isn't really "Why Rust?" but rather "Does Rust, and the way we're using it here, help us satisfy the requirements we have for the CLI?"

lmilbaum:
Since this PR aims to preserve bootstrap.sh behavior, could we defer the additional CLI flags to a follow-up PR? For now, the Rust CLI could support the existing positional profile and environment variables only. That would keep this change focused on behavioral parity and let us review the expanded CLI interface separately.

lmilbaum:
Replacing the two direct `curl` calls with `reqwest` makes sense for a Rust binary. However, `curl` is still required for the `local-host` profile because the existing `mise oci-push` task invokes it. Could you clarify this in the PR description so it doesn’t suggest that the bootstrap flow no longer depends on `curl`? We should also verify that the new timeout and retry behavior matches the existing script, especially for the local registry readiness check.

lmilbaum:
Using tokio to run the existing external tools seems reasonable. The added signal handling and child-process cleanup introduce behavior beyond the current script, though. If complete cancellation handling—including terminating and reaping every active child process—is not required for behavioral parity, could we defer it to a focused follow-up PR? That would keep this PR centered on reproducing the existing bootstrap flow and avoid introducing partially implemented lifecycle guarantees.

shrinedogg:
All four points from today's comments are addressed. CI is green on 1c46991 (fmt, clippy `-D warnings`, locked build, 17/17 tests).

**1. CLI requirements and choice of Rust**

Agreed on making the requirements explicit; they are a better review lens than line parity. How the current implementation maps:

- Functional parity: step order, terminal messages, the env contract, and exit codes match `bootstrap.sh`; the #92 acceptance runs gate retirement on full-parity runs against current main.
- Maintainability: one ~1500-line binary with typed config and unit tests replacing a 261-line script plus a 249-line sourced library; behavior changes become visible diffs against named functions instead of shell globals.
- Reliability: restart-safe cluster reuse with health validation, rerun-safe steps, preflight that reports all missing tools at once.
- Extensibility: the profile enum and the per-step functions are the seams for new providers; the pivot (BOOTSTRAP_PIVOT, bootstrap-common.sh, pivot.sh) is the known structural gap, tracked as the checklist for the #92 parity pass.
- Testability: 17 unit tests over the pure helpers and the config resolution (env semantics are injectable via `Config::from_env`); the remaining untested layer is thin process plumbing.
- Portability: single binary, no language runtime; the tools on PATH remain the runtime dependencies.
- Security: secrets travel in stdin manifests only and never on argv, the PAT is env-only (there is no token flag), and command errors can echo the full command line safely.

On your closing question: Rust serves these requirements well at this size, and the tradeoffs you list (contributor pool, compile times) are the real costs rather than the language's capabilities. If the CLI grows toward richer provider logic, the type system and error model start paying for themselves beyond what Bash could express; if it stays a thin imperative wrapper, Go would cover the same ground with a larger contributor base. I would like to capture this requirements list in `docs/` as part of #92 so it outlives the PR discussion and becomes the yardstick for the CLI's growth; happy to do that there.

**2. Defer the additional CLI flags**

Done in 613f073. The CLI now accepts only the positional profile and `--recreate` (kept from your first-round ask). All twelve script env knobs are read from the environment with the script's `${VAR:-default}` semantics, including empty-string behaving as unset, resolved in a unit-tested `Config::from_env`; `NODE_READY_TIMEOUT` folds back to the script's hardcoded `kubectl wait --timeout=120s`. clap's `env` feature was dropped along with its last user, and a test pins the reduced surface by asserting the deferred flags fail to parse. The expanded interface can be re-proposed as its own PR against the `Config` seam, where it is now a small, reviewable diff.

**3. curl clarification and readiness timeout/retry parity**

The description now states it explicitly: reqwest replaces the script's own two curl calls only, and `curl` remains required for `local-host` because the existing `mise oci-push` task shells out to it for its registry availability check. It stays in the local-host preflight list (the `required_tools` test covers it) until #92 rewires that task.

Your verification nudge caught a real imprecision: the loop was retrying all 5xx responses, but `curl --retry` retries a documented transient set only (408, 429, 500, 502, 503, 504). Fixed in 1c46991; the poll now matches the script's curl invocation exactly: any response below 400 succeeds (`--fail`'s threshold), connection errors retry (`--retry-connrefused`), the transient set retries once per second for `REGISTRY_READY_RETRIES` retries after the initial attempt, and any other answer fails immediately with the status code instead of burning the retry budget. One deliberate divergence remains and is now stated in the description: each attempt is bounded by the explicit 10s connect and 30s request timeouts, which the script's plain curl invocation does not apply per attempt.

**4. Defer signal handling and child-process cleanup**

Done in 613f073: `shutdown_signal` and the `tokio::select!` wrapper are removed, along with the tokio `signal` feature and its two signal-handler `expect`s (no `unwrap`/`expect`/`panic` remains outside tests). Cleanup coverage is now identical to what the script's EXIT trap covers: normal exit and error exit run the drop guards (temp kubeconfigs, the registry config temp file, the flux logs follower). The residual gap also matches the script's: destructors do not run on default SIGINT/SIGTERM termination, so an interrupt during the final workload wait leaves the temp kubeconfig in TMPDIR until the OS reclaims it. Full cancellation with terminating and reaping every active child belongs in the focused follow-up you suggested; it is listed in the description's out-of-scope note alongside the CLI flags.

PR #94, opened 2026-08-28
feat: verify signed airgap SBOMs offline

lmilbaum:
## Summary

- require every air-gap package build to sign the completed Zarf archive
- use GitHub OIDC keyless signing in upstream CI while retaining operator-managed key signing
- verify package signatures and checksums, extract embedded Syft SBOMs, and validate them before any cluster mutation
- retain extracted SBOMs with isolated and monitored deployment evidence
- document the signing and offline trust workflow

Refs #80

## Testing

- `python3.13 -m unittest discover -s airgap/scripts -p 'test_*.py' -v`\n- `python3.13 deps/scripts/check.py --forbid`\n- `python3.13 airgap/scripts/check-inventory.py`\n- `bash -n bootstrap.sh teardown.sh pivot.sh bootstrap-common.sh airgap/scripts/*.sh`\n- parsed `.github/workflows/air-gapped.yml` with `yq`\n- built every `mgmt/` and `workload/` Kustomize overlay with `kubectl kustomize`\n- created a fixture package with Zarf v0.83.0, extracted its embedded SBOM, and validated the real emitted schema\n\nThe full signed ARM64 package build and egress-isolated deployment run only on upstream `main`, so that end-to-end OIDC verification will run after merge via the scheduled/manual `air-gapped` workflow.


lmilbaum:
Does `zarf package inspect sbom --verify=always` already fail when the embedded SBOM is missing, corrupt, or not a valid Syft document? If so, could we rely on that command’s exit status instead of maintaining `verify-sbom.py` and its tests? If it does not provide the required guarantee, could we use an established Syft or JSON Schema validation command rather than implementing a partial schema check ourselves?

airgap/scripts/build-package.sh
```diff
@@ -96,9 +106,26 @@ helm pull oci://ghcr.io/controlplaneio-fluxcd/charts/flux-operator --version 0.5
 helm pull oci://ghcr.io/stefanprodan/charts/podinfo --version 6.14.0 -d airgap/archives/charts
 echo "    staged charts: $(ls airgap/archives/charts/)"
 
-echo "==> 4/4 zarf package create"
+echo "==> 4/5 zarf package create (SBOM generation enabled)"
 cd airgap
 mise x -- zarf package create . --confirm
 
+PACKAGE="$PWD/zarf-package-knr-ops-airgap-arm64-0.1.0.tar.zst"
```
shrinedogg, line 113: Suggestion: this hardcodes the version from zarf.yaml metadata. A version bump now needs the literal updated here (and in the offline-run.sh:24 default, pre-existing). Consider deriving the name from airgap/zarf.yaml or keeping a glob for the existence check. Minor.

airgap/scripts/offline-run.sh
```diff
@@ -72,7 +84,19 @@ else
   done
 fi
 
-step "1. stage: docker load + kind create + seed registry"
+step "1. verify package signature, checksums, and embedded SBOMs"
+SBOM_OUTPUT=$(mktemp -d "${TMPDIR:-/tmp}/airgap-sbom.XXXXXX")
```
shrinedogg, line 88: Suggestion: mktemp honors TMPDIR while the workflow uploads the fixed glob /tmp/airgap-sbom.*. GitHub-hosted runners do not set TMPDIR so this matches today; a self-hosted runner with TMPDIR set would silently skip the evidence upload (if-no-files-found: warn). A fixed path next to the other evidence files, or echoing the dir into the summary, would make retention deterministic.

airgap/scripts/offline-run.sh
```diff
@@ -72,7 +84,19 @@ else
...
+if "$ZARF" package verify "$PACKAGE" "${VERIFY_ARGS[@]}" && \
+   "$ZARF" package inspect sbom "$PACKAGE" \
```
shrinedogg, line 96: Note (no change requested): verified against v0.83.0 --help that with the keyless identity flags set, zarf auto-disables --insecure-ignore-tlog, so this verify path requires the Rekor inclusion proof embedded in the package; sign --keyless auto-enables --tlog-upload, so CI-built packages carry it. Consistent, but the first post-merge nightly run is the end-to-end confirmation for the keyless path.

shrinedogg (approved):
Reviewed and verified locally against the pinned zarf v0.83.0.

Verified locally: unit tests (19 pass), deps-check, inventory check, bash -n on all touched scripts, YAML parse of both workflows, and the full operator-key chain on a real fixture package (create, sign, verify, inspect sbom --verify=always, verify-sbom.py). Wrong-key probes fail closed on both verify and inspect. No dangling references to the moved tests or the old deploy flow.

Suggestions: two minor inline comments (hardcoded package filename at build-package.sh:113, TMPDIR-dependent evidence upload at offline-run.sh:88). One note on the keyless Rekor requirement.

Looks good: signing refuse-unsigned / refuse-both-modes gating; verification moved before any cluster mutation; OIDC grant scoped to the build job only; docs, AGENTS.md, and CI all updated consistently; the docs transfer table now matches where the build actually writes the package.

Merge note: rebased cleanly against current main (PR #93 has no overlap).

lmilbaum:
Does it make sense to migrate some or all of `build-package.sh` into the GitHub Actions workflow?

This change introduces GitHub-specific OIDC signing into a script that is also intended for operator-managed local builds. Before extending this design further, I think we should evaluate the broader architecture and its implications in a separate issue.

That evaluation should include:

- Where package creation and signing should run: locally, in CI, or both.
- Whether GitHub-specific keyless signing belongs in the portable build script or only in the workflow.
- Whether operator-key and keyless builds should share one implementation.
- How signed packages are published, retained, versioned, and discovered.
- Whether GitHub Actions artifacts are sufficient or durable release assets are needed.
- How a demo obtains and verifies an existing package without rebuilding it.
- How packages and trust material are transferred into the offline environment.
- Reproducibility and provenance differences between CI and operator builds.
- Which workflow identity, branch, and repository the offline verifier should trust.

I do not propose resolving this architecture in the current PR, but I would like us to open a follow-up issue so the build, signing, distribution, and offline-consumption model is designed intentionally.

shrinedogg:
Checked against the zarf v0.83.0 source. Partly yes, partly no.

What `inspect sbom --verify=always` already guarantees: `LoadPackage` verifies the signature and the checksum manifest, and `sboms.tar` is checksum-covered (only `zarf.yaml` is excluded). A missing SBOM fails with `NoSBOMAvailableError` (`src/pkg/packager/layout/package.go:469`), and a corrupt archive fails decompression. So missing and tampered are covered by the exit status.

What it does not guarantee: `GetSBOM` only extracts the tar. Zarf never parses the documents, so a package whose SBOM entries are not valid Syft JSON extracts cleanly and exits zero. That gap is what `verify-sbom.py` was covering.

Your second suggestion closes it without the custom script. `zarf tools sbom` is the full vendored syft CLI, so its `convert` subcommand runs syft's real decoder against each document and exits nonzero when a document does not decode as Syft JSON. That is a stronger check than the partial field test in `verify-sbom.py`, uses only the already-shipped binary, and works offline.

Proposed change: drop `verify-sbom.py` and `airgap/tests/test_verify_sbom.py`, and replace the step 1 validation in `offline-run.sh` with a loop over the extracted documents:

```sh
found=0
for doc in "$SBOM_OUTPUT"/*/*.json; do
  [ -e "$doc" ] || continue
  found=1
  "$ZARF" tools sbom convert "$doc" -o syft-json > /dev/null
done
[ "$found" -eq 1 ]
```

The only piece `zarf` does not cover is the "at least one document extracted" check, which the `found` flag keeps as one line. I will push that as a fix commit to this branch if you agree.

shrinedogg:
Pushed as 14edf51 (now the head of this PR): replaces `verify-sbom.py` and its tests with syft's own decoder in `offline-run.sh` step 1, and updates `docs/airgap.md`.

Verified locally with zarf v0.83.0 on macOS before pushing: a valid Syft document decodes (exit 0); invalid JSON and valid-JSON-wrong-schema both fail (exit 1, `failed to decode SBOM: sbom format not recognized`); the loop propagates failures explicitly since the script runs `set -uo pipefail` without `-e`. CI on the new head is green (unit tests, inventory check, overlays, YAML lint, Renovate digest pinning).

One correction to my earlier comment: it claimed the syft `convert` path might be a sandbox problem on macOS. That was wrong; I tested it after posting and it works, which is what this commit relies on.

Issue #95, opened 2026-08-29
Port the CAPI pivot (pivot.sh) into the Rust bootstrap CLI


shrinedogg:
## Problem

Since #79, `bootstrap.sh` does not end at the completion message: with `BOOTSTRAP_PIVOT` unset or 1 (the default), it execs `pivot.sh`, which moves the CAPI inventory into the self-managed management cluster and deletes the kind bootstrap cluster. The pivot is the default exit of the bootstrap.

The Rust port (`bootstrap-rs/`, PR #93, head 1c46991) has no pivot: no `BOOTSTRAP_PIVOT` handling, no `clusterctl move`. Its only `clusterctl` call is `get kubeconfig local-workload` in the Step 5 reconciliation watch. The binary currently reproduces only the `BOOTSTRAP_PIVOT=0` opt-out path, so a full run stops at the completion message and leaves kind as the management cluster.

This was flagged in the #93 review as the one structural gap, and #92 gates `bootstrap.sh` retirement on full-parity runs against current main, which now include the pivot.

## Goal

Make `knr-bootstrap` reproduce the full default flow of `bootstrap.sh`: bootstrap, then pivot, ending with a self-managed management cluster and no kind cluster (unless opted out or skipped).

## Proposed scope

1. Port the `pivot.sh` phases into `bootstrap-rs/src/main.rs`, preserving step order, messages, and exit behavior:
   - Phase 0 preflight: required tools, current context must be `kind-mgmt` (`BOOTSTRAP_KUBECONTEXT`), the management Cluster exists; aws: the `require_flux_env` checks; local-host: local registry reachable.
   - Phase 1: wait for the `<cluster>-kubeconfig` secret, polling every `MGMT_POLL_INTERVAL` within `MGMT_READY_TIMEOUT`.
   - Phase 2: export the kubeconfig to `MGMT_KUBECONFIG` (chmod 600), CAPD localhost endpoint rewrite, context rename to `knr-ops-mgmt`, node Ready wait.
   - Phase 3: imperative target installs: cert-manager and the CAPI operator at the `deps/versions.toml` versions (`cert_manager`, `capi_operator_chart`), provider CRs from `mgmt/<env>/capi-providers/`, the CAPA credentials secret from `mise -E aws run aws-credentials`, provider Ready waits.
   - Phase 4: suspend Flux Kustomizations in kind, `clusterctl move`, unpause moved Clusters, verify `AWSClusterControllerIdentity` (aws).
[...]

PR #78, opened 2026-08-27
refactor: retire central version catalog in favor of Renovate

shrinedogg:
## Summary

Final phase of #74 (stacked on #77, which is stacked on #76): retires the central version catalog and its synchronization machinery now that Renovate covers every native version surface.

- Deleted `deps/` entirely: `versions.toml`, `scripts/check.py`, `scripts/catalog.env.sh`.
- Key finding: exhaustive grep found zero consumers of `catalog.env.sh`; airgap scripts pin versions as literals directly, so the env emitter was dead machinery and no migration helper was needed.
- `mise.toml`: removed the `deps-check` task and its block in `validate`; added the new airgap unit tests to `validate`.
- `.github/workflows/validate.yml`: removed the "Dependency drift" step.
- `docs/dependencies.md`: rewritten for the Renovate model (managed surfaces, renovate.json5 links, dependency dashboard, update procedure); generated inventory table removed.
- `README.md` and `AGENTS.md`: dependency-management sections updated to describe Renovate; `deps/` removed from the layout tree.

## Coverage equivalence

Every former catalog concern is now handled by Renovate managers from #76: mise tools (per-tool custom managers), clusterctl providers, Flux/Helm/Kubernetes manifests, kindest/node images, `airgap/images.txt`, `airgap/zarf.yaml`, and GitHub Actions pins. EKS addon versions (`*-eksbuild.*`) remain manual in both models; they have no public registry datasource and are noted in the renovate.json5 comments.

## Verification

- `mise run validate`: PASS (bash -n, 16 unit tests, 38 kustomize overlays).
- `airgap/scripts/render-and-inventory.sh`: 18 rendered, 0 fail.
- `airgap/scripts/check-inventory.py`: all rendered images present.

Closes #74

## Required follow-up (repo admin)

- Add repo secret `RENOVATE_TOKEN` (fine-grained PAT, contents and pull request read/write on this repo only) and trigger the renovate workflow with dry-run enabled (see #76).

PR #96, opened 2026-08-29
chore: make Renovate manage the bootstrap-rs flux chart pin

shrinedogg:
## What

Makes Renovate manage the imperative chart pin in the Rust bootstrap binary, pulled forward from #95 (the pivot port) so the pin Renovate coverage gap closes now rather than waiting for the pivot work.

- `renovate.json5`: new customManager matching `FLUX_CHART_VERSION: &str = "..."` in `bootstrap-rs/src/main.rs` (docker datasource against the chart's OCI path `ghcr.io/controlplaneio-fluxcd/charts/flux-operator`, so the pin joins the existing `flux` package group and bumps alongside the Flux manifest pins).
- `deps/versions.toml`: new `flux_operator_chart` slot validating the literal in `main.rs` via `mise run deps-check`, so the catalog, the binary, and the HelmChartProxies cannot drift.
- `bootstrap-rs/tests/test-renovate-chart-pins.py`: integration test running real Renovate (`--platform=local --dry-run=lookup`, same harness and pinned CLI 44.50.1 as the airgap digest-pinning test) against the real `main.rs` and `renovate.json5`, asserting the pin is detected; wired as a step in the existing renovate job in `validate.yml`.
- `deps/scripts/check.py`: `bootstrap-rs/target` excluded from `--forbid` scanning. Reproduced before fixing: after `cargo build`, 4 false positives on the `helm` literal (4.2 embedded in `libtokio` rlib/rmeta artifact names) failed `mise run deps-check`. Build artifacts are not version pins.
- `docs/dependencies.md`: inventory table regenerated from the catalog; `flux_operator_chart` now annotated `(validated in bootstrap-rs/src/main.rs)`.

Cert-manager and CAPI-operator customManagers are deliberately not added yet: their constants do not exist until the #95 pivot port lands. The renovate.json5 header comment notes they join then.

## Why stacked on #93

The customManager matches a file that only exists on this branch (`bootstrap-rs/src/main.rs`), and this PR extends the same `deps/versions.toml` + `docs/dependencies.md` territory #93 already touches. Base is `feat/bootstrap-rs`; retarget to `main` when #93 merges.

## Validation

- Local: `deps-check --forbid` green (with the slot added and the target exclusion); both Renovate integration tests pass under Renovate 44.50.1 on Node 24 (the renovate job's Node version): chart-pin detection reports `ghcr.io/controlplaneio-fluxcd/charts/flux-operator`, airgap digest test unchanged (45 + 14 pins).
- Registry check: `0.58.0` confirmed present in the ghcr.io tags list (117 tags), so the datasource resolves; `0.58.1` already exists upstream, meaning Renovate will have a real bump to propose on its next scheduled run once this merges.
- CI on this PR covers everything: validate.yml runs on all PRs (deps-check, yaml lint), the renovate job runs the new test, bootstrap-rs.yml runs fmt/clippy/test on the unchanged Rust code (path-filtered but triggered by the test file living under `bootstrap-rs/`).

Refs #95. Stacked on #93.


deps/scripts/check.py
lmilbaum, line 1: I understand that deps/scripts/check.py is intended to enforce deps/versions.toml as the authoritative version catalog, but its value is becoming less clear as we add more file-specific slots, exclusions, and false-positive handling. In this PR we are adding both a new validation rule and a special exclusion for Rust build artifacts. Could you clarify the long-term design for this checker and why this complexity is preferable to generating consumers directly from the catalog or limiting validation to tracked source files?

deps/versions.toml
lmilbaum, line 1: I have the same concern about deps/versions.toml. It is described as the authoritative dependency catalog, but many consumers still hardcode the same versions, and check.py only verifies that the duplicated values match. Renovate also appears to update the consumer rather than the catalog. Could you clarify what value the catalog provides over managing versions directly in their consuming files? If it is intended to be the true source of truth, I would expect updates to start there and consumers to be generated from it, rather than maintaining the same value in multiple places.

docs/dependencies.md
lmilbaum, line 1: I have the same concern about docs/dependencies.md. It appears to be a generated copy of information already available in deps/versions.toml, which adds another file that must remain synchronized and creates extra review noise on dependency changes. Could you clarify who this document is intended for and what information it provides beyond the catalog itself? If it does not add meaningful context, I would prefer linking directly to deps/versions.toml instead of committing a generated duplicate.

renovate.json5
lmilbaum, line 1: This file feels overloaded with explanatory comments, which makes the actual configuration harder to scan. Some comments describe implementation history, future work, or PR context rather than information needed to maintain the configuration. That context may be better placed in the commit message, PR description, or relevant documentation. Could we keep only comments that explain non-obvious configuration decisions and remove the rest?

renovate.json5
```diff
@@ -164,6 +167,21 @@
       "extractVersionTemplate": "^v(?<version>.+)",
       "autoReplaceStringTemplate": "clusterawsadm = { version = \"{{newValue}}\""
     },
+    // ── bootstrap-rs imperative chart pins ────────────────────────────────
+    // The binary installs these charts before Flux exists, so the versions
+    // live as Rust constants in bootstrap-rs/src/main.rs rather than in
+    // HelmReleases the flux manager could read. The docker datasource
+    // resolves OCI chart refs by their registry path; depName carries the
+    // full path so these pins join the existing flux package group.
+    {
```
lmilbaum, line 172: Could we make this custom Renovate manager generic and annotation-driven instead of hardcoding FLUX_CHART_VERSION and the Flux Operator dependency in renovate.json5?
```
For example, the dependency metadata could live next to the version:
// renovate: datasource=docker depName=ghcr.io/controlplaneio-fluxcd/charts/flux-operator
const FLUX_CHART_VERSION: &str = "0.58.0";
```
The regex manager could then extract datasource, depName, and currentValue from the annotation and adjacent constant. This would establish a reusable convention for future bootstrap dependencies without requiring a separate manager configuration for every constant.
Renovate documents this annotation-driven custom-regex pattern here: [Custom regex manager — advanced capture](https://docs.renovatebot.com/modules/manager/regex/#advanced-capture).
The inline comment is not processed automatically in arbitrary Rust files, so a custom regex manager is still required; the suggestion is to make that manager generic rather than specific to this one dependency.

bootstrap-rs/tests/test-renovate-chart-pins.py
lmilbaum, line 1: This is now the second standalone Python integration test following a similar pattern. It may be worth considering a common testing approach—such as pytest, shared Python test utilities, or Rust-based integration tests—to reduce duplicated harness and parsing logic. I do not think that refactoring belongs in this PR, but it would be useful to track as a follow-up before more tests of this kind are added.

lmilbaum (changes requested):

docs/dependencies.md
shrinedogg: Fair challenge; answered rather than changed in this PR. The document predates this PR by a wide margin (the inventory table and its generation were introduced with the catalog itself, issue #62 and #74 territory) - this PR only regenerated it because a slot was added. Its intended audience is humans browsing docs/ who will not open a TOML file, and it carries non-generated sections (bump procedure, grouping rationale, the unpinned-by-decision notes). But your core point stands: a generated duplicate of the catalog table adds sync work and review noise on every dependency change. If the table were dropped in favor of linking to deps/versions.toml, the docs would lose nothing the catalog does not already say. That is a docs-wide decision rather than a bootstrap-rs one, so I have not changed it inside this stacked PR; I suggest folding it into the #74 catalog-retirement discussion (same open question about which artifacts deserve to exist). Happy to do the removal as its own small PR if you prefer it gone now.

renovate.json5
shrinedogg: Done in 9dde18e. The additions now carry one line each: the header version-sources list gained "annotated version constants in bootstrap-rs sources", and the manager keeps a three-line comment explaining the one non-obvious thing (metadata lives in source annotations, not in this file). The history/PR-context comments (why the pin exists, which future constants join, the exclusion rationale) moved to the commit message and PR description where they belong.

bootstrap-rs/tests/test-renovate-chart-pins.py
shrinedogg: Tracked as #97 before more tests of this kind appear, as you suggested. It records the three candidate conventions (pytest fixtures, a shared Python module keeping plain-python3 invocations, or Rust integration tests in the cargo workflow), the pieces to extract (fixture setup, env, subprocess invocation, JSON event parsing, diagnostics), and a hard requirement to preserve: CI must keep running the tests with Node 24, since Renovate 44.50.1 requires RegExp.escape. The #95 pivot constants will be the first consumers of whatever convention lands.

deps/scripts/check.py
shrinedogg: Addressed in 9dde18e along the lines you suggested: `iter_repo_files` now yields git-tracked files only (`git ls-files`), so validation is inherently limited to tracked source files; the `bootstrap-rs/target` exclusion I had added is gone entirely. Verified by running deps-check with real cargo build artifacts present in the worktree: no false positives, because untracked files are never scanned. On the long-term design: the per-file slots and exclusions are the cost of the validate-consumers approach, and the alternative you name (generate consumers from the catalog) is the direction #74 is already tracking - generation exists today for zarf.yaml, images.txt, and the airgap scripts, with validation-only as the interim for files where generation is awkward. This thread's push is noted there so the checker trends toward less per-file configuration, not more.

deps/versions.toml
shrinedogg: Answered rather than changed here. The catalog's value today is being the single place a human looks up "what version do we run" and the machine check that every consumer agrees with it; it is deliberately not yet the source consumers are generated from, and you are right that this makes it a consistency checker over duplicated values rather than a true source of truth. The end-state you describe - updates start in the catalog, consumers generated from it - is exactly what #74 exists to reach (its title is "Automate dependency updates with Renovate and retire the central version catalog"); some consumers are already generated (zarf.yaml, images.txt, airgap scripts) while shell scripts and Rust constants are validation-only for now, since generating into those formats buys little. The interim tension you spotted is real and known: Renovate updates the consumer, the slot check keeps the catalog in step, and PR review carries the pair. Whether that interim is worth it or the catalog should be retired sooner is the open question #74 owns; your comment strengthens the case for retiring it and I have linked this thread there.

renovate.json5
shrinedogg: Addressed in 9dde18e. The customManager is now generic and annotation-driven exactly as you suggested: datasource and depName live next to the constant in the Rust source as `// renovate: datasource=docker depName=ghcr.io/controlplaneio-fluxcd/charts/flux-operator`, and one manager for `bootstrap-rs/src/*.rs` extracts datasource, depName, optional versioning, and currentValue from the annotation plus the adjacent `const NAME: &str = "..."` line (Renovate's advanced-capture pattern; RE2 constraints respected: no lookarounds, explicit line handling). The replacement template rewrites the annotation and the constant together, so bumps preserve the metadata. Verified live: a dry-run against a fixture with the pin lowered to 0.57.0 proposes 0.58.1 grouped on the `renovate/flux` branch, and the detection test still passes against the real main.rs. When #95 adds CERT_MANAGER_VERSION and CAPI_OPERATOR_VERSION, they need only their own annotation comments, no further renovate.json5 changes.

shrinedogg:
Overlap with #99: this PR's Renovate coverage is included there, and its catalog parts are obsolete.

- The annotation-driven customManager (regex over `bootstrap-rs/src/*.rs` reading `// renovate: datasource=... depName=...` comments) and the `FLUX_CHART_VERSION` annotation in main.rs are carried into #99 as-is.
- #99 also adds `tests/test-renovate-coverage.py` (wired into CI), which generalizes this PR's `bootstrap-rs/tests/test-renovate-chart-pins.py` to also cover the pivot.sh chart pins and the Renovate CLI pin.
- The `deps/versions.toml` `flux_operator_chart` slot and the check.py git-ls-files hardening are moot: an audit found that #89's stale base resurrected the catalog that #78 had retired (so Renovate bumps and the drift check would fight over the same literals), and #99 retires it again.
- Broader context from the same audit: Renovate had never actually run because renovate.yml set `RENOVATE_REPO` (not a valid option) instead of `RENOVATE_REPOSITORIES`; #99 fixes that, and a full-repo extract with the pinned Renovate 44.50.1 shows 55 files with extractable pins and zero errors.

Suggest closing this once #99 merges rather than merging both, so the catalog slot does not come back with this branch (the same stale-base mechanism that resurrected deps/ after #78).

Issue #100, opened 2026-08-29
Port teardown.sh into the Rust CLI


shrinedogg:
## Problem

`teardown.sh` is 977 lines of bash and the largest imperative surface left once the bootstrap (#92) and pivot (#95) ports land. It mixes nine ordered destruction steps with cross-region AWS orphan cleanup (pod identity associations, nodegroups, EKS, RDS, versioned S3 purge, VPC resource ordering, IAM roles and users, the CloudFormation bootstrap stack), best-effort semantics on every call, and the EXIT-trap deletion guard (`CLUSTERS_CONFIRMED_GONE` / `FORCE_KIND_DELETE`) that keeps the CAPI controller alive until clusters are confirmed gone. It has the testability, error-handling, and portability problems that motivated #92, with higher stakes: a silent skip or wrong ordering orphans real cloud resources.

It also has post-#79 semantics to design regardless of language: after the pivot, step 2's `kubectl get clusters -A` runs against the self-managed management cluster and includes the management cluster itself, so the sweep deprovisions the cluster the teardown runs from, and the deletion guard's assumption (controller hosted in kind) no longer holds. The script predates the pivot entirely (last touched at #34; #79 is still open).

## Goal

Port `teardown.sh` into the Rust CLI with parity step order, guards, refusal messages, and skip-if-absent behavior, resolving the post-pivot semantics as part of the design.

## Proposed scope

1. Port the nine steps and both environment preflights; the AWS orphan sweep as typed units keeping the per-resource best-effort semantics; the cluster-aware deletion guard replacing the EXIT trap, preserving the `CLUSTERS_CONFIRMED_GONE` / `FORCE_KIND_DELETE` contract and `AWS_ONLY` mode.
2. Entry shape: `teardown` subcommand of `knr-bootstrap` vs a separate binary; `mise run teardown` wiring.
3. AWS interaction: shell out to the `aws` CLI for parity with the rest of the port, or use an SDK.
4. Tests: unit tests for pure helpers (per-region resource name derivation, guard state transitions), failure-path probes against the built binary.
5. Docs and AGENTS.md updates.

## Open decisions

- Entry shape (subcommand vs separate binary).
- Post-pivot semantics: deleting the self-managed management cluster from itself; whether the deletion guard generalizes past kind; `AWS_ONLY` in the self-managed world.
- Ordering vs #98: the port grows the constant surface (region lists, IAM names, bucket patterns) that #98 externalizes.

## Acceptance criteria

[...]

Issue #98, opened 2026-08-29
Externalize knr-ops-specific configuration from the Rust bootstrap CLI


shrinedogg:
## Problem

`bootstrap-rs/src/main.rs` compiles knr-ops-specific values into the binary as constants: the kind cluster name `mgmt` (and the `kind-mgmt` context), `REGISTRY_NAME = "knr-registry"`, `GIT_BRANCH = "main"`, `FLUX_CHART_VERSION = "0.58.0"`, the FluxInstance sync paths `mgmt/aws` and `mgmt/local-host`, the `flux-system` namespace and secret names (`flux-github-pat`, `sops-age`), and the profile set itself (the `Profile` enum). The runtime knobs (`GIT_REPO_URL`, `REGISTRY_PORT`, `OCI_REPOSITORY`, `AGE_KEY_FILE`, and the rest of `Config::from_env`) are already externalized through the environment, but everything that describes this repository's layout is compile-time.

This couples the binary to knr-ops more tightly than necessary. It also means chart pins the binary installs imperatively are maintained as annotated Rust source constants via Renovate customManagers (#95 scope item 6), instead of being read from the repository the binary is bootstrapping. (#99 has since retired `deps/versions.toml`; the annotations are now the only pin slot outside the Git manifests.)

The pivot port (#95) adds more of the same surface: `CERT_MANAGER_VERSION`, `CAPI_OPERATOR_VERSION`, per-environment management cluster names, and provider lists. The config surface should be designed once, after that lands.

## Goal

Make `knr-bootstrap` a generic bootstrap engine that reads everything repository-specific from a declarative config file owned by the target repository, so knr-ops is its first consumer rather than its definition, and a future extraction or reuse of the binary requires no code changes.

## Proposed scope

1. Define a config file at the repo root (for example `bootstrap.toml`) that carries the values listed above, with environments as config sections replacing the compiled `Profile` enum. The existing env knobs keep working and take precedence over file values, preserving the current override surface.
2. Read chart versions the binary installs imperatively (`flux_operator_chart`; after #95 also `cert_manager` and `capi_operator_chart`) from `bootstrap.toml` instead of compiling them in. Each version then exists in both the TOML and the Git HelmRelease/HelmChartProxy manifests; `mise run validate` cross-checks that they agree, so no version literal is hand-maintained in Rust source.
[...]

PR #101, opened 2026-08-29
feat: port the CAPI pivot (pivot.sh) into the Rust bootstrap CLI

shrinedogg:
## Summary

Ports `pivot.sh` into `knr-bootstrap` (#95): the pivot is now the bootstrap binary's default exit, not just the script's. After the completion message, unless `BOOTSTRAP_PIVOT=0`, the binary runs the six pivot phases and ends self-managed with the kind cluster deleted — the same flow as `bootstrap.sh` exec-ing `pivot.sh`.

Per the entry-point decision on #95: no subcommand, no new CLI surface. The standalone rerun path is a plain binary rerun (reuse-by-default cluster validation, rerun-safe bootstrap steps, re-runnable move).

## What

- **Pivot phases 0–6** with step order, messages, and guards preserved: context/Cluster/registry preflight; `<cluster>-kubeconfig` secret poll (`MGMT_READY_TIMEOUT`, default 40m aws / 15m local-host); kubeconfig export (chmod 600, CAPD localhost endpoint rewrite, context rename to `knr-ops-mgmt`, node Ready wait); imperative target installs (cert-manager, CAPI operator, provider CRs verbatim from `mgmt/<env>/capi-providers/`, CAPA credentials secret from `mise -E aws run aws-credentials` — via stdin manifest, off argv); Flux suspend + `clusterctl move` + unpause (with the script's recovery guidance on failure); Flux seeding against the target (oci-push first for local-host); guarded kind deletion (`PIVOT_SKIP_DELETE`, Clusters-present guard, local-host registry guard).
- **Env knobs with script semantics**: `BOOTSTRAP_PIVOT`, `PIVOT_SKIP_DELETE`, `MGMT_KUBECONFIG`, `MGMT_READY_TIMEOUT`, `MGMT_POLL_INTERVAL`, `BOOTSTRAP_KUBECONTEXT`. Pinned env-only by `cli_rejects_deferred_flags`.
- **Target-aware helpers** (#95 scope item 3): `install_flux_operator`, `create_aws_secrets`, `install_flux_instance`, `kubectl_apply` accept an optional kubeconfig; the seed sequence `pivot_seed_target` reuses them against the management cluster.
- **Renovate-owned pins**: `CERT_MANAGER_VERSION` / `CAPI_OPERATOR_VERSION` carry `// renovate:` annotations (helm datasource with explicit `registryUrl`, grouped into `platform-charts` by the existing packageRule); the generic annotation-driven customManager learns the optional `registryUrl=` key; the coverage test now expects all three pins in `main.rs`.

## Intentional divergences from the script

- The pivot's helm installs use `upgrade --install` instead of `install`: rerun-safe after a partial failure, same policy as the #93 review decisions. Same charts, same pinned versions, same values.
- The CAPA credentials secret is applied as a stdin manifest rather than `kubectl create secret --from-literal`: keeps secret material off argv (house rule), same `stringData` shape.
- `clusterctl` and `mise` become required on both profiles (the pivot invokes them; previously aws required neither).

## Validation

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` (23 tests, 10 new): clean.
- `python3 tests/test-renovate-coverage.py`: Renovate 44.50.1 dry-run detects all three pins in `bootstrap-rs/src/main.rs` (`flux-operator`, `cert-manager`, `cluster-api-operator`) plus `pivot.sh`'s two. (Needs Node >= 24 locally; the CI job pins renovate 44.50.1.)
- Failure-path probes against the built binary: invalid profile (exit 2), non-numeric `MGMT_POLL_INTERVAL` (exit 1, names the variable), `--no-pivot`/`--pivot`/`--mgmt-kubeconfig` rejected (exit 2), pivot knobs parse and flow to preflight.
- `mise run validate`: passes.
- Full bootstraps+pivot deliberately not run here: they would touch the running `mgmt` kind cluster and provision real infrastructure. Both environments are part of the #92 acceptance runs before `bootstrap.sh`/`pivot.sh` retire.

## Not in this PR

- `teardown.sh` port: filed as #100.
- Config externalization of the new constants: #98 owns it (blocked on this PR finalizing the surface).
- `mise run bootstrap` still calls `bootstrap.sh`; the task wiring flips when the scripts retire under #92.

Refs #95.

### DIFF
--- a/bootstrap-rs/src/main.rs
+++ b/bootstrap-rs/src/main.rs
@@ -17,7 +17,7 @@ use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::time::Duration;
 
-use anyhow::{bail, Context, Result};
+use anyhow::{bail, ensure, Context, Result};
 use clap::{Parser, ValueEnum};
 use serde_json::json;
 use tokio::io::AsyncWriteExt;
@@ -195,6 +195,24 @@ impl Config {
         )
         .parse::<u32>()
         .context("REGISTRY_READY_RETRIES must be a non-negative integer")?;
+        // Pivot knobs are validated here, not in pivot Phase 1, so a bad
+        // value fails at startup instead of after the full bootstrap.
+        let mgmt_ready_timeout = value("MGMT_READY_TIMEOUT").unwrap_or_else(|| match profile {
+            Profile::Aws => DEFAULT_MGMT_READY_TIMEOUT_AWS.to_string(),
+            Profile::LocalHost => DEFAULT_MGMT_READY_TIMEOUT_LOCAL.to_string(),
+        });
+        parse_duration_seconds(&mgmt_ready_timeout)
+            .context("MGMT_READY_TIMEOUT must be a duration (40m, 2h, 90s, or bare seconds)")?;
+        let mgmt_poll_interval = with_default(
+            "MGMT_POLL_INTERVAL",
+            &DEFAULT_MGMT_POLL_INTERVAL.to_string(),
+        )
+        .parse::<u64>()
+        .context("MGMT_POLL_INTERVAL must be a positive integer")?;
+        ensure!(
+            mgmt_poll_interval > 0,
+            "MGMT_POLL_INTERVAL must be a positive integer, got 0"
+        );
         Ok(Config {
             profile,
             recreate: cli.recreate,
@@ -220,16 +238,8 @@ impl Config {
             mgmt_kubeconfig: value("MGMT_KUBECONFIG")
                 .map(PathBuf::from)
                 .unwrap_or_else(default_mgmt_kubeconfig),
-            mgmt_ready_timeout: value("MGMT_READY_TIMEOUT").unwrap_or_else(|| match profile {
-                Profile::Aws => DEFAULT_MGMT_READY_TIMEOUT_AWS.to_string(),
-                Profile::LocalHost => DEFAULT_MGMT_READY_TIMEOUT_LOCAL.to_string(),
-            }),
-            mgmt_poll_interval: with_default(
-                "MGMT_POLL_INTERVAL",
-                &DEFAULT_MGMT_POLL_INTERVAL.to_string(),
-            )
-            .parse::<u64>()
-            .context("MGMT_POLL_INTERVAL must be a positive integer")?,
+            mgmt_ready_timeout,
+            mgmt_poll_interval,
             bootstrap_kubecontext: with_default(
                 "BOOTSTRAP_KUBECONTEXT",
                 DEFAULT_BOOTSTRAP_KUBECONTEXT,
@@ -2220,6 +2230,30 @@ mod tests {
         })
         .unwrap_err();
         assert!(err.to_string().contains("MGMT_POLL_INTERVAL"));
+    }
+
+    #[test]
+    fn config_rejects_zero_poll_interval() {
+        // 0 parses as u64 but divides by zero in the Phase 1 attempt
+        // arithmetic; reject it at startup like any other invalid value.
+        let cli = Cli::try_parse_from(["knr-bootstrap", "aws"]).unwrap();
+        let err = Config::from_env(&cli, |name| match name {
+            "MGMT_POLL_INTERVAL" => Some("0".into()),
+            _ => None,
+        })
+        .unwrap_err();
+        assert!(err.to_string().contains("MGMT_POLL_INTERVAL"));
+    }
+
+    #[test]
+    fn config_rejects_invalid_ready_timeout() {
+        let cli = Cli::try_parse_from(["knr-bootstrap", "aws"]).unwrap();
+        let err = Config::from_env(&cli, |name| match name {
+            "MGMT_READY_TIMEOUT" => Some("abc".into()),
+            _ => None,
+        })
+        .unwrap_err();
+        assert!(err.to_string().contains("MGMT_READY_TIMEOUT"));
     }
 
     #[test]
--- a/bootstrap-rs/src/main.rs
+++ b/bootstrap-rs/src/main.rs
@@ -1,7 +1,12 @@
 //! knr-bootstrap – One-time imperative bootstrap for the management cluster.
 //! Everything after this program runs is driven by GitOps (Flux).
 //!
-//! Rust port of `bootstrap.sh`. The CLI surface is the script's surface:
+//! Rust port of `bootstrap.sh` including its default exit: unless
+//! BOOTSTRAP_PIVOT=0, the bootstrap continues into the port of `pivot.sh`
+//! (issue #95), which moves the CAPI inventory into the self-managed
+//! management cluster and then deletes the kind bootstrap cluster.
+//!
+//! The CLI surface is the script's surface:
 //! a positional profile, `--recreate`, and the environment. Unlike the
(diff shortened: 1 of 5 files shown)


bootstrap-rs/src/main.rs
```diff
@@ -1162,6 +1345,648 @@ async fn watch_local_reconciliation(cfg: &Config, engine: &str) -> Result<()> {
...
+/// waited on instead of erroring immediately. Node readiness is verified in
+/// Phase 2 against the exported kubeconfig.
+async fn pivot_wait_for_management_cluster(cfg: &Config, mgmt_cluster: &str) -> Result<()> {
+    println!(
+        ">>> Waiting for the management cluster kubeconfig (timeout: {})...",
+        cfg.mgmt_ready_timeout
+    );
+    let secret = format!("secret/{mgmt_cluster}-kubeconfig");
+    let timeout_s = parse_duration_seconds(&cfg.mgmt_ready_timeout)?;
+    let max_attempts = (timeout_s / cfg.mgmt_poll_interval).max(1);
```
shrinedogg, line 1429: Warning: `MGMT_POLL_INTERVAL=0` passes `parse::<u64>()` in from_env (the error text says "must be a positive integer", but 0 is accepted) and then panics here on division by zero. The panic surfaces in pivot Phase 1, after the entire bootstrap has run. Suggest rejecting 0 at parse time, e.g. check the parsed value is > 0 in the same `context` guard.

bootstrap-rs/src/main.rs
```diff
@@ -1162,6 +1345,648 @@ async fn watch_local_reconciliation(cfg: &Config, engine: &str) -> Result<()> {
...
+    .is_err()
+    {
+        eprintln!();
+        eprintln!("ERROR: clusterctl move failed.");
+        eprintln!("       The move is re-runnable: objects are deleted from the source only");
+        eprintln!("       after they were created on the target, so kind stays authoritative.");
+        eprintln!("       NEVER delete moved Cluster / AWSManaged* / MachinePool / Dev* objects");
+        eprintln!("       on the target to work around a failure — the provider would");
+        eprintln!("       deprovision the real infrastructure. See docs/operations.md");
+        eprintln!("       'Pivot recovery'.");
```
shrinedogg, line 1774: Warning: docs/operations.md has no "Pivot recovery" section (the file contains no pivot content at all). This dangles in pivot.sh already, so it is pre-existing, but the port carries it into the message users see mid-failure with real infrastructure at stake. Worth adding the section or fixing the pointer, here or as a tracked follow-up.

bootstrap-rs/src/main.rs
```diff
@@ -1162,6 +1345,648 @@ async fn watch_local_reconciliation(cfg: &Config, engine: &str) -> Result<()> {
...
+/// endpoint). Poll rather than kubectl-wait so a not-yet-created secret is
+/// waited on instead of erroring immediately. Node readiness is verified in
+/// Phase 2 against the exported kubeconfig.
+async fn pivot_wait_for_management_cluster(cfg: &Config, mgmt_cluster: &str) -> Result<()> {
+    println!(
+        ">>> Waiting for the management cluster kubeconfig (timeout: {})...",
+        cfg.mgmt_ready_timeout
+    );
+    let secret = format!("secret/{mgmt_cluster}-kubeconfig");
+    let timeout_s = parse_duration_seconds(&cfg.mgmt_ready_timeout)?;
```
shrinedogg, line 1428: Suggestion: an invalid `MGMT_READY_TIMEOUT` (e.g. `abc`) passes from_env and only fails here, in pivot Phase 1 after the full bootstrap. MGMT_POLL_INTERVAL is validated at parse time; validating the timeout in from_env too would be consistent and fail-fast.

shrinedogg (commented):
**Verdict: looks good, comments only** (2 warnings, 1 suggestion; nothing blocking)

Port is phase-faithful to pivot.sh: phase order, messages, guards, and patch payloads match, and the documented divergences (helm upgrade --install, stdin secret manifest, clusterctl/mise required on both profiles) are sound.

Verified locally (head e05320c):
- cargo fmt --check, cargo clippy --all-targets -- -D warnings: clean
- cargo test: 23 passed, 0 failed
- Renovate regex: reconstructed the new matchStrings pattern and confirmed all three pins in bootstrap-rs/src/main.rs match, with registryUrl captured for cert-manager and cluster-api-operator (local Node is v22, so no full dry-run here; the CI job ran it with Node 24 and is green)
- Failure-path probes against the built binary: invalid profile exits 2, non-numeric MGMT_POLL_INTERVAL exits 1 and names the variable, --no-pivot / --pivot / --mgmt-kubeconfig all rejected with exit 2
- mise run validate: passes
- Version pins match Git: cert-manager 1.21.1 and capi-operator 0.28.0 HelmReleases, crds.enabled=true and cert-manager.enabled=false values match the --set flags, release names and target namespaces match (Flux adoption claim holds)
- All 13 Git manifest paths referenced by provider_manifests, plus mgmt/aws/infrastructure/aws-identity/identity.yaml, exist
- BOOTSTRAP_PIVOT semantics match bootstrap.sh's `[ "${BOOTSTRAP_PIVOT:-1}" = 1 ]` exactly
- Did not run a live bootstrap+pivot: a real mgmt kind cluster is running on this machine, consistent with the PR's stated validation boundary

Warning: MGMT_POLL_INTERVAL=0 is accepted by config parsing and panics on division by zero in pivot Phase 1 (inline).

Warning: the move-failure guidance points at docs/operations.md 'Pivot recovery', which does not exist (pre-existing in pivot.sh, carried over; inline).

Suggestion: validate MGMT_READY_TIMEOUT in from_env like MGMT_POLL_INTERVAL, so a typo fails fast instead of after the full bootstrap (inline).

Looks good: the target-aware helper refactor (optional kubeconfig threading) is clean, the env-knob semantics match the scripts precisely, and test coverage of the new surface is thorough.

bootstrap-rs/src/main.rs
shrinedogg: Fixed in acb4305: `from_env` now rejects 0 in the same guard (`MGMT_POLL_INTERVAL must be a positive integer, got 0`), so it exits 1 at startup instead of panicking in Phase 1. Verified against the built binary; pinned by `config_rejects_zero_poll_interval`.

bootstrap-rs/src/main.rs
shrinedogg: Added the section in 564fac1: docs/operations.md now has 'Pivot recovery' between the bootstrap verification and Teardown, covering the pivot lifecycle, the re-runnable move, the recovery steps, and the never-delete-moved-objects warning the error message promises. The title matches the pointer verbatim, so this also resolves the pre-existing dangle in pivot.sh (both its header comment and Phase 4 message).

bootstrap-rs/src/main.rs
shrinedogg: Done in acb4305: `from_env` now runs `parse_duration_seconds` on the resolved `MGMT_READY_TIMEOUT` (including the profile default) and fails at startup with `MGMT_READY_TIMEOUT must be a duration (40m, 2h, 90s, or bare seconds)` plus the underlying parse error as the cause. Pinned by `config_rejects_invalid_ready_timeout`; verified against the built binary (exit 1 at config load, before any side effect).

PR #106, opened 2026-08-30
feat: add bootstrap.toml and cross-check it against the Git manifests

shrinedogg:
First PR of the issue #98 implementation (plan: schema-first, consumption second, docs third). This lands the contract only: the config file, plus the guardrail that keeps it honest. No Rust changes; the binary does not read the file yet.

## What this adds

- **`bootstrap.toml`** (repo root): the repository-owned bootstrap configuration. `[bootstrap]` carries the global names (git branch, kind cluster/context, registry, Flux namespace, secret names, mgmt namespace/context); `[charts]` carries the three imperative chart pins with `# renovate:` annotations; `[environments.local-host]` / `[environments.aws]` carry the per-environment values (kind, sync path, mgmt cluster name, ready timeout, infra provider, provider-manifest list). Environment order in the file preserves bootstrap.sh's error-message wording (`local-host` before `aws`); a follow-up PR makes the binary read this and delete the compiled `Profile` enum.
- **`tests/test-bootstrap-config.py`**: cross-check failing validate/CI when (a) a `[charts]` pin disagrees with the version in any of the six Git manifest files, (b) a declared `provider-manifests` path or `sync-path` does not exist, or (c) the TOML does not parse (Python 3.11+ tomllib; mise python is 3.13, CI is 3.11+).
- **`mise.toml` / `.github/workflows/validate.yml`**: the cross-check runs in `mise run validate` and as a `validate` workflow step.

## Review focus

The schema is the reviewable part here (section names, the `kind` discriminator, field granularity); PR 2 will be mechanical against whatever this PR approves. Every value was verified equal to today's compiled constants in `bootstrap-rs/src/main.rs` (lines 26-76, 348-375) and every manifest path exists on disk.

## Verification

- `mise exec -- python3 tests/test-bootstrap-config.py` -> `bootstrap.toml cross-check OK (3 charts, 2 environments)`, exit 0.
- Mutation test: `cert-manager = "9.9.9"` -> FAILED listing both cert-manager manifests, exit 1; restored -> OK, exit 0.
- `mise run validate` -> exit 0 with the `==> bootstrap.toml cross-check` / OK lines; all kustomize builds pass.
- No `mgmt/**` or `workload/**` changes; konflate should render no diff.

## Interim-window note

Until PR 2 lands, chart pins are dual-homed: `main.rs` annotations remain Renovate-authoritative and the TOML annotations are inert (no Renovate manager targets them yet). If a Renovate chart-bump PR merges in between, its body must also bump `bootstrap.toml` or the cross-check fails CI on main.

Refs: #98

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -79,6 +79,9 @@ jobs:
           RENOVATE_GITHUB_COM_TOKEN: ${{ github.token }}
         run: python3 tests/test-renovate-coverage.py
 
+      - name: bootstrap.toml cross-check
+        run: python3 tests/test-bootstrap-config.py
+
   yaml-lint:
     name: Lint YAML
     runs-on: ubuntu-latest
--- a/bootstrap.toml
+++ b/bootstrap.toml
@@ -8,6 +8,10 @@
 #
 # Environment order matters: error messages list environments in file order
 # to preserve bootstrap.sh's exact wording ('local-host' before 'aws').
+#
+# Profile resolution: the [environments.*] section name is authoritative;
+# KNR_OPS_PROFILE matches it exactly. Each environment's `kind` restates
+# the section name, and the cross-check fails if the two disagree.
 
 [bootstrap]
 default-environment = "aws"
@@ -26,6 +30,8 @@ mgmt-context = "knr-ops-mgmt"
 # with the HelmReleases/HelmChartProxies in mgmt/<env>/ so Flux adopts the
 # installs without drift; `mise run validate` cross-checks them and Renovate
 # updates both sides together.
+# flux-operator and flux-instance version together upstream, so this one
+# pin serves both chart installs (FLUX_CHART_VERSION in main.rs).
 # renovate: datasource=docker depName=ghcr.io/controlplaneio-fluxcd/charts/flux-operator
 flux-operator = "0.58.0"
 # renovate: datasource=helm registryUrl=https://charts.jetstack.io depName=cert-manager
--- a/bootstrap.toml
+++ b/bootstrap.toml
@@ -1,0 +1,66 @@
+# Repository-owned configuration for knr-bootstrap (issue #98).
+#
+# Everything knr-ops-specific the binary needs is declared here; the binary
+# itself is a generic bootstrap engine. Environment variables (REGISTRY_PORT,
+# GIT_REPO_URL, BOOTSTRAP_KUBECONTEXT, ...) keep their precedence over these
+# values at runtime. Locate this file with BOOTSTRAP_CONFIG when not running
+# from the repository root.
+#
+# Environment order matters: error messages list environments in file order
+# to preserve bootstrap.sh's exact wording ('local-host' before 'aws').
+
+[bootstrap]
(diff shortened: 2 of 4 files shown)


shrinedogg (commented):
## Code Review Summary

**Verdict: Comment - contract verified value-for-value; two schema notes before PR 2 builds on it**

Verified locally (head 26ca2a9):

- Every TOML value checked against the compiled constants in bootstrap-rs/src/main.rs at this head (post-#101): all three chart pins (0.58.0 / 1.21.1 / 0.28.0), git branch, registry name, namespaces, secret names, both mgmt cluster names, both ready timeouts, both infra-provider tuples, and all 12 provider-manifest paths (the lists match provider_manifests() exactly, including caaph-system's addon-provider.yaml on aws).
- Cross-check exercised in all three failure directions: TOML pin drift fails naming both cert-manager manifests, manifest drift fails naming the drifted file, and a missing provider-manifest path fails naming it. Restores return to OK; tree left clean.
- `mise run validate` exits 0 with the cross-check integrated; all five CI checks green; konflate shows no rendered changes, matching the no mgmt/workload claim.
- flux-instance.yaml's `version: "2.x"` is the Flux distribution pin, a different concept from the chart pins, and is correctly outside the cross-check's scope.

Suggestion (schema): the `kind` field duplicates its section name ([environments.local-host] kind = "local-host"). As written the two can silently disagree, and nothing in this PR says which is authoritative for profile resolution. If PR 2 deletes the Profile enum and resolves KNR_OPS_PROFILE through this file, document whether it matches on the section name or on kind; otherwise drop kind and key purely on section names.

Suggestion (schema): the flux-operator chart pin serves both the flux-operator and flux-instance chart installs (main.rs uses one FLUX_CHART_VERSION for both, since the charts version together upstream). The [charts] comment does not say that; without it, a future reader may add a separate flux-instance key and desynchronize the pair. One comment line mirroring main.rs's own note prevents it.

Note: the cross-check's scope (chart pins, path existence) is right for the contract-first stage; the [bootstrap] names have no Git-side consumers to check against until PR 2, which is also when the cross-check starts earning its keep on every bump PR. The interim-window note in the PR (main.rs annotations stay Renovate-authoritative until then) is accurate.

Looks good: the file is a faithful contract of what exists today, the guardrail fails loudly in every direction I could construct, and landing schema before consumption is the right order for #98.

shrinedogg:
Addressed in 5756397 (both schema suggestions):

- **kind vs section name**: documented in `bootstrap.toml` that the `[environments.*]` section name is authoritative and `KNR_OPS_PROFILE` matches it exactly; `kind` restates the name. Disagreement is no longer silent: the cross-check now fails with `environments.<name> kind '...' must match its section name`. Mutation-tested (`kind = "local-host"` under `[environments.aws]` fails, restore passes). PR 2 resolves `KNR_OPS_PROFILE` against section names.
- **flux-operator pin scope**: added the `[charts]` comment line mirroring main.rs's note: the two charts version together upstream, so the one pin serves both installs (`FLUX_CHART_VERSION`).

`mise run validate` exit 0 locally; all five checks green on the new head.

Issue #105, opened 2026-08-30
Single-node Talos Linux management cluster environment (local-talos) via CAPI pivot


shrinedogg:
## Problem

The pivot target for the local environment is `local-management`, a Docker-container cluster built by CAPD from the ClusterClass in `mgmt/local-host/clusters/management/` (KubeadmControlPlaneTemplate + DevMachineTemplate). CAPD is development-grade by design: the management cluster that owns the CAPI inventory after the pivot runs as containers on the operator's laptop, with kubeadm-managed nodes and no path to a hardened, immutable OS. There is no environment where the pivot lands on a real machine without going to AWS.

## Goal

Add a `local-talos` environment alongside `local-host`: the same kind-bootstrap-then-pivot flow, but the pivot target is a single-node [Talos Linux](https://www.talos.dev/) cluster PXE-booted onto operator-provided bare metal by [Cluster API Provider Tinkerbell](https://github.com/tinkerbell/cluster-api-provider-tinkerbell) (CAPT), with the Talos bootstrap and control plane providers supplied by the maintained forks [shrinedogg/cluster-api-bootstrap-provider-talos](https://github.com/shrinedogg/cluster-api-bootstrap-provider-talos) (CABPT) and [shrinedogg/cluster-api-control-plane-provider-talos](https://github.com/shrinedogg/cluster-api-control-plane-provider-talos) (CACPPT). The existing CAPD reference path stays intact.

## Proposed scope

1. New environment tree `mgmt/local-talos/` mirroring the `mgmt/local-host/` layout (`clusters/management/`, `capi-providers/`, `addons/`, `infrastructure/`), a `mise.local-talos.toml` task layer, and the qualified environment label (`knr-ops.polarsquad.com/environment: local-talos`, per #87).
2. Provider CRs following the pattern in `docs/extending.md`: `BootstrapProvider` (talos) in `cabpt-system`, `ControlPlaneProvider` (talos) in `cacppt-system`, `InfrastructureProvider` (tinkerbell) in `capt-system`, alongside `capi-system`; registered via `flux-ks.yaml` with `dependsOn` ordering like the existing `capd-system`. The Talos provider CRs use `fetchConfig` URLs pointing at the forks' release assets. Prerequisite: the forks publish no releases yet; their release pipelines (`bootstrap-components.yaml` / `control-plane-components.yaml` and images, per the forks' registry-repointing commits) must produce consumable assets first.
[...]

PR #107, opened 2026-08-30
feat: consume bootstrap.toml from the Rust binary

shrinedogg:
Second PR of the #98 implementation (schema #106 → consumption here → docs third). The binary now reads `bootstrap.toml`; the compiled constants and the `Profile` enum are deleted.

## What changes

- **`bootstrap-rs/src/config.rs`** (new): parses and validates `bootstrap.toml` (serde + toml, `preserve_order` so error messages list environments in file order, preserving bootstrap.sh's `'local-host' before 'aws'` wording). `BOOTSTRAP_CONFIG` overrides the default `./bootstrap.toml` location (env-only, per the #95 entry-point decision). Validation rejects unknown fields, `kind`/section-name disagreement, an empty environments table, and a `default-environment` that names no section.
- **`bootstrap-rs/src/main.rs`**: everything repository-specific now resolves from the config:
  - chart pins (`flux-operator` serves both flux-operator and flux-instance installs, per the [charts] comment), `git-branch`, registry/kind names and contexts, Flux namespace, secret names, mgmt namespace/context
  - per-environment `sync-path`, `mgmt-cluster`, `mgmt-ready-timeout` (env override unchanged), `provider-manifests`, infra provider
  - `KNR_OPS_PROFILE` matches `[environments.*]` section names — the authority PR #106 documented; unknown names fail with the script's exact message, names listed in file order
  - `mise -E <name>` invocations derive from the resolved section name
- **`[[environments.<name>.move-fallbacks]]`** (new TOML key, per review discussion): objects that normally travel with the `clusterctl move` but are re-applied from the checkout when missing afterward. Replaces the one hardcoded `AWSClusterControllerIdentity` path; the cross-check verifies the declared manifest exists. Absent key = no fallback (local-host unaffected). Chosen over a documented exception so the binary stays generic for #105's third environment.
- **`renovate.json5`**: the annotation-driven manager retargets from Rust source constants to `bootstrap.toml` (`# renovate:` comments now match `key = "value"` lines). Same PR as the constant removal, so pin coverage never lapses — this also closes the dual-homing window noted in #106's interim note.
- **Tests**: `tests/test-renovate-coverage.py` EXPECTED map follows the retarget. Cargo tests grew from 24 to 34: config parsing/validation, env-over-file precedence, per-environment resolution (including the file-order error message), move-fallbacks, and a test that the checked-in `bootstrap.toml` parses and carries the expected layout (mutation-tested: drifting `mgmt-cluster` fails it).

## Deliberate exceptions (reviewed with the user)

- `"aws"` / `"local-host"` remain as literals in main.rs: they are the binary's sequence archetypes (which steps run: git vs OCI sync, aws preflight, reconciliation watch), per issue #98 decision 3. #105 (local-talos) will extend behavior in Rust, not just data.
- `"flux-system"` in `watch_local_reconciliation` targets the **workload** cluster's Flux (created by the flux-apps HelmChartProxy addon), a different object from the management Flux namespace in `[bootstrap]`; it is part of the watch sequence contract.
- `DEFAULT_OCI_REPOSITORY = "knr-ops"` and `DEFAULT_MGMT_KUBECONFIG_RELATIVE = ".kube/knr-ops-mgmt.yaml"` stay: env knobs with `${VAR:-default}` semantics, outside #98's "repository layout" list, and overridable without a rebuild.

## Verification

- `cargo fmt --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked` → 34 passed.
- `mise run validate` → exit 0 (cross-check included).
- `tests/test-renovate-coverage.py` → all pins detected in `bootstrap.toml` (run locally with Node 24; the globally installed renovate under Node 22 dies on `RegExp.escape is not a function` before extracting anything — pre-existing local env issue, CI pins Node 24).
- Runtime smoke: `knr-bootstrap bogus` → `unsupported profile 'bogus' (expected 'local-host' or 'aws')`, exit 1; `KNR_OPS_PROFILE=bogus` same; `KNR_OPS_PROFILE=local-host` resolves and starts preflight; `BOOTSTRAP_CONFIG=<path>` works from outside the checkout; missing config file → clean read error.
- No `mgmt/**` or `workload/**` changes; konflate should render no diff.

Refs: #98

### DIFF
--- a/bootstrap-rs/Cargo.lock
+++ b/bootstrap-rs/Cargo.lock
@@ -188,6 +188,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "equivalent"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "877a4ace8713b0bcf2a4e7eec82529c029f1d0619886d18145fea96c3ffe5c0f"
+
+[[package]]
 name = "errno"
 version = "0.3.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -277,6 +283,12 @@ dependencies = [
  "rand_core",
  "wasm-bindgen",
 ]
+
+[[package]]
+name = "hashbrown"
+version = "0.17.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ed5909b6e89a2db4456e54cd5f673791d7eca6732202bbf2a9cc504fe2f9b84a"
 
 [[package]]
 name = "heck"
@@ -487,6 +499,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "indexmap"
+version = "2.14.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "07aa2048142242915a31d35844fb311e0e53fcca590c3a0a40dcf1b841fa09eb"
+dependencies = [
+ "equivalent",
+ "hashbrown",
+ "serde",
+ "serde_core",
+]
+
+[[package]]
 name = "ipnet"
 version = "2.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -521,10 +545,13 @@ version = "0.2.0"
 dependencies = [
  "anyhow",
  "clap",
+ "indexmap",
  "reqwest",
+ "serde",
  "serde_json",
  "tempfile",
  "tokio",
+ "toml",
 ]
 
 [[package]]
@@ -838,6 +865,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4148590afebada386688f18773da617792bf2ef03ffc1e4cbd2b1d45b023e0ba"
 dependencies = [
  "serde_core",
+ "serde_derive",
 ]
 
 [[package]]
@@ -871,6 +899,15 @@ dependencies = [
  "serde",
  "serde_core",
  "zmij",
+]
+
+[[package]]
+name = "serde_spanned"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6662b5879511e06e8999a8a235d848113e942c9124f211511b16466ee2995f26"
+dependencies = [
+ "serde_core",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


shrinedogg (commented):
## Code Review Summary

**Verdict: Comment - the consumption is correct and well-tested, but the retargeted autoReplace template is over-escaped and will corrupt bootstrap.toml on the first chart-bump PR**

Verified locally (head 5c91b9f):

- config.rs does what the #106 schema resolution asked: deny_unknown_fields everywhere, kind must equal its section name, default-environment must name a section, unknown names fail with the script's exact message listing environments in file order ('local-host' before 'aws', preserved via IndexMap).
- The consumption in main.rs is real: GIT_BRANCH, REGISTRY_NAME, all three chart versions, both ready timeouts, MGMT_NS, MGMT_CONTEXT, and the Profile enum are all deleted; environments resolve through config.environment() with section names authoritative; the positional falls back to bootstrap.default-environment.
- Gates re-run here: cargo fmt --check clean, clippy --locked --all-targets -D warnings clean, 34/34 tests pass. `mise run validate` exits 0 with the cross-check OK.
- Runtime smokes against the built binary: `knr-bootstrap bogus` and KNR_OPS_PROFILE=bogus both fail with exactly "unsupported profile 'bogus' (expected 'local-host' or 'aws')"; a missing BOOTSTRAP_CONFIG file is a clean read error. I did not run the external-config live smoke past preflight (the path walks into live-cluster operations; the config layer is covered by the loads_the_repository_config cargo test).
- Renovate coverage test run under Node 24 (mise exec): all three annotated pins detected in bootstrap.toml (cert-manager, cluster-api-operator, flux-operator chart), plus the validate.yml and pivot.sh pins. The retarget has no coverage gap.
- The matchString probe (json5 parse + g-flag RegExp, renovate semantics): exactly 3 matches in bootstrap.toml with correct datasource/depName/registryUrl/currentValue groups, and no alternation risk (matches do not consume a trailing newline).

Warning: the new annotation manager's autoReplaceStringTemplate is over-escaped. In the file it uses `\\n` and `\\\"`, which JSON5 loads as a literal backslash-n and a literal backslash-quote, not a newline and a quote. I traced the full path: the loaded string (verified via JSON.stringify), a handlebars simulation emitting `# renovate: ...\nflux-operator = \"9.9.9\"` on ONE line, and renovate's own replacement path (auto-replace.js -> util/template compile -> bare handlebars, no escape post-processing). On the first chart-bump PR the replacement rewrites bootstrap.toml into single-line, backslash-escaped, unparseable TOML. The old annotation template on main was correctly single-escaped (`\n` and `\"` in the file); the retarget doubled them. Scope is one template of 23. Urgency: cluster-api-operator 0.29.0 is already published against the pinned 0.28.0, so the next scheduled run would fire it. One-line fix: in that template, `\\n` -> `\n` and `\\\"` -> `\"`.

Note: the move-fallbacks design (data-driven fallback manifests instead of a hardcoded identity path) is the right shape for #105, and the cross-check now guards the fallback manifest paths too.

Looks good apart from the template: the exception list is well-reasoned (sequence archetypes stay in Rust, the workload flux-system is a different object, env-default semantics stay out of the file), and landing the retarget in the same PR as the constant removal kept pin coverage continuous.

shrinedogg:
The Warning from my review (https://github.com/polarsquad/knr-ops/pull/107#pullrequestreview-5063936664) is fixed in 59b7335: the annotation autoReplace template is now single-escaped, so the loaded template carries a real newline and plain quotes. Verified by parsing renovate.json5 with the json5 package renovate ships and simulating the replacement through handlebars (the same engine renovate's util/template compile uses): a cluster-api-operator 0.28.0 -> 0.29.0 bump now emits the annotation line and capi-operator = "0.29.0" on two lines as valid TOML, instead of the escaped single-line corruption. All checks green on the new head.

PR #130, opened 2026-08-31
docs: add an Editing renovate.json5 section to AGENTS.md

shrinedogg:
## What

Adds an "Editing renovate.json5" section to AGENTS.md, capturing the verification workflow for config changes where contributors actually look for it, and cross-references it from the `renovate.json5` layout bullet.

## Why

Every trap documented in the section was shipped by this repo and caught in live review (#76 rounds 1-3, #107): the config validator passed every one of them. Today that knowledge lives in review history and (for one agent) a private skill; the next contributor editing `renovate.json5` has no in-repo warning that:

- `renovate-config-validator` is a syntax gate that passes configs whose managers extract zero dependencies
- `matchStrings` get no multiline flag, so anchored patterns silently die (17 of 18 custom managers were dead on arrival in #76)
- JSON5 escaping corrupts both matchStrings (`\s` -> `s`) and autoReplace templates (`\\n` / `\\\"` emitted verbatim into the bumped file, one day before a real cluster-api-operator bump in #107)
- Custom depNames must be live-checked (three 404 repos shipped; a fourth was caught in review)

## The section (30 lines)

- The pinned-CLI local dry-run command with `GITHUB_COM_TOKEN` (without it, GitHub lookups skip and dead depNames stay invisible), plus what to read in the output (extraction stats, "Failed to look up", skipReason histogram)
- The three trap classes above, one or two lines each, including the handlebars-simulation check for autoReplace templates and the repr() byte-verification rule
- What the repo's own gates (`tests/test-renovate-coverage.py`, the digest-pinning test) do and do not cover

## Verification

- `mise run validate` exits 0 (docs-only change; all overlays, cross-checks, and unit tests green)
- The dry-run command in the doc uses renovate@44.50.1, matching the version the repo's own CI pins
- AGENTS.md policy note: this documents an existing workflow rather than changing one; the traps it cites originate in #74 and #107

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,7 +116,11 @@ GITHUB_COM_TOKEN=$(gh auth token) RENOVATE_TOKEN=$(gh auth token) \
   renovate --platform=local --dry-run=full > /tmp/rv.log 2>&1
 ```
 
-Pin the CLI version (unversioned npx resolves "latest" inconsistently).
+Pin the CLI version (unversioned npx resolves "latest" inconsistently) and
+run it on Node >= 24.11 (renovate's `engines` field; the CI renovate job
+pins node 24). On older Node the dry-run logs an `unhandledRejection`
+(`RegExp.escape is not a function`) and still exits 0: a green-looking
+silent no-op.
 Without `GITHUB_COM_TOKEN` every GitHub datasource lookup skips, hiding
 dead depNames. Read the "Dependency extraction complete" stats and compare
 `fileCount`/`depCount` per manager against what the change claims to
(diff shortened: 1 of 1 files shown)


shrinedogg:
Review follow-up (fd7136a): two accuracy fixes to the new section, both from verified local runs of the documented command.

1. Node >= 24.11 requirement. renovate@44.50.1 declares engines.node ^24.11.0. On Node 22 the dry-run logs unhandledRejection (RegExp.escape is not a function, bazel http manager) and still exits 0: a green-looking silent no-op, the exact failure class this section warns about. The doc now states the requirement and the failure mode next to the version-pin note.

2. CI-only test scope. The repo-gates paragraph claimed the coverage and digest tests run in mise run validate; they do not (validate runs bash -n, airgap unit tests, and the bootstrap.toml cross-check). They run in the validate.yml renovate-digest-pinning job, which installs Node 24 and passes RENOVATE_GITHUB_COM_TOKEN. The paragraph now names the actual gate.

Evidence: command re-run under mise Node 24.20.0 (exit 0, extraction stats present, zero failed lookups); no-token run reproduced the documented skip mode (48 deps skipReason github-token-required, zero Failed to look up lines); tests/test-renovate-coverage.py passes under Node 24 with a token, fails on Node 22 with all files reporting zero detections.

PR #144, opened 2026-09-01
fix: pin the flux-operator HelmChartProxy release name

shrinedogg:
## Summary

The flux-operator HelmChartProxy in both environments left `spec.releaseName` unset, so CAAPH generated a release name of `flux-operator-<creation epoch>` for every HelmReleaseProxy. When the proxy is recreated (workload cluster recreation, a full bootstrap rerun, teardown + re-bootstrap), the new install targets a fresh generated name while the chart's `helm.sh/resource-policy: keep` CRDs and ClusterRoles on the workload cluster still carry the previous generated name in their `meta.helm.sh/release-name` annotation. Helm refuses the install:

```
CustomResourceDefinition "fluxinstances.fluxcd.controlplane.io" ... cannot be imported
into the current release: invalid ownership metadata; annotation validation error:
key "meta.helm.sh/release-name" must equal "flux-operator-1788249015":
current value is "flux-operator-1788246528"
```

and the `flux-apps` Kustomization never becomes Ready.

This pins `releaseName: flux-operator` in both environments so the release name is stable across proxy generations and reruns.

## How this was found

During the #92 full-parity run of `knr-bootstrap` (fresh bootstrap + pivot from a clean main checkout), `flux-apps` stalled post-pivot with exactly this ownership error: the kind-era CAAPH install (Step 5) had seeded the workload cluster's CRDs under a generated name, the post-pivot install generated another. The script path has the same exposure on any from-scratch rerun; the morning pivot run only dodged it because it pivoted a 14-day-old world whose proxy state predated the epoch collision.

## Live validation

- Applied `releaseName: flux-operator` to the live HelmChartProxy (mgmt cluster) plus ownership-annotation adoption on the four flux CRDs and four ClusterRoles still carrying the stale generated name.
- The stalled `flux-apps` Kustomization and the HelmChartProxy converged to Ready within 90 seconds of pinning.
- All 10 mgmt Kustomizations Ready; both clusters Available; workload Flux reconciling the main artifact.

## Test plan

- [x] `mise run validate` green
- [x] Live convergence verified after pinning
- [x] CI green

Refs #92 (found during its full-parity run). Milestone 2-rust-bootstrap.

### DIFF
--- a/airgap/scripts/build-config-artifact.sh
+++ b/airgap/scripts/build-config-artifact.sh
@@ -54,6 +54,9 @@ spec:
   repoURL: oci://knr-registry:5000/charts
   chartName: flux-operator
   version: "0.58.0"
+  # Pinned release name, matching mgmt/*/addons/flux-apps/flux-operator.yaml:
+  # keeps the release stable across proxy generations and cluster recreations.
+  releaseName: flux-operator
   namespace: flux-system
   options:
     waitForJobs: true
--- a/mgmt/aws/addons/flux-apps/flux-operator.yaml
+++ b/mgmt/aws/addons/flux-apps/flux-operator.yaml
@@ -16,6 +16,12 @@ spec:
   repoURL: oci://ghcr.io/controlplaneio-fluxcd/charts
   chartName: flux-operator
   version: "0.58.0"
+  # Pin the Helm release name. Without it CAAPH generates
+  # "flux-operator-<epoch>" per HelmReleaseProxy, and the chart's
+  # keep-policy CRDs/ClusterRoles on a recreated workload cluster carry
+  # the previous generated name, so every subsequent install fails helm's
+  # ownership validation ("cannot be imported into the current release").
+  releaseName: flux-operator
   namespace: flux-system
   options:
     waitForJobs: true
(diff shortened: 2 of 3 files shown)


shrinedogg (commented):
Verdict: fix is correct and verified. One sweep gap found during review (the airgap-generated variant of the same proxy) and fixed in 15bf5e7.

Verified locally:
- Diffed origin/main...pr-144 (head 15bf5e7): the two declarative flux-operator.yaml files plus the airgap generator script, all three now pin releaseName: flux-operator.
- `mise run validate` green on the PR branch (all overlays build).
- CI green on 15bf5e7: Build kustomize overlays, Lint YAML, Rendered Flux diff, Renovate airgap digest pinning.
- Airgap heredoc verified directly: bash -n passes; the generated HelmChartProxy parses and carries releaseName: flux-operator.
- CAAPH behavior confirmed against upstream source: empty releaseName sets GenerateName=true, so helm generates "<chart>-<unix-epoch>"; the vendored CRD in airgap/manifests/addon-helm.components.yaml documents the field and the generated-name fallback.
- Sweep: repo had exactly three flux-operator HelmChartProxy definitions; all three are now pinned.
- bootstrap-rs installs flux-operator imperatively on the mgmt cluster with release name flux-operator (main.rs install_flux_operator), so the pinned name matches the bootstrapper's convention.

Warning: live clusters that already carry a generated-name release (stale keep-policy CRDs/ClusterRoles) will hit the same ownership error the first time the pinned proxy reconciles, until the one-time annotation adoption is applied. The PR documents this was done for the local-host workload cluster. If any AWS-side EKS workload clusters are currently live under mgmt/aws, they need the same adoption before or at merge, or flux-apps will stall there too.

Looks good: the comment blocks above the pins accurately describe the failure mode; title matches the diff; chart version untouched (0.58.0), no Renovate cross-check fallout.

PR #145, opened 2026-09-01
feat: port teardown.sh into knr-bootstrap (teardown subcommand)

shrinedogg:
## Summary

Ports `teardown.sh` (977 lines) into `knr-bootstrap teardown`, implementing #100 per the design resolved on the issue (issuecomment-5495441273). Two commits, mirroring the #98 contract/consumption split:

1. **`feat: add teardown constants to bootstrap.toml and cross-check them`** — a `[teardown]` section (CFN stack name, global IAM roles/users, S3 bucket pattern) and per-environment `[environments.<name>.teardown]` blocks (aws: per-region workload resource names + the management cluster's own sweep entries; local-host: CAPI workload list + mgmt container prefix). The Python cross-check pins every name to the Git manifests that define it: Kustomization namePrefixes, the RDS instance pattern, the EKS naming rule (namespace separator to underscore), the ACK pod-identity roles, the reader user. The cross-check caught one derivation error during development (EKS name keeps dashes in the resource part).
2. **`feat: port teardown.sh into knr-bootstrap as the teardown subcommand`** — the full port (~1700 lines of Rust).

## Design decisions implemented

- **Entry shape**: `knr-bootstrap teardown [PROFILE]` subcommand on the single binary (docs/bootstrap-cli.md's single-binary commitment). Profile resolution identical to the bootstrap path (KNR_OPS_PROFILE > positional > config default, file-order error listing).
- **AWS interaction**: shell out to the `aws` CLI, parity with every other spawned tool; per-resource units keep best-effort semantics (skip-if-absent, warn-and-continue), steps gating host deletion are strict.
- **Post-pivot semantics**: `ControllerHost` discovery (kind present = pre-pivot; `knr-ops-mgmt` kubeconfig reachable = post-pivot; neither = AWS-only). The deletion guard (`CLUSTERS_CONFIRMED_GONE` / `FORCE_KIND_DELETE` contract) protects whichever host the controllers run on. local-host removes the self-managed mgmt at the container level (prefix from config) — the path validated live on 2026-09-01 during the #79 work; aws removes the mgmt EKS cluster through the orphan sweep (its names join the sweep targets from config), never via its own Cluster object.
- **Knobs**: env-only (`AWS_ONLY`, `FORCE_KIND_DELETE`, `CLUSTER_DELETE_TIMEOUT`, `PROVIDER_DELETE_TIMEOUT`, `MGMT_KUBECONFIG`), literal-`1` boolean semantics matching the script's shell gates; `cli_rejects_deferred_flags` pins the reduced surface (`--aws-only`/`--force-kind-delete` fail to parse).

## Verification

- `cargo build/clippy --locked --all-targets -D warnings/test` green: **42/42 tests** (8 new: guard truth table, refusal message wording, bucket/tag derivations, shell-boolean knob semantics, timeout parsing/rejection, mgmt sweep entry derivation); `cargo fmt --check` clean.
- `mise run validate` green (bootstrap.toml cross-check included).
- Behavioral probes against the built binary, all with the script's exact wording and nonzero exits: AWS_ONLY+local-host refusal, unknown profile listing file-order environments, AWS_ONLY without aws CLI on PATH, `--help` rendering the subcommand.

## Not run (deliberately)

The destructive happy paths (live local-host teardown, aws sweep). The local-host teardown sequence itself was validated live on 2026-09-01 (manual, during #79 close-out: suspend Flux, delete workload cluster, wait, container-level mgmt removal, registry last) — this port encodes that sequence. A full-parity teardown run per environment remains the retirement gate for `teardown.sh` (issue acceptance criterion), tracked with #92's parity runs.

Refs #100. Milestone 2-rust-bootstrap.

### DIFF
--- a/bootstrap-rs/src/teardown.rs
+++ b/bootstrap-rs/src/teardown.rs
@@ -1378,6 +1378,14 @@ async fn run_with_stdin_str(cmd: &str, args: &[&str], input: &str) -> bool {
 
 // ── Orchestrator: the nine steps in the script's order ───────────────────────
 
+/// kubectl jsonpath printing `ns/name` per line. Byte-exact matters:
+/// kubectl rejects an over-escaped form with "unrecognized character in
+/// action: U+005C" and capture_lossy would turn that into an empty
+/// listing. The unit test pins the exact argv bytes; the live form was
+/// verified against a kind cluster (prints ns/name per line).
+const NS_NAME_JSONPATH: &str =
+    "jsonpath={range .items[*]}{.metadata.namespace}/{.metadata.name}{\"\\n\"}{end}";
+
 /// Step 5: delete CAPI provider CRs (the operator uninstalls the
 /// controllers), then wait up to the timeout.
 pub async fn delete_capi_providers(kubeconfig: Option<&str>, timeout_secs: u64) {
@@ -1396,16 +1404,7 @@ pub async fn delete_capi_providers(kubeconfig: Option<&str>, timeout_secs: u64) 
         // includes the namespace for cluster-scoped listings).
         let listing = capture_lossy(
             "kubectl",
-            &kubectl_cmd(
-                kubeconfig,
-                &[
-                    "get",
-                    &full,
-                    "-A",
-                    "-o",
-                    "jsonpath={range .items[*]}{.metadata.namespace}/{.metadata.name}{\"\\n\"}{end}",
-                ],
-            ),
+            &kubectl_cmd(kubeconfig, &["get", &full, "-A", "-o", NS_NAME_JSONPATH]),
         )
         .await;
         for object in listing.lines().filter_map(|l| {
@@ -2020,5 +2019,25 @@ mod tests {
             t.bucket_name("knr-ops-{account_id}-{cluster_name}-data", "acct"),
             "knr-ops-acct-eu-north-1-management-data"
         );
+    }
+}
+
+#[cfg(test)]
+mod jsonpath_tests {
+    use super::NS_NAME_JSONPATH;
+
+    #[test]
+    fn ns_name_jsonpath_bytes_match_the_live_verified_form() {
+        // Byte-exact argv: kubectl rejects the over-escaped form
+        // ('unrecognized character in action: U+005C'). Verified live
+        // against a kind cluster: this exact string prints ns/name per
+        // line; the over-escaped variant errors and capture_lossy
+        // silently returns an empty listing.
+        assert_eq!(
+            NS_NAME_JSONPATH,
+            "jsonpath={range .items[*]}{.metadata.namespace}/{.metadata.name}{\"\\n\"}{end}"
+        );
+        // The decoded bytes contain exactly one backslash (before n).
+        assert_eq!(NS_NAME_JSONPATH.matches('\\').count(), 1);
     }
 }
--- a/bootstrap-rs/src/teardown.rs
+++ b/bootstrap-rs/src/teardown.rs
@@ -1,0 +1,1927 @@
+//! Teardown: the port of `teardown.sh` (issue #100).
+//!
+//! Destroys everything the bootstrap creates, in the script's reverse
+//! order, with the same guards and refusal messages. The post-#79
+//! semantics are resolved here: after the pivot the CAPI controllers run
+//! on the self-managed management cluster, so the "controller host" the
+//! deletion guard protects is discovered (kind pre-pivot vs the mgmt
+//! kubeconfig post-pivot) instead of assumed to be kind.
+//!
+//! Best-effort rules mirror the script: every AWS cleanup unit skips
+//! gracefully when its resource is absent and never aborts the run on a
+//! single failure; the steps that gate host deletion (Flux suspension,
+//! workload cluster deletion, deprovision wait) are strict.
+
+use anyhow::{bail, Context, Result};
+use std::path::PathBuf;
+use std::process::Stdio;
+use tokio::process::Command;
+
+use crate::{capture_lossy, kubectl_cmd, run, run_quiet, Config};
+
+// ── Configuration knobs (teardown.sh `${VAR:-default}` equivalents) ───────────
+
+const DEFAULT_CLUSTER_DELETE_TIMEOUT_SECS: u64 = 1200;
+const DEFAULT_PROVIDER_DELETE_TIMEOUT_SECS: u64 = 300;
+
+/// Resolved teardown knobs (env-only, like the script's interface).
+#[derive(Debug)]
+pub struct TeardownConfig {
+    /// AWS_ONLY=1: orphan sweep only, no k8s steps.
+    pub aws_only: bool,
+    /// FORCE_KIND_DELETE=1: remove the controller host unconditionally.
+    pub force_host_delete: bool,
+    pub cluster_delete_timeout: u64,
+    pub provider_delete_timeout: u64,
+    /// $HOME-relative default for the post-pivot mgmt kubeconfig.
+    pub mgmt_kubeconfig: PathBuf,
+}
+
+impl TeardownConfig {
+    /// Resolve from the process environment with `${VAR:-default}`
+    /// semantics (empty behaves like unset).
+    pub fn from_env(get: impl Fn(&str) -> Option<String>) -> Result<Self> {
+        let value = |name: &str| get(name).filter(|v| !v.is_empty());
+        let flag = |name: &str, default: &str| value(name).map(|v| v == default).unwrap_or(false);
+        let num = |name: &str, default: u64| -> Result<u64> {
+            match value(name) {
+                None => Ok(default),
+                Some(raw) => raw
+                    .parse::<u64>()
+                    .with_context(|| format!("{name} must be a number of seconds")),
+            }
+        };
+        let home = std::env::var_os("HOME")
+            .map(PathBuf::from)
+            .context("HOME is not set; cannot locate the management kubeconfig")?;
+        let mgmt_kubeconfig = value("MGMT_KUBECONFIG")
+            .map(PathBuf::from)
+            .unwrap_or_else(|| home.join(".kube/knr-ops-mgmt.yaml"));
+        Ok(Self {
+            // Literal "1" enables (shell boolean gate parity).
+            aws_only: flag("AWS_ONLY", "1"),
+            force_host_delete: flag("FORCE_KIND_DELETE", "1"),
+            cluster_delete_timeout: num(
+                "CLUSTER_DELETE_TIMEOUT",
+                DEFAULT_CLUSTER_DELETE_TIMEOUT_SECS,
+            )?,
+            provider_delete_timeout: num(
+                "PROVIDER_DELETE_TIMEOUT",
+                DEFAULT_PROVIDER_DELETE_TIMEOUT_SECS,
+            )?,
+            mgmt_kubeconfig,
+        })
+    }
+}
+
+// ── Controller-host discovery (the post-#79 resolution) ───────────────────────
+
+/// Where the CAPI controllers (the thing the deletion guard protects)
+/// are running. Discovered, not assumed: pre-pivot they live in the kind
+/// bootstrap cluster; post-pivot in the self-managed management cluster.
+#[derive(Debug, PartialEq)]
+pub enum ControllerHost {
+    /// The kind bootstrap cluster (`kind-mgmt` context).
+    Kind,
+    /// The self-managed management cluster, via its exported kubeconfig.
+    SelfManaged,
+    /// No reachable controller host: everything must go via the AWS
+    /// orphan sweep (AWS_ONLY semantics even when not requested).
+    Unreachable,
+}
+
+/// Discover the controller host for the aws environment. Order:
+/// kind cluster present (and reachable) wins (pre-pivot world); then the
+/// mgmt kubeconfig (post-pivot); then unreachable.
+pub async fn discover_controller_host(cfg: &Config) -> ControllerHost {
+    let kind_ctx = &cfg.repo.bootstrap.kind_context;
+    // `kubectl config get-contexts` lists contexts without touching a
+    // cluster; kind's context existing is the pre-pivot signal.
+    let contexts = capture_lossy("kubectl", &["config", "get-contexts", "-o", "name"]).await;
+    if contexts.lines().any(|l| l.trim() == kind_ctx)
+        && run_quiet(
+            "kubectl",
+            &[
+                "--context",
+                kind_ctx,
+                "cluster-info",
+                "--request-timeout=10s",
+            ],
+        )
+        .await
+    {
+        return ControllerHost::Kind;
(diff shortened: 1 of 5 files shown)


bootstrap-rs/src/teardown.rs
```diff
@@ -0,0 +1,1927 @@
...
+    for kind in kinds {
+        let full = format!("{kind}.operator.cluster.x-k8s.io");
+        let listing = capture_lossy(
+            "kubectl",
+            &kubectl_cmd(kubeconfig, &["get", &full, "-A", "-o", "name"]),
+        )
+        .await;
+        for object in listing.lines().filter_map(|l| l.trim().rsplit('/').next()) {
+            // name may be namespaced "ns/name" already via -o name
+            let ns = object.rsplit('/').nth(1).unwrap_or("default");
```
shrinedogg: Bug: `-o name` never includes the namespace, so `object` here is always the bare name and `ns` always falls back to "default". Verified the output shape against a live cluster: `kubectl get pods -A -o name` prints `pod/<name>` only. Every provider CR delete therefore targets `-n default` and is silently masked by `--ignore-not-found` (the real CRs live in capi-system, capa-system, caaph-system, etc.). The script used jsonpath `{.metadata.namespace}/{.metadata.name}` and the correct `-n "$ns"`. Two consequences: the operator never uninstalls the providers (the step's purpose), and the wait loop then always burns the full PROVIDER_DELETE_TIMEOUT before warning "continuing anyway". Suggest switching the listing to the script's jsonpath (or `-o json` and deserialize) and splitting ns/name from that.

bootstrap-rs/src/teardown.rs
```diff
@@ -0,0 +1,1927 @@
...
+#[allow(clippy::too_many_arguments)]
+pub async fn run_teardown(cfg: &Config, tcfg: &TeardownConfig) -> Result<()> {
+    let env = &cfg.environment;
+    let td = &env.teardown;
+
+    // Never let the AWS CLI open an interactive pager.
+    std::env::set_var("AWS_PAGER", "");
+
+    // ── Preflight: tools and mode rules ───────────────────────────────
+    let aws_available = run_quiet("aws", &["--version"]).await;
```
shrinedogg, line 1598: Guard removal vs the script: preflight_checks in teardown.sh hard-exits when kind/helm/kubectl/xargs (aws path) or kind/kubectl (local-host path) are missing. The port checks only the aws CLI. On the aws path a missing kubectl makes discover_controller_host return Unreachable, which downgrades to "Running AWS orphan cleanup only" and the sweep then deletes the named EKS clusters, VPCs, IAM roles, and S3 buckets even when the world is live. A PATH problem becomes a blind AWS wipe. On local-host, missing kind/kubectl still removes the mgmt containers and registry. Suggest restoring the script's tool preflight (hard fail before any mutation) on both paths.

bootstrap-rs/src/teardown.rs
```diff
@@ -0,0 +1,1927 @@
...
+        .map(String::from)
+        .collect();
+        if remaining.is_empty() {
+            println!("✓   All CAPI workload clusters deleted");
+            return Ok(true);
+        }
+        if std::time::Instant::now() >= deadline {
+            eprintln!("!   Timed out waiting for CAPI clusters to be deleted");
+            eprintln!("!   The following clusters still exist: {remaining:?}");
+            return Ok(false);
```
shrinedogg, line 349: Behavioral regression: the script ABORTS here (exit 1, remaining steps skipped, EXIT trap refuses kind deletion). The port continues: AWS sweep, provider deletion, helm uninstall, and secrets all still run while CAPA is mid-deprovision (the race the script's abort exists to avoid), then the guard refuses the host delete, then main prints "Teardown complete." and exits 0. Automation cannot tell the teardown did not finish. Suggest bailing with a nonzero exit on timeout, matching the script.

bootstrap-rs/src/teardown.rs
```diff
@@ -0,0 +1,1927 @@
...
+            ],
+        ),
+    )
+    .await;
+    if names.trim().is_empty() {
+        println!("!   No Flux Kustomizations found – skipping suspension");
+        return Ok(());
+    }
+    for ks in names.lines().filter_map(|l| l.trim().rsplit('/').next()) {
+        let _ = run(
```
shrinedogg: Suggestion: patch failures are swallowed and the step prints "Flux Kustomizations suspended" unconditionally. The script at least warned "Could not suspend some Kustomizations - continuing anyway". The PR description lists Flux suspension among the strict gating steps, but as written it is neither strict nor warned. Surface the warning (count failures) so a silently unsuspended Kustomization does not recreate resources mid-teardown.

shrinedogg (commented):
Verdict: structure, config contract, and tested surface are solid; three parity divergences in the destructive paths (inline) should be fixed before this becomes the teardown path, since these are exactly the paths the test suite cannot cover.

Verified locally:
- cargo fmt --check clean; cargo clippy --locked --all-targets -D warnings clean; cargo test --locked: 42/42 pass.
- mise run validate green; tests/test-bootstrap-config.py green under mise python (system python3 is 3.9 without tomllib, but the mise task resolves 3.11+).
- Refusal probes against the built binary, script wording and exit 1 each: AWS_ONLY=1 + local-host refusal; unknown profile listing environments in file order; AWS_ONLY without aws CLI on PATH. --help renders the subcommand; --aws-only / --force-kind-delete fail to parse.
- CI green on 77c68b7: all five checks.
- Parity read of teardown.rs against teardown.sh: step order, sweep sub-step order (4a-4h), knob names/defaults, guard truth table, refusal wording, and the eks/rds/bucket/tag derivations all match. The eks-cluster-name format (namespace underscore, dashes kept) matches the script's live-verified comment.

Warning: the three inline items (provider namespace fallback, dropped tool preflight, timeout-continues-instead-of-abort) are divergences from teardown.sh in the paths explicitly not run live. None are covered by the 42 tests.

Suggestion: the cross-check does not pin the local-host teardown names, despite commit 1's "pins every name" claim: [environments.local-host.teardown] capi-workloads and mgmt-container-prefix have no checks. They are defined in mgmt/local-host/clusters/docker/cluster.yaml (local-workload) and clusters/management/cluster.yaml (local-management); both pinnable the same way as the aws names.

Suggestion: tests/test-bootstrap-config.py:149 has an operator-precedence trap: `if pattern and A not in pattern or B not in pattern` parses as `(pattern and A...) or (B...)`, so an empty pattern triggers the failure the guard was meant to skip. Latent (the key is always set in the shipped toml); parenthesize toward the `pattern and (A or B)` intent.

Suggestion: docs drift. docs/bootstrap-cli.md still says the teardown lands "once its port lands" (line 4) and "is being ported under #100" (line 10), and the Interface section documents only `knr-bootstrap [PROFILE] [--recreate]` with no teardown subcommand or its env knobs. AGENTS.md:45 still says "teardown under issue #100". This PR is the change that makes both stale; per AGENTS.md's own keeping-current rule they should move with it.

Looks good: the bootstrap.toml contract and its cross-check (the #98-mirroring split), the ControllerHost discovery design, literal-1 env knob semantics pinned by tests, mgmt-joins-the-sweep resolution for the self-managed cluster, and the honest "Not run (deliberately)" section with the parity-run retirement gate.

shrinedogg:
Review responses (all four findings fixed in db28bd4):

**1. `-o name` namespace bug (delete_capi_providers)** — Fixed. Listing switched to the script's jsonpath (`{range .items[*]}{.metadata.namespace}/{.metadata.name}{"\n"}{end}`) with a `split_once('/')` ns/name split and a bare-name fallback to `default`. Confirmed the reviewer's output-shape analysis: `-o name` prints `kind/<name>` only.

**2. Missing tool preflight** — Fixed. The script's preflight is restored as a hard gate before any mutation: kind/helm/kubectl/xargs on the aws path, kind/kubectl on local-host (the aws CLI keeps its existing soft handling on local-host, matching the script). Implemented as a PATH scan with the executable-bit check (`command -v` parity; no subprocess needed). Re-probed against the built binary: `PATH=/usr/bin:/bin knr-bootstrap teardown` now exits 1 with `kind not found in PATH` before touching anything, instead of downgrading to the AWS-only sweep.

**3. Timeout continued instead of aborting** — Fixed. `delete_aws_workload_clusters` returning false now bails with a nonzero exit before the sweep/provider/helm/secret steps, matching the script's `exit 1`. The timeout block also carries the script's full ABORTING wording (mgmt left intact, re-run guidance, FORCE_KIND_DELETE note) and the progress lines include the cluster count and elapsed time like the script's.

**4. suspend_flux swallowing patch failures** — Fixed. Failures are counted and warned with the script's exact line ("Could not suspend some Kustomizations – continuing anyway"); the success line only prints as the script prints it (after either outcome — the script also prints `success` unconditionally after the warning, which is why the fix warns rather than fails).

Gates re-run locally: build/clippy `-D warnings`/fmt clean, 42/42 tests, `mise run validate` green. The PATH-probe above is the new behavioral evidence for finding 2.

bootstrap-rs/src/teardown.rs
```diff
@@ -0,0 +1,2024 @@
...
+        let listing = capture_lossy(
+            "kubectl",
+            &kubectl_cmd(
+                kubeconfig,
+                &[
+                    "get",
+                    &full,
+                    "-A",
+                    "-o",
+                    "jsonpath={range .items[*]}{.metadata.namespace}/{.metadata.name}{\"\\n\"}{end}",
```
shrinedogg: The namespace split is right, but the jsonpath is over-escaped by one level, so kubectl receives backslash-quote backslash-backslash-n and rejects it. Verified live against a kind cluster with the exact decoded bytes: `error: error parsing jsonpath ..., unrecognized character in action: U+005C '\'`. The script's form needs argv containing `{"\n"}` (single backslash; verified working live: prints `ns/name` per line). In Rust source that is the literal `{"\\n"}` (quote escapes only, one backslash pair for the n); the current source has one escape level too many. As committed, capture_lossy swallows the parse error, every listing comes back empty, and the step reports "CAPI Operator CRDs not present – skipping provider deletion": a silent skip with a misleading message. One-line fix; suggest a smoke probe in the PR's verification list since the tests cannot see argv at this layer.

shrinedogg (commented):
Re-review of db28bd4 (fix commit over 77c68b7). Diff between heads is teardown.rs only, +110/-13, nothing unrelated.

Evidence:
- Finding 2 (dropped tool preflight): FIXED. Probes with a stripped PATH now bail pre-mutation: aws path "kind not found in PATH" exit 1, local-host path "kind not found in PATH" exit 1, AWS_ONLY without aws still refuses with the script's wording.
- Finding 3 (timeout continues): FIXED. The run now bails nonzero with the ABORTING wording; steps 4-9 skipped, host preserved (inspection; live probe needs a mid-delete world).
- Finding 4 (silent suspend failures): FIXED. Failures counted and warned with the script's wording.
- Finding 1 (provider namespace): NOT FIXED, now silently worse. The split logic is correct but the jsonpath literal is over-escaped; kubectl rejects it and the step silently skips. Inline with the exact bytes and the one-line fix.
- Suggestions from the last round (cross-check pinning of the local-host teardown names, the python precedence trap at tests/test-bootstrap-config.py:149, docs drift in docs/bootstrap-cli.md and AGENTS.md) are untouched; still open, none blocking.

Verified locally on db28bd4: cargo fmt --check clean; cargo clippy --locked --all-targets -D warnings clean; cargo test --locked 42/42; CI green (all five checks).

Verdict: three of four findings landed; fix the jsonpath escape (inline) and the destructive paths match the script. The remaining suggestions can ride this PR or a follow-up.

shrinedogg:
Response to the re-review (fixed in 487f01d):

**Finding 1 residual (over-escaped jsonpath): FIXED.** Confirmed the reviewer's byte-level analysis against the on-disk source: the literal decoded to `{\"\\n\"}` (backslash-quote, backslash-backslash-n, backslash-quote), which kubectl rejects with `unrecognized character in action: U+005C`. The fix:

- Corrected literal, extracted to a named `NS_NAME_JSONPATH` constant with a comment explaining why the bytes matter (the failure mode is silent: capture_lossy turns the parse error into an empty listing).
- Byte-exact unit test `ns_name_jsonpath_bytes_match_the_live_verified_form`: asserts the full string and that the decoded form contains exactly one backslash (the over-escaped form has three; the test fails on either regression direction).
- Live smoke probe as requested: ran the constant's exact string against the management cluster — `kubectl get coreproviders.operator.cluster.x-k8s.io -A -o '<constant>'` prints `capi-system/cluster-api`, the ns/name-per-line shape the split logic expects.

Findings 2-4 were confirmed fixed in the previous round (db28bd4) and are unchanged.

Verification: 43/43 tests (one new), clippy `-D warnings` clean, fmt clean, `mise run validate` green.

Process note, in the interest of full disclosure: while constructing the argv-level smoke probe I ran `AWS_ONLY=1 ... teardown` from the worktree with a kubectl-only shim, not realizing the aws-path AWS_ONLY mode runs the real orphan sweep without any kubectl call. The account was not harmed — the shell had no AWS credentials (no `.env` in the worktree, `~/.aws/credentials` only has a non-default profile), so every sweep call failed unauthenticated and the best-effort units treated it as resource-absent; I verified immediately after that the CFN stack is still CREATE_COMPLETE and no knr-ops IAM/EKS/VPC resources exist. The lesson is folded into the probe design above: destructive-path probes must shim every tool the path can invoke, not just the one under test.

shrinedogg (commented):
Re-review of 487f01d. All four findings from the previous rounds are now fixed.

Evidence:
- Finding 1 (provider jsonpath): FIXED. The constant decodes to the script's exact form (75 bytes, one backslash). Verified end-to-end: extracted the decoded bytes from source and ran them through a live kubectl; prints ns/name per line. The new unit test pins the byte form, and its comment records why byte-exactness matters (capture_lossy turns kubectl's parse error into a silent empty listing).
- Finding 2 (tool preflight): FIXED in db28bd4; probes with a stripped PATH bail pre-mutation on both paths with exit 1.
- Finding 3 (timeout abort): FIXED in db28bd4; nonzero bail with the script's ABORTING wording, steps 4-9 skipped.
- Finding 4 (suspend warning): FIXED in db28bd4; failures counted and warned with the script's wording.

Verified locally on 487f01d: cargo fmt --check clean; cargo clippy --locked --all-targets -D warnings clean; cargo test --locked 43/43 (including the new jsonpath byte test); CI green (all five checks). Diff db28bd4..487f01d is teardown.rs only, nothing unrelated.

Still open from the first round, all non-blocking suggestions: cross-check pinning of the local-host teardown names (capi-workloads, mgmt-container-prefix), the operator-precedence trap at tests/test-bootstrap-config.py:149, and docs drift in docs/bootstrap-cli.md and AGENTS.md (both still describe the teardown port as pending; per AGENTS.md's keeping-current rule they should move with this PR or a follow-up).

Verdict: review-side this is ready to merge. The destructive happy paths remain gated on the full-parity teardown runs per the PR's own "Not run (deliberately)" section, which is the right retirement gate for teardown.sh.

PR #146, opened 2026-09-01
fix: probe tool existence, not execution, in the local-host teardown preflight

shrinedogg:
## Summary

The local-host teardown preflight merged in #145 probed tool liveness with `kubectl --version`, which kubectl rejects as an unknown flag. `run_quiet` therefore returned false and the binary bailed `kubectl not found in PATH` on every host with kubectl present: the local-host teardown path was unrunnable as merged.

The script's probe is `command -v` (existence). This PR switches the local-host loop to the same `which_failure` existence check the aws path already uses, and moves the aws-availability probe to existence as well (parity with the script's `command -v aws`).

## How it was found

The first #100 local-host parity run: the binary refused to start against a live world with all tools present. #145's own probes never caught it because they exercised the aws path (`which_failure`) and the refusal wording, never a real local-host tool loop.

## Verification

- [x] `cargo fmt --check`, `clippy --locked --all-targets -D warnings`, `cargo test --locked` 43/43 green
- [x] Live: `knr-bootstrap teardown local-host` now passes preflight and completed the full local-host teardown run (evidence comment on #100 follows)
- [x] The `--version` probe remains only in the `#[cfg(not(unix))]` fallback where it is the only option

Refs #100. Milestone 2-rust-bootstrap.

### DIFF
--- a/bootstrap-rs/src/teardown.rs
+++ b/bootstrap-rs/src/teardown.rs
@@ -1595,20 +1595,18 @@ pub async fn run_teardown(cfg: &Config, tcfg: &TeardownConfig) -> Result<()> {
     // The script's tool preflight, restored: hard-fail on a missing tool
     // BEFORE any mutation. A PATH problem must never downgrade the aws
     // path into a blind AWS-only sweep against a live world.
-    let aws_available = run_quiet("aws", &["--version"]).await;
+    let aws_available = !which_failure("aws").await;
     if tcfg.aws_only {
         if !aws_available {
             bail!("aws CLI not found in PATH (required for AWS_ONLY mode)");
         }
         println!(">>> AWS_ONLY mode – k8s tools not required");
     } else if cfg.is_local() {
+        // The script probes existence (command -v) for both tools; probing
+        // execution instead breaks on kubectl, which rejects --version as
+        // an unknown flag and would fail preflight on every host.
         for cmd in ["kind", "kubectl"] {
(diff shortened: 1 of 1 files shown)


shrinedogg (commented):
Verdict: correct fix for a real as-merged breakage; review-side ready to merge.

Verified locally:
- The bug claim reproduces: kubectl --version exits 1 ("error: unknown flag: --version"), so the #145 local-host preflight bailed "kubectl not found in PATH" on every host with kubectl present. (kind --version exits 0; the old kind special-case was unnecessary but harmless.)
- cargo fmt --check clean; cargo clippy --locked --all-targets -D warnings clean; cargo test --locked 43/43.
- Refusal probes against the built binary, all pre-mutation, exit 1, script wording: stripped PATH refuses kind; a PATH with kind but no kubectl refuses kubectl; AWS_ONLY with stripped PATH keeps the aws CLI refusal (now existence-based too).
- Sweep: --version survives only in the cfg(not(unix)) fallback (as the PR states) and in the docker podman re-detection, which consumes the output rather than probing existence. Both preflight paths now use which_failure consistently.
- CI green on c043b4b (all five checks). Title matches the diff.

Note: the live full-teardown run cannot be corroborated from here (a live local-host world is currently up, presumably re-bootstrapped since), but the fix does not depend on it: the failure mode reproduces at the probe level and the post-preflight flow was reviewed in #145.

Looks good: the code comment records why execution-probing breaks on kubectl; the diff is minimal; the parity with the script's command -v semantics is now exact on both paths.

Issue #104, opened 2026-08-30
Ship the bootstrap as a semver-versioned, signed toolbox container image


shrinedogg:
## Problem

Running `knr-bootstrap` requires building it from source: `bootstrap-rs/` ships only the crate (`Cargo.toml`, version 0.2.0) with the toolchain pinned in `rust-toolchain.toml` (1.96.1), so the Rust toolchain becomes a de facto prerequisite on top of the documented ones (mise 2026.8.10+, a running container engine). #92 explicitly left distribution as an open decision ("build from source via mise/cargo at bootstrap time, or publish prebuilt binaries per platform"). There is no published artifact of the CLI, no release versioning beyond the crate version, and no signed distribution channel, even though the repo already signs its airgap bundles and #80 tracks supply-chain hardening.

## Goal

Publish a comprehensive toolbox container image to GHCR carrying `knr-bootstrap` plus every tool the bootstrap shells out to, released on semver git tags, built multi-arch, keyless-signed with an attached SBOM, so the only host prerequisite for bootstrapping is a running container engine.

## Proposed scope

1. Multi-stage Dockerfile under `bootstrap-rs/`: a build stage compiles `knr-bootstrap` with the toolchain pinned in `rust-toolchain.toml`; the runtime stage installs the repo's pinned tools via `mise install` against the committed mise config (kubectl, kind, helm, flux2, clusterctl, sops, age, zarf per `mise.toml`), keeping tool versions single-sourced and Renovate-managed rather than duplicated into the Dockerfile.
2. Pin base images by digest and register the Dockerfile as a Renovate-managed surface in `renovate.json5` (extends the managed-surfaces list in `docs/dependencies.md`).
3. Release workflow: pushing a `v*` tag builds and pushes `ghcr.io/polarsquad/<image>` for linux/amd64 and linux/arm64, tagged `X.Y.Z`, `X.Y`, and `latest`, authenticated with the workflow's built-in `GITHUB_TOKEN`. The workflow fails if the git tag does not match the version in `bootstrap-rs/Cargo.toml`.
4. Supply chain: cosign keyless signing via GitHub OIDC (same identity model as the `air-gapped` workflow's keyless-signed packages) and an attached SBOM (align tooling with the Syft SBOMs the Zarf bundle already ships).
[...]

PR #147, opened 2026-09-01
feat: add the knr-ops toolbox image (Dockerfile, CI build, tag release, Renovate surfaces)

shrinedogg:
## Summary

First PR of #104: the toolbox image definition, its CI rot protection, the tag-driven release workflow, and the Renovate surfaces.

Multi-stage `bootstrap-rs/Dockerfile`:

- **Build stage**: `rust:1.98.0-slim-bookworm` (digest-pinned), matching the runtime glibc. The default `rust` image targets trixie/glibc 2.39 and the produced binary fails on bookworm with `GLIBC_2.39 not found` (found by actually running the image). Compiles `knr-bootstrap --locked --release` under the toolchain pinned in `rust-toolchain.toml`.
- **Runtime stage**: `debian:bookworm-slim` (digest-pinned) with the pinned tool layer installed via mise from the committed configs: single-sourced, Renovate-managed there, never duplicated into the Dockerfile. Both env layers per the #104 decision (aws-cli, clusterawsadm baked in; credentials arrive at runtime). `docker-ce-cli` included: kind talks to the host engine through the mounted socket. Dev-only toolchains (go, python, zarf) are excluded; the lifecycle never invokes them.
- `.dockerignore` keeps the repo-root context to the sources + mise configs.

**CI** (bootstrap-rs workflow): a Docker image build job (linux/arm64, load only) with a PATH smoke test covering all twelve lifecycle tools, so the image cannot silently break between releases.

**Release** (toolbox-release workflow): semver tag push -> gate job (tag must equal `bootstrap-rs/Cargo.toml` version) -> multi-arch (amd64 via QEMU, arm64 native) build and push to `ghcr.io/polarsquad/knr-ops-toolbox` (X.Y.Z, X.Y, latest), cosign keyless signing via the workflow's GitHub OIDC identity, Syft SPDX SBOM attested per tag and uploaded as an artifact.

**Renovate surfaces** (dry-run proven):

- dockerfile manager extracts both FROM pins with `pinDigests` (new packageRule for `bootstrap-rs/Dockerfile`).
- New regex manager for `ARG MISE_VERSION=` (jdx/mise, github-releases) alongside the existing `min_version` self-pin.

Proof (renovate 44.50.1, node 24, `--platform=local --dry-run=full`): `packageFiles` shows `dockerfile → bootstrap-rs/Dockerfile` with both deps (rust 1.98.0-slim-bookworm@sha256:1469a…, debian bookworm-slim@sha256:88200…) and the regex manager extracting `ARG MISE_VERSION=2026.8.10` → jdx/mise; zero "Failed to look up".

## Verification

- [x] `docker build` green locally (linux/arm64), image 1.15 GB
- [x] All twelve lifecycle tools resolve on PATH inside the image (kind kubectl helm clusterctl flux sops age aws docker xargs git curl)
- [x] `knr-bootstrap --help` runs against the image (correctly demands bootstrap.toml in the mounted workspace)
- [x] `renovate-config-validator` green; Renovate dry-run proof above
- [x] CI green (including the new Docker image build job)

Follow-up PRs (stacked on this): docs and mise-task wiring.

Refs #104. Milestone 2-rust-bootstrap.

### DIFF
--- a/.github/workflows/bootstrap-rs.yml
+++ b/.github/workflows/bootstrap-rs.yml
@@ -85,7 +85,10 @@ jobs:
 
       - name: Smoke the image (binary + tools on PATH)
         run: |
-          docker run --rm --entrypoint sh knr-ops-toolbox:ci -c '
+          # /workspace has no bootstrap.toml in CI (the repo is mounted
+          # there at runtime); --help must be probed from the checkout.
+          docker run --rm --entrypoint sh -v "$PWD:/workspace:ro" \
+            -w /workspace knr-ops-toolbox:ci -c '
             for t in kind kubectl helm clusterctl flux sops age aws docker xargs git curl; do
               command -v "$t" >/dev/null || { echo "::error::$t missing on PATH"; exit 1; }
             done
--- a/.github/workflows/bootstrap-rs.yml
+++ b/.github/workflows/bootstrap-rs.yml
@@ -6,10 +6,16 @@ on:
     branches: [main]
     paths:
       - "bootstrap-rs/**"
+      - "mise.toml"
+      - "mise.aws.toml"
+      - ".dockerignore"
       - ".github/workflows/bootstrap-rs.yml"
   pull_request:
     paths:
       - "bootstrap-rs/**"
+      - "mise.toml"
+      - "mise.aws.toml"
+      - ".dockerignore"
       - ".github/workflows/bootstrap-rs.yml"
 
 permissions:
--- a/.github/workflows/bootstrap-rs.yml
+++ b/.github/workflows/bootstrap-rs.yml
@@ -56,3 +56,37 @@ jobs:
 
       - name: Test
         run: cargo test --locked
+
+  docker:
+    name: Docker image build
+    runs-on: ubuntu-24.04-arm
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7.0.1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4.3.0
+
+      # Rot protection for the toolbox image (issue #104): the image
+      # cannot silently break between releases. Loads only; the release
+      # workflow (toolbox-release.yml) pushes.
+      - name: Build the toolbox image (load only)
+        uses: docker/build-push-action@v7.3.0
+        with:
+          context: .
+          file: bootstrap-rs/Dockerfile
+          platforms: linux/arm64
+          push: false
+          load: true
+          tags: knr-ops-toolbox:ci
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+      - name: Smoke the image (binary + tools on PATH)
+        run: |
+          docker run --rm --entrypoint sh knr-ops-toolbox:ci -c '
+            for t in kind kubectl helm clusterctl flux sops age aws docker xargs git curl; do
+              command -v "$t" >/dev/null || { echo "::error::$t missing on PATH"; exit 1; }
+            done
+            knr-bootstrap --help >/dev/null'
--- a/.github/workflows/toolbox-release.yml
+++ b/.github/workflows/toolbox-release.yml
@@ -1,0 +1,133 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+name: toolbox-release
+
+# Publishes ghcr.io/polarsquad/knr-ops-toolbox on semver tags (issue #104).
+# The tag must equal the version in bootstrap-rs/Cargo.toml; the gate job
+# fails the release when they disagree. Signing is cosign keyless via the
+# workflow's GitHub OIDC identity (same identity model as the air-gapped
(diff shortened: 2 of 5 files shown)


shrinedogg (commented):
Verdict: the image definition and Renovate surfaces are verified end-to-end; one warning (CI paths blind spot, inline) and one first-tag watch item (inline). The stale PR metadata found during review (summary promised the release workflow as a stacked follow-up while a00713e included it here; title undercovered the larger half of the diff) has been corrected: retitled and the summary refreshed as part of this review.

Verified locally:
- Both base-image digests resolve and match their tags; the dry-run's digest lookups returned etags identical to the pins.
- rust-toolchain.toml channel 1.98.0 matches the build-stage image (no rustup re-download); min_version matches ARG MISE_VERSION (2026.8.10); the flux2 install name matches mise.toml; go/python/zarf excluded as claimed.
- All nine exact action tags resolve (checkout v7.0.1, upload-artifact v7.0.1, setup-buildx v4.3.0, setup-qemu v4.3.0, build-push v7.3.0, login v4.6.0, metadata v6.2.0, cosign-installer v4.1.2, sbom-action v0.24.2).
- Renovate dry-run (CLI 44.50.1, node 24, both tokens, exit 0): dockerfile manager fileCount 1 / depCount 2 (both FROM pins); regex manager extracts ARG MISE_VERSION=2026.8.10 -> jdx/mise with a pending update and the safe plain autoReplaceStringTemplate; zero Failed to look up; skipReasons are the pre-existing floating-pin classes.
- Pattern probe: the MISE_VERSION matchString matches exactly one line in the Dockerfile with the correct currentValue; template has no escape risk.
- CI green on 98149f1, all six checks including the new Docker image build with its PATH smoke.

Warning: inline, the new CI job does not trigger on mise.toml / mise.aws.toml / .dockerignore changes, undercutting its stated rot protection.

Looks good: digest pins current at review time; mise tool layer single-sourced from the committed configs with no version duplication in the Dockerfile; .dockerignore minimal with defense-in-depth secret exclusions; release workflow has least-privilege permissions with job-scoped packages/id-token grants; gate job pins tag to Cargo.toml version; smoke test covers all twelve lifecycle tools.

.github/workflows/bootstrap-rs.yml
```diff
@@ -56,3 +56,40 @@ jobs:
 
       - name: Test
         run: cargo test --locked
+
+  docker:
+    name: Docker image build
```
shrinedogg, line 67: Warning: this job's rot protection has a blind spot in the workflow-level paths filter (lines 6-12): only bootstrap-rs/** and the workflow file trigger it, but the image's content is equally driven by root mise.toml, mise.aws.toml, and .dockerignore. A Renovate tool bump that touches only mise.toml never builds the image, so a version that fails to install in-image (backend asset rename, new dependency) surfaces only at the next tag release: the silent break this job exists to catch. Suggest adding mise.toml, mise.aws.toml, and .dockerignore to both path filters.

.github/workflows/toolbox-release.yml
```diff
@@ -0,0 +1,133 @@
...
+
+      - name: Docker metadata (X.Y.Z, X.Y, latest)
+        id: meta
+        uses: docker/metadata-action@v6.2.0
+        with:
+          images: ghcr.io/polarsquad/knr-ops-toolbox
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern=latest
```
shrinedogg: Note: type=semver,pattern=latest is not the documented idiom (that is type=raw,value=latest). A constant pattern should render literally, so this likely works, but the release path cannot be exercised until the first tag push. Worth confirming all three tags (X.Y.Z, X.Y, latest) on that first run, plus the QEMU-emulated amd64 cargo build against the 60m timeout.

.github/workflows/bootstrap-rs.yml
shrinedogg: Fixed in 394f34a. Added `mise.toml`, `mise.aws.toml`, and `.dockerignore` to both the `push` and `pull_request` path filters, so changes to every root input consumed by the image trigger the Docker build. The new-head Docker image check passed in 2m22s.

.github/workflows/toolbox-release.yml
shrinedogg: Addressed in 394f34a by removing the redundant undocumented `type=semver,pattern=latest` rule. In metadata-action v6.2.0, the default `latest=auto` flavor causes the documented `{{version}}` semver rule to add `latest` only for stable versions. A focused test against v6.2.0 produced `:0.1.0`, `:0.1`, and `:latest` for `v0.1.0`, while `v0.2.0-rc.1` produced only `:0.2.0-rc.1`. I also built `linux/amd64` on this arm64 host in 145s and ran `knr-bootstrap --help` plus all 12 tool checks inside the amd64 image. This is well within the 60m timeout. The first registry publish, signing, and attestation still need observation when the first tag is cut.

PR #155, opened 2026-09-03
feat(local-talos): environment manifest tree (#105, 1/3)

shrinedogg:
Adds the `mgmt/local-talos/` manifest tree, PR 1 of 3 for #105 (the single-node Talos management cluster environment). Follow-up PRs: environment wiring (bootstrap.toml + mise), then Renovate coverage + docs.

## What's in the tree

- `infrastructure/`: cert-manager and capi-operator HelmReleases, synced from GitHub (GitRepository source like `mgmt/aws`, not the laptop OCI registry - a physical machine cannot reach knr-registry).
- `capi-providers/`: CAPI core v1.14.0 + CABPT v0.7.6 + CACPPT v0.6.4 + CAPT v0.7.0, each with a Flux Kustomization gated on the provider CRDs.
- `clusters/management/`: imperative Cluster + TalosControlPlane (replicas 1, k8s v1.36.0, talosVersion v1.14, allowSchedulingOnControlPlanes via strategic patch) + TinkerbellCluster/TinkerbellMachineTemplate.
- Deliberately absent vs local-host: no CNI addon (Talos ships flannel), no addons/flux-apps or caaph-system (management-only scope fence; the management cluster's own Flux is seeded by knr-bootstrap).

## Deviation from the issue: upstream providers, not the forks

The issue names the shrinedogg CABPT/CACPPT forks. During implementation I verified the fork work was fully upstreamed: fork `main` is an ancestor of `sidero-community/cluster-api-bootstrap-provider-talos` **v0.7.6** and `sidero-community/cluster-api-control-plane-provider-talos` **v0.6.4** (both released 2026-08-30), with zero fork-only commits remaining. The forks have no releases of their own. Pinning upstream releases satisfies the issue's actual requirements (maintained line, v1beta2 contract via `cluster.x-k8s.io/v1beta2: v1beta1` in the released metadata, installer-image handoff commit `494dd22` included) without the fork-release prerequisite. Happy to switch to fork tags if you prefer an explicit fork line.

All three provider CRs set explicit `fetchConfig` URLs: the CAPI operator's embedded clusterctl defaults (CAPI v1.12.10 inside operator 0.28.0) point `talos` at the stale `siderolabs` org, and the well-known CAPT name is `tinkerbell-tinkerbell`.

## Site-specific values needing operator input (flagged, not secrets)

Two values in `clusters/management/cluster.yaml` are per-machine placeholders to be set for real hardware (repo convention: committed site values like `mgmt/aws/clusters/eu-north-1/management/`):

- `spec.controlPlaneEndpoint.host: 192.168.1.50` - the machine's stable IP
- `TinkerbellMachineTemplate.spec.template.spec.hardwareName: talos-mgmt-01` - the Tinkerbell Hardware CR name

The installer image is intentionally not pinned: CABPT (>= v0.7.6, which includes handoff commit `494dd22`) injects the InfraMachine's `status.installerImage` as `machine.install.image` when the infrastructure provider publishes it. Note: upstream CAPT v0.7.0 does not populate that field, so installs currently use the default Image Factory schematic; the CAPT-side resolution is tracked in #156.

## Verification

- `mise run validate`: 49 overlay builds pass (40 before + 9 new local-talos overlays).
- Field names grounded in the released CRD schemas (CACPPT v0.6.4, CAPT v0.7.0) and CAPT's cluster-template: `talosVersion` lives at `controlPlaneConfig.controlplane.talosVersion` in the v1beta1 API, not the spec root; `hardwareName` is the direct hardware binding.
- End-to-end run still requires the operator-side Tinkerbell prerequisite (boots/PXE stack + Hardware CR), documented in PR 3.

### DIFF
--- a/mgmt/local-talos/clusters/management/cluster.yaml
+++ b/mgmt/local-talos/clusters/management/cluster.yaml
@@ -10,10 +10,12 @@
 #   spec.controlPlaneEndpoint.host       - the machine's stable IP
 #   TinkerbellMachineTemplate hardwareName - your Tinkerbell Hardware CR name
 #
-# The installer image is deliberately NOT pinned here: CAPT resolves it from
-# the Hardware (status.installerImage) and CABPT injects it as
-# machine.install.image, so hardware-specific system extensions arrive
-# without hand-maintained image refs (#105).
+# The installer image is deliberately NOT pinned here: CABPT (>= v0.7.6)
+# injects the InfraMachine's status.installerImage as machine.install.image
+# when the infrastructure provider publishes it. Upstream CAPT v0.7.0 does
+# not populate that field (#156), so installs use the default Image Factory
+# schematic until the CAPT-side resolution lands; CABPT treats the absent
+# field as "no override".
 ---
 apiVersion: cluster.x-k8s.io/v1beta2
 kind: Cluster
--- a/mgmt/local-talos/clusters/management/cluster.yaml
+++ b/mgmt/local-talos/clusters/management/cluster.yaml
@@ -1,0 +1,90 @@
+# Single-node Talos management cluster for the local-talos environment (#105).
+#
+# Shape: imperative (explicit controlPlaneRef), NOT ClusterClass topology —
+# Tinkerbell hardware selection is operator/site-specific and one machine
+# carries the whole management plane.
+#
+# SITE-SPECIFIC VALUES (edit before deploying to your metal; the repo
+# convention for management clusters is committed site values, like
+# mgmt/aws/clusters/eu-north-1/management/):
+#   spec.controlPlaneEndpoint.host       - the machine's stable IP
+#   TinkerbellMachineTemplate hardwareName - your Tinkerbell Hardware CR name
+#
+# The installer image is deliberately NOT pinned here: CAPT resolves it from
+# the Hardware (status.installerImage) and CABPT injects it as
+# machine.install.image, so hardware-specific system extensions arrive
+# without hand-maintained image refs (#105).
+---
+apiVersion: cluster.x-k8s.io/v1beta2
+kind: Cluster
+metadata:
+  name: local-talos-management
+  namespace: default
+  labels:
+    knr-ops.polarsquad.com/environment: local-talos
+spec:
+  clusterNetwork:
+    pods:
+      cidrBlocks:
+        - 192.168.0.0/16
+    serviceDomain: cluster.local
+    services:
+      cidrBlocks:
+        - 10.128.0.0/12
+  controlPlaneEndpoint:
+    host: 192.168.1.50
+    port: 6443
+  controlPlaneRef:
+    apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+    kind: TalosControlPlane
+    name: local-talos-management
+  infrastructureRef:
+    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+    kind: TinkerbellCluster
+    name: local-talos-management
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: TinkerbellCluster
+metadata:
+  name: local-talos-management
+  namespace: default
+spec: {}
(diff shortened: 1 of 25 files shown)


mgmt/local-talos/clusters/management/cluster.yaml
```diff
@@ -0,0 +1,90 @@
...
+# Tinkerbell hardware selection is operator/site-specific and one machine
+# carries the whole management plane.
+#
+# SITE-SPECIFIC VALUES (edit before deploying to your metal; the repo
+# convention for management clusters is committed site values, like
+# mgmt/aws/clusters/eu-north-1/management/):
+#   spec.controlPlaneEndpoint.host       - the machine's stable IP
+#   TinkerbellMachineTemplate hardwareName - your Tinkerbell Hardware CR name
+#
+# The installer image is deliberately NOT pinned here: CAPT resolves it from
```
shrinedogg: Warning: this comment overstates what the pinned CAPT does. I checked the TinkerbellMachine v1beta1 CRD and types in upstream tinkerbell/cluster-api-provider-tinkerbell at v0.7.0 and at main: there is no `status.installerImage` field anywhere (status keys are addresses, conditions, errorMessage, errorReason, initialization, instanceStatus, ready, targetNamespace), and no PR or issue in that repo adds one. With upstream CAPT the 494dd22 read will always find an absent field, which the commit deliberately treats as "no override", so nothing breaks, but the node installs and upgrades with the default Image Factory schematic and the #105 acceptance criterion "installer image resolved from the Hardware" is not actually delivered by this combination. Suggest correcting the comment here and tracking the CAPT-side resolution explicitly; filed as #156.

shrinedogg (commented):
Manifest tree reviewed against the checked-out head. All structural claims verified; one behavioral claim needs correcting.

Verified locally:
- `mise run validate` passes: 49 overlays build (40 on main plus 9 new local-talos overlays; the PR body says 54/45, actual is 49/40).
- Fork ancestry via the GitHub compare API: both shrinedogg fork `main` branches are 0 commits ahead of sidero-community v0.7.6 / v0.6.4, so the upstream releases contain all fork work including 494dd22; the forks publish zero releases.
- capi-operator 0.28.0 embeds CAPI v1.12.10 (go.mod); its clusterctl defaults point `talos` at the stale siderolabs org and name CAPT `tinkerbell-tinkerbell`, so the explicit fetchConfig URLs and provider names here are correct.
- Field paths checked against the released CRDs: `controlPlaneConfig.controlplane.talosVersion` (CACPPT v0.6.4), `hardwareName` and the valid empty `spec: {}` (CAPT v0.7.0); both providers' metadata.yaml declare contract v1beta2 for the pinned series.
- Versions match mgmt/aws and mgmt/local-host (cert-manager 1.21.1, capi-operator 0.28.0, CAPI v1.14.0); flux-ks patterns mirror local-host; existing addon selectors are scoped to environment=local-host so the local-talos label cannot collide.

Warning: the installer-image handoff as described does not function with upstream CAPT v0.7.0 (details inline on cluster.yaml; tracked in #156). Not a manifest defect; a scoping question for #105.

Suggestion: AGENTS.md's own golden rule asks for same-PR updates when repository structure changes, and this PR adds a whole mgmt/ environment. Consider adding the local-talos tree entry to AGENTS.md here rather than deferring all of it to PR 3, or note the deferral explicitly.

Note: the pre-placed `# renovate:` annotations in the provider CRs currently match no manager (the annotation manager targets bootstrap.toml equals-style pins; the capi-providers managers hardcode depName and are scoped to aws|local-host paths). Fine as PR 3 groundwork, but PR 3 will need a new matchStrings pattern, not just a path extension. The HelmRelease pins are already covered by the built-in flux manager's mgmt/** scope.

Looks good: the tinkerbell-tinkerbell provider name (deviating from the issue text) is verified against the clusterctl constant; the v1beta2 health-check pattern has precedent in local-host; the GitHub-vs-OCI source rationale is documented in the tree where a future reader will look for it.

PR #176, opened 2026-09-03
fix: toolbox wrapper bash 3.2 compat and post-pivot REGISTRY_HOST forwarding (#172, #174)

### LINKED ISSUE #172: toolbox-run.sh fails on macOS system bash 3.2 (empty array expansion under set -u)
## Problem

`mise run bootstrap` (and pivot/teardown) fails immediately on stock macOS
when invoked through `scripts/toolbox-run.sh`:

```
scripts/toolbox-run.sh: line 149: TTY_ARGS[@]: unbound variable
scripts/toolbox-run.sh: line 149: CLI_ARGS[@]: unbound variable
```

Two triggers, same root cause:

- non-interactive invocation (no TTY, or stdout piped): `TTY_ARGS=()` stays
  empty
- `bootstrap` with no extra args: `CLI_ARGS=("$@")` is an empty array

Under `set -u`, expanding an empty array with `"${ARR[@]}"` is an unbound
variable error on bash < 4.4. macOS system bash is 3.2.57, so the wrapper
fails on any Mac that does not shadow `env bash` with a newer shell. CI and
Linux hosts use bash 5, which is why this never surfaced there.

## Reproduction

On a Mac with only the system bash:

```sh
TOOLBOX_IMAGE=knr-ops-toolbox:dev mise -E local-host run bootstrap
```

## Proposed fix

Bash-3.2-safe expansion idiom at the two sites in `scripts/toolbox-run.sh`
(PASS_ENV is never empty and needs no change):

```diff
-exec "$CONTAINER_ENGINE" run --rm "${TTY_ARGS[@]}" \
+exec "$CONTAINER_ENGINE" run --rm ${TTY_ARGS[@]+"${TTY_ARGS[@]}"} \
   ...
-  "${CLI_ARGS[@]}"
+  ${CLI_ARGS[@]+"${CLI_ARGS[@]}"}
```

The usage text in the same file still says `KNR_OPS_PROFILE aws |
local-host`; worth updating to include `local-talos` while touching the
file.

## Notes

Discovered while running the local-host regression for #105 acceptance on
macOS 26 (Docker Desktop) with the system bash 3.2.57 and no homebrew bash.
The raw `docker run` invocation from docs/operations.md is unaffected (it
does not use the wrapper).

### LINKED ISSUE #174: bootstrap CLI: post-pivot OCI publish misses REGISTRY_HOST forwarding, fails in toolbox runs
## Problem

In toolbox runs, the local-host bootstrap fails at the post-pivot OCI
publish with:

```
>>> Publishing the current checkout as the OCI artifact...
curl: (7) Failed to connect to localhost port 5001 after 0 ms: Couldn't connect to server
ERROR: local registry is unavailable at localhost:5001
Error: 'mise -E local-host run oci-push' failed with exit status: 1
```

Root cause: `pivot_seed_target` in `bootstrap-rs/src/main.rs` spawns
`mise -E local-host run oci-push` with `REGISTRY_PORT`/`OCI_REPOSITORY`/
`OCI_TAG` but, unlike the initial publish in the same file, does not
forward `REGISTRY_HOST` and uses `cfg.registry_port` instead of
`cfg.registry_endpoint()`. Inside the toolbox the task therefore probes
`localhost:5001` (nothing listens there in the container) instead of
`knr-registry:5000` on the kind network.

The bug predates the local-talos work: the call site has not forwarded
`REGISTRY_HOST` since the pivot port (e05320c). It only fires in toolbox
mode on a local (OCI-synced) environment; host-native runs work because
`localhost:5001` is correct there.

## Reproduction

On a host with Docker:

```sh
docker build -f bootstrap-rs/Dockerfile -t knr-ops-toolbox:dev .
TOOLBOX_IMAGE=knr-ops-toolbox:dev mise -E local-host run bootstrap
# fails after the clusterctl move, at pivot_seed_target
```

## Proposed fix

Mirror the initial-publish call site: forward `REGISTRY_HOST` and use
`cfg.registry_endpoint()` for host and port:

```rust
let (registry_host, oci_port) = cfg.registry_endpoint();
let status = Command::new("mise")
    .args(["-E", &cfg.profile, "run", "oci-push"])
    .env("REGISTRY_PORT", oci_port.to_string())
    .env("REGISTRY_HOST", &registry_host)
    ...
```

## Notes

Found during the local-host regression run for #105 acceptance: the pivot
completed (clusterctl move succeeded, both clusters on the management
cluster), then the seed publish failed. The bootstrap is rerun-safe, so
the run resumes at the failed step once the binary is fixed.

shrinedogg:
## What

Two independent fixes found while running the local-host regression for #105 acceptance on macOS; one commit each, one issue each.

**`fix: expand empty arrays safely in toolbox-run.sh for macOS bash 3.2`** (closes #172)

Under `set -u`, `"${ARR[@]}"` on an empty array is an unbound variable error on bash < 4.4, and macOS system bash is 3.2.57. The wrapper failed before reaching Docker in two trigger paths: non-interactive invocation (`TTY_ARGS` empty) and `bootstrap` with no extra args (`CLI_ARGS` empty). The `${ARR[@]+"${ARR[@]}"}` idiom expands safely on both bash 3.2 and 5. The usage text also gains `local-talos`.

**`fix: forward REGISTRY_HOST in the post-pivot OCI publish`** (closes #174)

`pivot_seed_target` spawned `mise run oci-push` without `REGISTRY_HOST` and with `cfg.registry_port` instead of `cfg.registry_endpoint()`, so toolbox runs probed `localhost:5001` inside the container after an otherwise successful pivot. The fix mirrors the initial publish call site: host and port come from `registry_endpoint()` (`knr-registry:5000` in the toolbox, `localhost:5001` on the host).

## Verification

- #172: before the fix the wrapper died at the `exec` line on this Mac (system bash 3.2.57, no homebrew bash); after the fix the same invocation runs the container end to end. `bash -n` clean; `mise run validate` shell check passes.
- #174: `cargo fmt --check`, clippy `-D warnings`, and 50/50 tests pass. In the live regression run, the pre-fix binary failed at the post-pivot publish (`localhost:5001` probe inside the toolbox); with the patched binary the publish succeeds against `knr-registry:5000` and the run proceeded through the seed into the reconciliation watch.
- Full bootstrap + teardown evidence comes from the #105 acceptance regression run; both fixes are the exact patches it ran with.

### DIFF
--- a/bootstrap-rs/src/main.rs
+++ b/bootstrap-rs/src/main.rs
@@ -2088,9 +2088,14 @@ async fn pivot_seed_target(
     // contain the management manifests.
     if cfg.is_local() {
         println!(">>> Publishing the current checkout as the OCI artifact...");
+        // Same endpoint forwarding as the initial publish: the mise
+        // oci-push task probes REGISTRY_HOST:REGISTRY_PORT, which is the
+        // kind-network name inside the toolbox and localhost on the host.
+        let (registry_host, oci_port) = cfg.registry_endpoint();
         let status = Command::new("mise")
             .args(["-E", &cfg.profile, "run", "oci-push"])
-            .env("REGISTRY_PORT", cfg.registry_port.to_string())
+            .env("REGISTRY_PORT", oci_port.to_string())
+            .env("REGISTRY_HOST", &registry_host)
             .env("OCI_REPOSITORY", &cfg.oci_repository)
             .env("OCI_TAG", &cfg.oci_tag)
             .status()
--- a/scripts/toolbox-run.sh
+++ b/scripts/toolbox-run.sh
@@ -27,7 +27,8 @@ Env:
   TOOLBOX_IMAGE   image reference (default: ${TOOLBOX_IMAGE};
                   build locally with: docker build -f bootstrap-rs/Dockerfile \\
                     -t knr-ops-toolbox:dev . && TOOLBOX_IMAGE=knr-ops-toolbox:dev)
-  KNR_OPS_PROFILE aws | local-host (default: the mise environment in use)
+  KNR_OPS_PROFILE aws | local-host | local-talos
+                  (default: the mise environment in use)
 EOF
   exit 2
 }
@@ -146,7 +147,7 @@ case "$LIFECYCLE" in
   *)         usage ;;
 esac
 
-exec "$CONTAINER_ENGINE" run --rm "${TTY_ARGS[@]}" \
+exec "$CONTAINER_ENGINE" run --rm ${TTY_ARGS[@]+"${TTY_ARGS[@]}"} \
   -v "$REPO_ROOT:/workspace" \
   -w /workspace \
   -v "$SOCK_SOURCE:/var/run/docker.sock" \
@@ -154,4 +155,4 @@ exec "$CONTAINER_ENGINE" run --rm "${TTY_ARGS[@]}" \
   -e KUBECONFIG="$KUBECONFIG_IN" \
   "${PASS_ENV[@]}" \
   "$TOOLBOX_IMAGE" \
-  "${CLI_ARGS[@]}"
+  ${CLI_ARGS[@]+"${CLI_ARGS[@]}"}


shrinedogg:
Regression evidence, completed after this PR's branch was cut (run from a clean checkout of origin/main at 85f7c2e plus exactly these two patches, toolbox image rebuilt from them):

- Clean-slate `mise -E local-host run bootstrap`: exit 0. Full chain: kind bootstrap, OCI publish, Flux handoff, workload reconciliation watch, pivot, post-pivot seed publish (succeeded against knr-registry:5000; this is the step that failed before the REGISTRY_HOST fix), kind deletion.
- Post-pivot state: `local-management` and `local-workload` both Provisioned at v1.37.0; management `flux-apps` Kustomization Ready; workload root Kustomization Ready.
- Application delivery: podinfo HelmRelease "Helm install succeeded" (chart 6.14.1), pod Running 1/1.
- `mise -E local-host run teardown`: exit 0 via the post-pivot path (CAPD workload deleted through CAPI, management containers removed, registry removed). Only the unrelated airgap clusters remained on the host.

One environment note for reviewers: a rerun hitting the half-pivoted state left by the pre-fix #174 failure (kind still present, inventory already moved, kind-side Flux suspended) times out in the step-5 watch; teardown plus a fresh bootstrap recovers. That state only arises from the bug this PR fixes.

PR #188, opened 2026-09-06
Fix duplicate Renovate Helm registry URLs

lmilbaum:
## Summary
- remove duplicate Helm registry assignments in Renovate
- centralize the registry per chart in packageRules
- add a regression check for single registry URL per tracked chart pin

## Why
Renovate was logging the duplicate registry warning because the same Helm package was assigned the same registry multiple times.

### DIFF
--- a/tests/test-renovate-coverage.py
+++ b/tests/test-renovate-coverage.py
@@ -5,8 +5,15 @@ The catalog drift check was retired; these detection checks are what proves
 a bump cannot silently stop being proposed for the non-manifest surfaces
 (issue #74 acceptance criteria). Each entry maps a tracked file to the
 depNames Renovate must extract from it via custom.regex managers.
+
+The repo-level regression for #187 is checked separately via a full
+`--dry-run=full` run: the custom-regex harness only exercises the regex
+manager and therefore cannot merge the flux/helm and regex registry inputs
+for the same chart, which is where Renovate emits the real warning.
 """
 
+import os
+import subprocess
 import sys
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


tests/test-renovate-coverage.py
```diff
@@ -40,6 +40,40 @@
...
 }
 
+REQUIRED_SINGLE_REGISTRY = {
+    "cert-manager": "https://charts.jetstack.io",
+    "cluster-api-operator": "https://kubernetes-sigs.github.io/cluster-api-operator",
+}
+
+
+def assert_single_registry_per_helm_package(result) -> list[str]:
+    """Regression for issue #187: each tracked Helm pin should resolve to one registry."""
```
shrinedogg: Warning: this check passes against the pre-fix config. I ran the harness with repo_root pointed at origin/main (registryUrlTemplate still in the managers, no packageRules) and it returns zero errors: each regex-extracted dep carries exactly one registry both before and after the fix. The #187 warning ("Excess registryUrls found for datasource lookup") only appears in a full run where the flux manager and the two regex pin sites merge three identical registryUrls for the same chart, and the harness enables custom.regex only, so that merge never happens here. The assertion is still useful (deleting one of the new packageRules leaves the pivot.sh pins with no registry and fails), but it does not reproduce #187 and "Regression for issue #187" overstates it. Suggest either noting the limitation in the docstring, or asserting at the level where the bug manifested (full dry-run, grep for "Excess registryUrls").

shrinedogg (commented):
Verified locally:
- Full dry-run (renovate 44.50.1, both tokens) on origin/main reproduces WARN "Excess registryUrls found for datasource lookup" for cert-manager and cluster-api-operator (3 identical URLs each); the same run on this head has zero occurrences and no other WARN/ERROR lines.
- Extraction stats identical before/after: 97 files, 281 deps (regex 136, flux 13).
- Updates still resolve: cluster-api-operator 0.29.0 proposed from bootstrap.toml and pivot.sh; cert-manager 1.21.1 current.
- tests/test-renovate-coverage.py passes on this head; CI checks green.

Warning: the new registry assertion passes against the pre-fix config (details inline), so it does not guard the #187 failure mode itself.

Looks good: centralizing the registry per chart in packageRules is the right shape, the platform-charts grouping still matches, and no dangling references to registryUrlTemplate remain.

Note: PR is still marked draft.

tests/test-renovate-coverage.py
lmilbaum: Thanks — that’s a good catch.

The earlier assertion was narrower than the real failure mode: the harness only exercises the custom.regex manager, and it never merges the Flux-managed Helm entries with the regex-derived pins that trigger the full-repo warning, "Excess registryUrls found for datasource lookup".

I updated the regression to keep the per-package single-registry check, and also assert that a full repo dry-run remains free of that duplicate-registry warning. That covers the actual https://github.com/polarsquad/knr-ops/issues/187 failure mode instead of only the regex-only case.

PR #227, opened 2026-09-09
fix(airgap): resolve zarf init package by glob instead of pinned version

shrinedogg:
## Problem

Every nightly `air-gapped` run has failed at the `zarf init` step since 2026-09-01 ([latest run](https://github.com/polarsquad/krops/actions/runs/34303671677)):

```
ERR unable to load package: opening ".../airgap/archives/zarf-init-arm64-v0.83.0.tar.zst": no such file or directory
FAIL: zarf init
```

`airgap/scripts/offline-run.sh` hardcoded the init package filename, but that file is produced by `zarf tools download-init` in `build-package.sh`, which names it after the zarf CLI version pinned in `mise.toml`. Renovate bumped the pin (v0.83.0 -> v0.84.0 in aa41a96, then v0.85.0 in #209) and no Renovate manager covers the hardcoded filename, so the bundle stopped containing the file the deploy script looks for. Last green nightly: 2026-08-31 (ran before aa41a96 merged); 8 consecutive failures since.

## Fix

Establish the init package's provenance at build time instead of pinning a version at deploy time:

- `build-package.sh`: `zarf tools download-init` now runs into a clean staging directory, the build fails unless exactly one output was generated, and that output is renamed to the stable `archives/zarf-init-arm64.tar.zst`. `mise.toml` stays the sole version declaration; the bundle provably contains the archive built by that pinned zarf CLI, regardless of stale archives left in `archives/` from earlier runs.
- `offline-run.sh` / `stage-and-create-cluster.sh`: reference the stable filename, no version arithmetic.
- `docs/airgap.md`: the bundle table documents the stable filename and its build-time provenance.

## Verification

- `bash -n` on all three scripts.
- Staging block simulated with a stubbed `download-init`: 1 output renames to `zarf-init-arm64.tar.zst`; 0 and 2 outputs fail with `ERROR: expected exactly one init package`.
- `python3 airgap/tests/test-airgap-image-digests.py --all` passes.

Note: the `air-gapped` workflow runs on upstream main only, so full end-to-end validation happens on the next nightly after merge (or a manual dispatch).

### DIFF
--- a/airgap/scripts/build-package.sh
+++ b/airgap/scripts/build-package.sh
@@ -63,9 +63,19 @@ fi
 echo "==> 3/5 offline host assets, workload-node images, and OCI charts"
 mkdir -p airgap/archives
 
+INIT_STAGING=$(mktemp -d "${TMPDIR:-/tmp}/krops-zarf-init.XXXXXX")
 mise x -- zarf tools download-init \
   --architecture arm64 \
-  --output-directory airgap/archives
+  --output-directory "$INIT_STAGING"
+shopt -s nullglob
+init_outputs=("$INIT_STAGING"/*)
+shopt -u nullglob
+if [ "${#init_outputs[@]}" -ne 1 ]; then
+  echo "ERROR: expected exactly one init package from zarf tools download-init, found ${#init_outputs[@]}" >&2
+  exit 1
+fi
+mv "${init_outputs[0]}" airgap/archives/zarf-init-arm64.tar.zst
+rm -rf "$INIT_STAGING"
 
 HOST_IMAGES=(
   kindest/node:v1.37.0@sha256:a1ed56cfb0e7b93589bdf97c8cd566405a265939e3620fc4f5de89adff580ae5
--- a/airgap/scripts/offline-run.sh
+++ b/airgap/scripts/offline-run.sh
@@ -122,7 +122,18 @@ else
 fi
 
 step "3. zarf init"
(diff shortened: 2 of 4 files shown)


lmilbaum (changes requested):
Request changes:

1. Please remove the explanatory comments added to the scripts. The glob/cardinality behavior is clear from the code; the version-drift rationale belongs in the PR description, commit message, or documentation.

2. Please do not select the init archive at deploy time with a glob. This decouples the selected archive from the Zarf version pinned in mise.toml: a single stale archive would be accepted, while retained archives make deployment fail. Keep the package selection deterministic and establish its provenance at build time from the mise pin (for example, generate the init package in a clean archives directory, validate exactly one generated output, then give it a stable local filename for consumers). This keeps mise.toml as the sole version declaration while ensuring the bundle contains the archive built by that pinned Zarf version.

shrinedogg:
Both points addressed in 28b802e:

1. Explanatory comments removed from `offline-run.sh` and `stage-and-create-cluster.sh`. The rationale is in the commit message and the provenance note in `docs/airgap.md`.

2. Deploy-time glob dropped. `build-package.sh` now runs `zarf tools download-init` into a clean staging directory, fails the build unless exactly one output was generated, and renames it to the stable `archives/zarf-init-arm64.tar.zst`. Both deploy scripts reference that stable name, so the selected archive is always the one built by the zarf CLI pinned in `mise.toml`; stale archives left in `archives/` can neither be silently accepted nor fail the deploy. Verified the staging block with a stubbed `download-init` for the 0/1/2-output cases.

Issue #236, opened 2026-09-10
azure: drive subscription prep from Git via ASO, shrink azure-bootstrap to SP + providers


shrinedogg:
## Context

`mise -E azure run azure-bootstrap` currently performs six one-time
subscription operations: register 7 resource providers, create the
`krops-shared` resource group, create the `krops-aso` user-assigned
identity, create the `krops-capz` service principal, grant it
Contributor, and grant it RBAC Administrator. Three of those are
ARM-side and can be declared as ASO resources that the Flux instance in
kind reconciles before the pivot, shrinking the task to the three
operations only Azure CLI can do.

## Moves to Git (ASO, reconciled in kind pre-pivot)

- `krops-shared` resource group to an ASO `ResourceGroup`.
- `krops-aso` to an ASO `UserAssignedIdentity`, created by ASO instead of
  `az identity create`. Export `clientId`/`principalId` via
  `operatorSpec.configMaps` and have the flux-apps Kustomization
  `substituteFrom` the export, replacing the literal
  `AZURE_ASO_CLIENT_ID`/`AZURE_ASO_PRINCIPAL_ID` in `cluster-vars` (same
  pattern as the OIDC issuer export). Its RoleAssignment already exists
  as ASO resources.
- The `azure-bootstrap` task then reduces to: register providers, create
  the SP, grant its two roles. It prints only the SP secret.

## Stays in azure-bootstrap (irreducible)

- SP creation: it is an Entra ID object created via Microsoft Graph, and
  it is the credential ASO uses for its first call. No workload identity
  in kind (Entra does not trust its OIDC issuer).
- The SP's own initial role grants: ASO cannot grant itself the roles it
  needs to run.
- Resource provider registration: no ASO kind.

## Pitfalls to handle

- Set `serviceoperator.azure.com/reconcile-policy: detach-on-delete` on
  the bootstrap-created ASO resources (RG, UAMI): a `flux uninstall` or
  namespace delete in kind before `kind delete cluster` would otherwise
  delete the real Azure resources.
- First reconcile is two-phase (Flux waits for the configMaps export);
  expected, same as the existing azure-identity transient failure.
- Adoption after pivot: the management cluster's Flux re-applies the same
  CRs and ASO adopts the existing Azure resources via idempotent PUT.
  Add an explicit adoption check to the live acceptance run.
- Extend `tests/test-azure-identity-chain.py` to cover the UAMI export.

## Explicitly out of scope

[...]

PR #237, opened 2026-09-10
bootstrap-rs: post-kind-create-task and pivot-manifests knobs

shrinedogg:
## Summary

Two new optional `[environments.<name>]` knobs for the krops-bootstrap CLI (issue #236). No behavior change for existing environments (both default to empty/absent).

- **`post-kind-create-task`** (optional string): name of a mise task in the active profile (`mise.<env>.toml`) that runs once the kind bootstrap cluster exists, on both the fresh-create and healthy-reuse paths (a rerun re-asserts the setup). Non-zero exit aborts the bootstrap before Flux installs anything. Same invocation shape as the existing `mise -E <profile> run aws-credentials` pivot call.
- **`pivot-manifests`** (optional list): plain (unencrypted) manifests applied to the pivot target before `clusterctl move`, after the provider CRs and before `pivot-sops-secrets`. The workload-identity replacement for `pivot-sops-secrets`.

## Pre-flight (Task 0.1)

Verified against the CAPZ v1.27.0 infrastructure-components manifest (41,270 lines) before any code:

- `AzureClusterIdentity` CRD `spec.type` enum includes `WorkloadIdentity`.
- Service accounts in `capz-system`: `capz-manager` (CAPZ), `azureserviceoperator-default` (bundled ASO).
- Both deployments project SA tokens with audience `api://AzureADTokenExchange` (ASO at `/var/run/secrets/tokens`, CAPZ at `/var/run/secrets/azure/tokens`).
- `connectedk8s` az CLI extension version 1.11.3 (used by the follow-up PR's `arc-federate` task).
- GNU sed 4.9 in the kind node; the apiserver issuer patch simulation against a real kind `kube-apiserver.yaml` verified (arc issuer becomes the first `--service-account-issuer` flag, kubeadm default kept second).

## Changes

- `bootstrap-rs/src/config.rs`: `post_kind_create_task: Option<String>` and `pivot_manifests: Vec<String>` on `Environment` (serde defaults).
- `bootstrap-rs/src/main.rs`: hook runs right after `ensure_kind_cluster`; pivot manifests applied in `pivot_install_capi_in_target` (Phase 3); source-order test pins provider-manifests < pivot-manifests < pivot-sops-secrets.
- `tests/test-bootstrap-config.py`: validates pivot-manifests files exist and are not `*.sops.yaml`; validates the hook task exists in `mise.<env>.toml`.
- `docs/bootstrap-cli.md`, `AGENTS.md`: knob documentation.

## Verification

`cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test --locked` (56 passed), `tests/test-bootstrap-config.py` (`OK (3 charts, 4 environments)`), `mise run validate`, `mise run docs-build` (strict, 0 warnings) / `docs-test` (9 passed). Both validator failure branches were exercised with deliberate fixtures.

Refs #236

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -82,9 +82,11 @@ resources. There is no app source code here, only declarative infrastructure.
   rerun-safe-by-default semantics. Chart versions it installs imperatively
   are Renovate-annotated constants in `src/main.rs`. CI (bootstrap-rs
   workflow) runs fmt/clippy/build/test; the toolchain is pinned in
-  `rust-toolchain.toml`. Two config-driven knobs added for azure:
+  `rust-toolchain.toml`. Four config-driven knobs added for azure:
   `pivot-sops-secrets` (SOPS manifests applied in the pivot target before
-  the move) and `teardown.manual` (refuse with operator text).
+  the move), `teardown.manual` (refuse with operator text),
+  `post-kind-create-task` (mise task run after kind creation) and
+  `pivot-manifests` (plain manifests applied in the target before the move).
 - `bootstrap.sh` / `pivot.sh` / `teardown.sh`: the shell equivalents of the
   CLI's phases. Kept until the binary completes full parity runs per
   environment, then retired (issues #92/#95/#100). The lifecycle mise tasks
--- a/bootstrap-rs/src/config.rs
+++ b/bootstrap-rs/src/config.rs
@@ -78,6 +78,18 @@ pub struct Environment {
     /// Cluster hierarchy. Relative to the repository root.
     #[serde(default)]
     pub pivot_sops_secrets: Vec<String>,
+    /// Mise task run once the kind bootstrap cluster exists (issue #236):
+    /// provider-specific setup that must be in place before Flux installs
+    /// anything (azure: Arc OIDC federation so CAPZ/ASO can use workload
+    /// identity instead of a service-principal secret).
+    #[serde(default)]
+    pub post_kind_create_task: Option<String>,
+    /// Plain (unencrypted) manifests applied to the pivot target before
+    /// `clusterctl move` (issue #236): non-secret objects the moved resources
+    /// reference by name that clusterctl does not carry (the workload-identity
+    /// replacement for pivot-sops-secrets). Relative to the repository root.
+    #[serde(default)]
+    pub pivot_manifests: Vec<String>,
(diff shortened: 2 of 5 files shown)


shrinedogg (commented):
Verdict: ready to merge, nothing to fix.

Verified locally against fresh origin/main and pull/237 head 4cad80e:
- Diff is the 5-file change (+126/-2), identical to pre-push.
- CI 8/8 pass (incl. Build/lint/test with fmt + clippy + 56 tests, strict site build, Renovate digest pinning, Docker image build). MERGEABLE.
- DCO sign-off present on all 5 commits.
- Re-ran gates on the PR ref: cargo fmt --check clean, cargo clippy exit 0 (only the pre-existing Cargo.toml semver-metadata manifest notice), cargo test --locked 56/56, tests/test-bootstrap-config.py OK (3 charts, 4 environments), mise run validate green.
- Probes: the post-kind-create hook call site is exactly one, inside run_bootstrap only (never teardown); mise is in required_tools for every environment so the hook binary is guaranteed; the pivot-manifests ordering-test anchors resolve to the real apply loops in the asserted provider-manifests < pivot-manifests < pivot-sops-secrets order.

Looks good:
- Both knobs are serde(default) optional: zero behavior change for aws/local-host/local-talos, confirmed by the unchanged loads_the_repository_config test and the Rendered Flux diff showing no rendered changes.
- Hook placement is correct fail-fast ordering: after kind exists, before Flux installs anything that would depend on the federation.
- pivot-manifests applies after the provider CRs (CAPZ must exist before its identity Secret is meaningful) and before the move, mirroring the proven pivot-sops-secrets pattern minus decryption.
- The cross-check validates both knobs (existence + not-*.sops.yaml for manifests; [tasks.<name>] must exist in mise.<env>.toml for the hook); both failure branches were exercised with deliberate fixtures.
- docs/bootstrap-cli.md and AGENTS.md describe both knobs; title matches the diff exactly.

Non-blocking observations (no action requested):
- The ordering test pins source order by string position; a future rename of those loop texts fails it loudly (red test, not silent drift). Acceptable guard, documented in the test comment.
- The post-kind-create-task docs say 'azure uses it for Arc OIDC federation' before PR 2 wires that up; forward-looking but referenced to #236.

Refs #236

PR #241, opened 2026-09-11
chore(renovate): propose latest dependency version only

lmilbaum:
## Summary

- configure Renovate to use one latest-version PR per dependency
- avoid parallel major and non-major update PRs

## Validation

- `git diff --check`
- Renovate dry-run not run: this environment has no GitHub auth token available

### DIFF
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -41,8 +41,10 @@ Renovate discovers and updates versions in:
 
 Grouping rules keep Flux updates together, CAPI updates together, imperative
 chart pins with their declarative counterparts, and node-version updates
-separate. Base images in `bootstrap-rs/Dockerfile` and air-gap images are
-digest-pinned while retaining readable tags. Nothing automerges.
+separate. Renovate proposes one PR at the newest available version for each
+dependency, rather than parallel major and non-major update PRs. Base images
+in `bootstrap-rs/Dockerfile` and air-gap images are digest-pinned while
+retaining readable tags. Nothing automerges.
 
 ## Toolbox release version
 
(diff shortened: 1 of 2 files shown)


shrinedogg (approved):
Verified locally (pinned `renovate@44.69.8` from validate.yml, Node 24.20.0, `GITHUB_COM_TOKEN` set, `--platform=local`):

- `--print-config` on `origin/main` resolves `separateMajorMinor: true` (the default; `config:recommended` does not set it), while this branch resolves `false`. The change is a real behavior change, not a redundant override.
- A/B dry-runs on identical trees with one forced pin (`docker.io/library/registry:2.7.1` in `airgap/images.txt`): the base tree plans a parallel minor (2.8.3, `renovate/docker.io-library-registry-2.x`) and major (3.1.1, `...-3.x`) for the same dependency; this branch plans only the newest (3.1.1). Digest-pin updates stay separate on both trees (pre-existing, unaffected).
- Extraction is identical between the two trees: 125 files, 332 deps, same skipReason histogram (invalid-version: 6, invalid-value: 4, disabled: 2). The setting affects branch planning only.
- The `kindest/node` rule (renovate.json5:475, `separateMajorMinor: true`) and the CAPZ one-minor-at-a-time rule (renovate.json5:464, `separateMultipleMinor: true`) both remain in the resolved config, so the docs' "node-version updates separate" and the ASO upgrade constraint still hold.
- CI is green; "Rendered Flux diff" skips as expected for a config-only change.
- `docs/dependencies.md` sentence is accurate as scoped ("parallel major and non-major update PRs").

Title matches the single commit and the diff.

PR #242, opened 2026-09-11
chore: group and pin GitHub Actions updates

lmilbaum:
Group only GitHub Actions references in Renovate, excluding workflow images, runners, tool versions, and reusable workflows. Pin the remaining Pages action references to exact versions.

Add an offline unit test using Renovate’s rule engine, wire it into CI, and document the grouping scope. The test passes all 11 cases and rejects the original broad rule in all seven non-action workflow cases.

Validation: `mise run validate`, Renovate configuration validation and full dry-run, the grouping unit test, and `git diff --check` passed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -203,7 +203,11 @@ discovered) and the digest-pinning test run in CI only (the validate.yml
 shared harness `tests/renovate_harness.py` (issue #97): a new Renovate test
 is a file list plus assertions, never a copy of the subprocess/parsing
 logic. The harness needs Node >= 24.11 (renovate's `engines` field);
-locally: `mise x node@24 -- python3 tests/test-renovate-coverage.py`. They
+locally: `mise x node@24 -- python3 tests/test-renovate-coverage.py`.
+The offline unit test `tests/test-renovate-actions-grouping.py` also runs in
+that CI job and uses the shared harness to apply Renovate’s real package-rule
+engine, checking that only action dependencies join the GitHub Actions group.
+Run locally with Renovate on PATH and Node >= 24.11. These tests
 do not cover lookup liveness or the replacement path; only the
 dry-run and the handlebars simulation cover those.
 
--- a/renovate.json5
+++ b/renovate.json5
@@ -435,6 +435,12 @@
       "automerge": false
     },
     {
+      "description": "Group GitHub Actions only, excluding workflow images and runners",
+      "matchManagers": ["github-actions"],
+      "matchDepTypes": ["action"],
+      "groupName": "github-actions"
+    },
+    {
       "description": "Group Flux controller images and CLI updates",
       "matchDatasources": ["docker", "github-releases", "helm"],
       "matchPackageNames": [
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -41,9 +41,9 @@ jobs:
       - name: Build
         run: uv run python tools/assemble_docs.py && uv run mkdocs build --strict
       - name: Configure Pages
-        uses: actions/configure-pages@v5
+        uses: actions/configure-pages@v6.0.0
       - name: Upload artifact
-        uses: actions/upload-pages-artifact@v4
+        uses: actions/upload-pages-artifact@v5.0.0
         with:
           path: build/site
 
@@ -58,4 +58,4 @@ jobs:
     steps:
       - name: Deploy
         id: deploy
-        uses: actions/deploy-pages@v4
+        uses: actions/deploy-pages@v5.0.1
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -67,6 +67,9 @@ jobs:
       - name: Install Renovate
         run: npm install --global renovate@44.69.8
 
+      - name: Verify only actions join the GitHub Actions group
+        run: python3 tests/test-renovate-actions-grouping.py
+
       - name: Verify airgap digest pinning
         env:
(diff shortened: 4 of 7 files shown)


renovate.json5
```diff
@@ -434,6 +434,11 @@
       "matchUpdateTypes": ["major", "minor", "patch", "digest", "pin", "pinDigest"],
       "automerge": false
     },
+    {
+      "description": "Group all GitHub Actions workflow dependencies",
```
shrinedogg: Suggestion: The grouping paragraph at docs/dependencies.md:42 lists the grouping rules ("Grouping rules keep Flux updates together, CAPI updates together, imperative chart pins with their declarative counterparts, and node-version updates separate") but does not mention the new `github-actions` group. A short clause, e.g. "and all GitHub Actions workflow updates together", would keep the docs in sync with the rule added in this PR.

shrinedogg (commented):
Verdict: looks good, one non-blocking suggestion.

Verified locally:
- `renovate-config-validator` (renovate@44.69.8, the same pin CI's validate.yml uses) passes against the PR's renovate.json5.
- Every `uses:` reference in all 7 workflow files is now pinned to an exact x.y.z version, matching the PR claim.
- The three new pins (configure-pages v6.0.0, upload-pages-artifact v5.0.0, deploy-pages v5.0.1) all exist as upstream tags and are the latest, so no version downgrades.
- `git diff --check` is clean.
- Full dry-run per the AGENTS.md procedure on node 24.20.0, renovate@44.69.8: github-actions manager extracts 96 dependencies.
- Behavior probe: I reverted the three pages pins to their pre-PR @v5/@v4 values and re-ran the dry-run; the pages actions plus the other non-major action updates collapse into a single `renovate/github-actions` branch. The group rule works as intended.
- CI: 4/4 runnable checks pass; Rendered Flux diff skips as expected for a non-manifest change.

Suggestion:
- docs/dependencies.md:42: the grouping-rules paragraph does not mention the new group (see inline comment).

Note, not an issue: the rule matches on the manager rather than action names, so every dependency extracted from workflow files, including runner and service images (ubuntu, node, registry), lands in the github-actions group, and major updates for those go to `renovate/major-github-actions` via the default separateMajorVersions. That seems consistent with the intent of one workflow-dependency group, so no change requested.

Looks good:
- All three pins are the latest upstream tags.
- Rule placement before the flux group conflicts with no other packageRule: the flux and CAPI groups match on datasource, not manager, and the airgap digest-pinning rule is scoped to airgap file paths.

renovate.json5
lmilbaum: Good catch, Joe! I narrowed the rule so only GitHub Actions are grouped together; workflow images, runners, and other dependencies are excluded. I also added a unit test to guard against future regressions and wired it into CI. The documentation and PR description will be updated to reflect this scope too. These changes are prepared locally and will be pushed to this PR.

lmilbaum: Also addressed the documentation suggestion: docs/dependencies.md now explicitly lists the GitHub Actions group and clarifies that workflow container images and runners are excluded. This documentation change is included in the local changes awaiting push. Thanks for the review!

AGENTS.md
```diff
@@ -203,7 +203,11 @@ discovered) and the digest-pinning test run in CI only (the validate.yml
 shared harness `tests/renovate_harness.py` (issue #97): a new Renovate test
 is a file list plus assertions, never a copy of the subprocess/parsing
 logic. The harness needs Node >= 24.11 (renovate's `engines` field);
-locally: `mise x node@24 -- python3 tests/test-renovate-coverage.py`. They
+locally: `mise x node@24 -- python3 tests/test-renovate-coverage.py`.
+The offline unit test `tests/test-renovate-actions-grouping.py` also runs in
+that CI job and uses the shared harness to apply Renovate’s real package-rule
```
shrinedogg, line 208: Suggestion: curly apostrophe in `Renovate's`; the rest of the repo uses straight quotes.

shrinedogg (approved):
Re-review after the force-push (narrowed rule, docs, new unit test, CI wiring). All re-verified locally (renovate@44.69.8, Node 24, `GITHUB_COM_TOKEN` set):

- The new grouping test passes 11/11 cases against the PR's config. Red-check: restoring the original broad rule (removing `matchDepTypes`) makes it fail 7 cases, confirming the guard rejects the old behavior.
- Dry-run on a tree with this PR plus merged #241: action updates land in `renovate/github-actions`, while non-action workflow deps (`astral-sh/uv`, `jdx/mise`, `node`, `registry`) keep their own branches. The narrowing works as claimed.
- All four Pages pins are exact versions and the latest upstream tags.
- CI is green; "Rendered Flux diff" skips as expected. `git diff --check` clean.

Looks good:
- The test drives the installed CLI's own rule engine (no second JS dependency) and is wired into the existing Renovate CI job.
- The docs paragraph now names the group and its exclusions.

AGENTS.md
lmilbaum: Thanks, Joe. Replaced the curly apostrophe with a straight apostrophe in AGENTS.md. git diff --check passes. The fix is prepared locally and has not been committed or pushed yet.

lmilbaum: Correction to the resolution above: Codex made the apostrophe fix locally but did not commit or push it before resolving this thread. As a result, #242 was merged without the fix. Sorry for resolving the conversation prematurely. The one-character correction is now committed and pushed in follow-up PR #259: https://github.com/polarsquad/krops/pull/259. That PR contains only this fix; mise run validate and git diff --check pass.

PR #268, opened 2026-09-13
fix(bootstrap): teardown targets and query guards (#247, #248)

### LINKED ISSUE #247: fix: bind teardown operations to the discovered Kubernetes controller host
## Problem

AWS teardown discovers the kind controller host using an explicit `kubectl --context <configured-kind-context> cluster-info`, but maps `ControllerHost::Kind` to `kubeconfig: None`. Subsequent `kubectl_cmd(None, ...)` calls use the current context, without selecting or passing the context that discovery verified.

If kind is reachable while the operator's current context points at a different cluster, teardown can suspend Flux and delete CAPI Cluster resources on that other cluster. This is a destructive target-selection bug.

Confirmed by source inspection of current main after reviewing #145. The local-talos path also uses the same discovery and `Kind => None` mapping and should be covered by the fix.

## Relevant code

- `bootstrap-rs/src/teardown.rs`: `discover_controller_host`, AWS and local-talos controller-host dispatch.
- `bootstrap-rs/src/main.rs`: `kubectl_cmd` adds a kubeconfig only when supplied; otherwise it uses the current context.

## Reproduction conditions

1. The configured kind context exists and its API is reachable.
2. The current context points at another Kubernetes cluster.
3. Run the AWS teardown path.
4. Discovery checks kind, but subsequent Kubernetes operations target the current context.

Reproduce with isolated command stubs, not a live teardown against real clusters.

## Expected behavior

Bind every Kubernetes and Helm operation to the controller host selected during discovery. Prefer passing an explicit context/kubeconfig throughout instead of relying on mutable global current-context state. Target-selection failures must abort before mutations.

## Acceptance criteria

- AWS teardown targets the discovered controller host regardless of the operator's current context.
- Audit and fix other teardown paths sharing this discovery/targeting logic, including local-talos and local-host context selection.
- All Kubernetes and Helm operations consistently use the selected target.
- An isolated regression test uses distinct current and discovered contexts and verifies all mutation commands target only the discovered host.
- Cover target-selection failure and verify no destructive commands run afterward.
- Rust fmt, clippy, and tests pass; `mise run validate` passes.

Related: #145. Keep this correctness fix separate from the CLI rename proposal #246.

### LINKED ISSUE #248: fix: preserve teardown deletion guards when Kubernetes queries fail
## Problem

Teardown can interpret failed Kubernetes queries as confirmation that workload clusters are gone, allowing management-cluster removal before deletion is actually confirmed.

Confirmed by source inspection of current main after reviewing #145:

- AWS: `delete_aws_workload_clusters` uses `capture_lossy` for initial discovery and deletion polling. The helper ignores command exit status and suppresses stderr. A failed query with empty stdout becomes an empty workload list and returns `Ok(true)`.
- Local-host: `delete_capi_workloads` treats any failed `kubectl get cluster` as absence, both before deletion and during polling. Authentication failures, permission errors, and API outages are indistinguishable from `NotFound`.

The caller can then continue cleanup and remove the controller host, orphaning workload resources or interrupting controller-driven deletion. The newly added local-talos teardown path uses similar lossy listing logic and should be audited too.

## Relevant code

- `bootstrap-rs/src/teardown.rs`: `delete_aws_workload_clusters`, `delete_capi_workloads`, and their controller-host deletion callers.
- `bootstrap-rs/src/main.rs`: `capture_lossy`.

## Reproduction conditions

Using fully isolated command stubs, let controller discovery succeed, then make a cluster query exit nonzero with empty stdout (simulate an API outage, expired credentials, or Forbidden response). The current logic treats that result as successful absence and can proceed to host deletion.

Do not exercise this failure scenario against live infrastructure.

## Expected behavior

Require positive evidence that workloads are absent before proceeding past the deletion guard:

- List operations must succeed before an empty list means no clusters remain.
- Individual lookups must distinguish confirmed absence from transport, authentication, authorization, and other errors.
- On unconfirmed state, abort or retry within a bounded timeout while preserving the controller host; never report deletion as successful.
- Preserve deliberate skip-if-absent behavior and keep any explicit force override separate from query-error handling.

## Acceptance criteria

- Initial discovery failures and polling failures cannot mark workloads as confirmed deleted.
- AWS and local-host guards preserve the management host when workload state is unknown; audit the equivalent local-talos path.
- Isolated regression tests cover failed initial queries, failures after deletion starts, successful empty lists/NotFound, and workloads that remain present.
- Tests assert that subsequent destructive cleanup/host removal does not run after an unconfirmed deletion state.
- Errors retain useful diagnostic context and return a nonzero exit when teardown aborts.
- Rust fmt, clippy, and tests pass; `mise run validate` passes.

Related: #145. Keep this correctness fix separate from the CLI rename proposal #246.

shrinedogg:
Closes #247 and closes #248.

Both are correctness bugs in the bootstrap-rs teardown path found during the
#145 review; they are fixed together because they touch the same functions
and share the `K8sTarget` seam.

**#247: teardown can target the wrong cluster.** The teardown path passed
`kubeconfig: Option<&str>` (None for kind) into the kubectl helpers, so every
kind operation silently bound to the operator's current-context cluster and
deleted there. This PR introduces a `K8sTarget` value (explicit `--context`
for kind, explicit `--kubeconfig` for self-managed) that is threaded through
all five teardown k8s functions. Destructive operations can no longer fall
back to the ambient context.

**#248: a failed query reads as "clusters gone".** The guard read workload
listings with `capture_lossy`/`run_quiet`, so a failed `kubectl get` (auth,
forbidden, API outage) returned empty output that was interpreted as "no
workloads / already deleted". Teardown would then delete the management
cluster while the workload clusters were still deprovisioning. This PR adds a
`Probe`/`ProbeOutcome` classifier that distinguishes confirmed absence
(successful empty listing, or a named lookup reporting NotFound) from a
failed query. `probe()` runs the query once and returns the stdout alongside
the outcome, so the listing is read from the same call (no double invocation,
no TOCTOU between probe and parse):

- `delete_capi_workloads_with`: a failed `get cluster` aborts with an error
  instead of skipping; the completion poll bails on an unknown state.
- `delete_aws_workload_clusters_with`: a failed discovery listing errors
  (previously read as "no workload clusters", i.e. `Ok(true)`); a failed poll
  returns `Ok(false)` so the caller keeps the management cluster intact.
  Confirmed-empty discovery and polls still return `Ok(true)`.
- local-talos hardware release: a failed listing or poll no longer releases
  the Hardware.

`HostGuard` itself was already correct; this makes the "confirmed gone"
signal true instead of accidentally true on query failure.

Verification: `cargo build` clean, `cargo test` 64 passed / 0 failed (2 new
tests for #247 asserting the explicit context/kubeconfig argv; 5 new tests
for #248: probe classification, failed CAPD get aborts, failed AWS discovery
errors, failed AWS poll returns `Ok(false)`, confirmed-empty poll returns
`Ok(true)`), `mise run validate` green.


bootstrap-rs/src/teardown.rs
```diff
@@ -93,6 +93,126 @@ pub enum ControllerHost {
...
+                o.status.success(),
+                String::from_utf8_lossy(&o.stdout).into_owned(),
+                String::from_utf8_lossy(&o.stderr).into_owned(),
+            )
+        })
+        .unwrap_or_else(|e| (false, String::new(), e.to_string()));
+    let (ok, stdout, stderr) = out;
+    if ok {
+        return if stdout.trim().is_empty() {
+            Probe { outcome: ProbeOutcome::Absent, stdout: String::new() }
```
shrinedogg: CI: the "Build, lint, and test" job is red because `cargo fmt --check` fails on this file: 13 diffs, covering the compact `Probe { ... }` literals here and at 192, hunks at 497, 2010, 2020, and the new test bodies (2279-2397). Verified locally on the PR head. Run `cargo fmt` and push; the tests and clippy in the same job already pass.

bootstrap-rs/src/teardown.rs
```diff
@@ -1464,14 +1629,12 @@ pub async fn delete_capi_providers(kubeconfig: Option<&str>, timeout_secs: u64)
...
+/// Steps 6-8: helm releases then secrets (aws path, target threaded).
+pub async fn uninstall_helm_and_secrets(cfg: &Config, target: &K8sTarget) {
     for release in ["flux", "flux-operator"] {
         println!(">>> Uninstalling {release} Helm release...");
-        let kc: Vec<&str> = match kubeconfig {
-            Some(k) => vec!["--kubeconfig", k],
-            None => vec![],
-        };
+        let prefix = target.prefix();
+        let kc: Vec<&str> = to_refs(&prefix);
```
shrinedogg, line 1735: Warning: helm rejects `--context`.

`K8sTarget::prefix()` emits `--context <name>`, and this is the only helm consumer of it. Helm's flag is `--kube-context`: verified against the pinned helm 4.2.4, `helm --context kind-mgmt status flux -n flux-system` fails with "unknown flag: --context" while `--kube-context` and `--kubeconfig` are valid.

So in the pre-pivot AWS path (host = kind) both `helm status` calls fail, `capture_lossy` returns empty, and the code reports "{release} Helm release not found - skipping": the uninstall silently never runs, under a message that claims the release is absent. It fails closed (the kind host is deleted immediately after, nothing orphans), but issue #247's acceptance criterion "all Kubernetes and Helm operations consistently use the selected target" is not met for helm+kind.

Suggest translating per tool here (kubectl `--context`, helm `--kube-context`) or adding a `helm_prefix()` next to `prefix()`.

bootstrap-rs/src/teardown.rs
```diff
@@ -1767,52 +1942,54 @@ pub async fn run_teardown(cfg: &Config, tcfg: &TeardownConfig) -> Result<()> {
...
+            ProbeOutcome::Absent => Vec::new(),
+            ProbeOutcome::Present => p
+                .stdout
+                .lines()
+                .filter_map(|l| l.trim().rsplit('/').next())
+                .map(String::from)
+                .collect(),
+            ProbeOutcome::Error => {
+                eprintln!("!   Failed to list CAPI clusters (auth, forbidden, or API error).");
+                eprintln!("!   The cluster state is UNKNOWN; not releasing the Hardware.");
```
shrinedogg: Warning: talos unknown state exits 0 with contradictory messages.

After this UNKNOWN branch the code sets `clusters = Vec::new()`, so execution falls into `clusters.is_empty()` and prints ">>> No CAPI clusters found; nothing to release", then "Teardown complete.", and `run_teardown` returns `Ok(())` so the process exits 0. The poll `Error` branch (line 2014) has the same outcome via `clusters_gone = false` and `break`.

Two problems: the tail messages contradict the abort text directly above them (an operator skimming the end sees "nothing to release" / "complete"), and issue #248's acceptance criterion "return a nonzero exit when teardown aborts" is unmet on this path while the aws path does `bail!`.

Consider returning an error in both talos `Error` branches, matching the aws path.

bootstrap-rs/src/teardown.rs
```diff
@@ -171,48 +307,44 @@ impl HostGuard {
...
+        &[
+            "get",
+            "kustomizations.kustomize.toolkit.fluxcd.io",
+            "-n",
+            "flux-system",
+            "-o",
+            "name",
+        ],
+    );
+    let names = capture_lossy(kubectl, &to_refs(&args)).await;
```
shrinedogg: Suggestion: remaining lossy reads, same #248 class.

This listing still uses `capture_lossy`, so a failed query reads as "No Flux Kustomizations found - skipping suspension" and Flux stays live mid-teardown; the completion poll then fights reconciliation and the operator gets a full timeout cycle instead of an immediate clear error. Same pattern in `delete_capi_providers` (lines 1583 and 1616): a failed listing prints "No CAPI providers to delete" and a failed poll counts zero lines and prints "All CAPI providers removed" from a failed query. Neither weakens the deletion guard (steps 5-8 run after it passes), so a follow-up issue is fine, but these are the same failure class #248 describes.

bootstrap-rs/src/teardown.rs
```diff
@@ -93,6 +93,126 @@ pub enum ControllerHost {
...
+/// captured (not inherited) so the failure text is available for the
+/// NotFound check without leaking into the user's terminal.
+async fn probe(kubectl: &str, args: &[String], single: bool) -> Probe {
+    let out = Command::new(kubectl)
+        .args(args)
+        .output()
+        .await
+        .map(|o| {
+            (
+                o.status.success(),
```
shrinedogg, line 206: Suggestion: probe has no request timeout.

`probe()` runs kubectl with no `--request-timeout` and the poll deadlines are only checked between iterations, so a single hung query (blackholed endpoint) stalls teardown indefinitely. `discover_controller_host` already uses `--request-timeout=10s`; adding it to the probe argv (larger than the 30s poll cadence) would surface a hung API server as `ProbeOutcome::Error`.

shrinedogg (commented):
Verdict: changes requested (1 CI-blocking, 2 warnings, 3 suggestions; one suggestion is docs-only with no inline anchor). Posted as a comment review because GitHub does not allow requesting changes on your own PR.

Verified locally (worktree at 6148233):

- `cargo test`: 64 passed, 0 failed; all 7 new tests run (matches the PR claim)
- `cargo clippy --all-targets`: clean
- `cargo fmt --check`: fails, 13 diffs in `src/teardown.rs`; this is the CI "Build, lint, and test" failure
- `mise run validate`: all gates pass (bootstrap.toml cross-check passes under mise python; system python3 lacks tomllib on this machine, pre-existing)
- helm 4.2.4 (pinned 4.2): rejects `--context`, accepts `--kube-context` / `--kubeconfig`

Warnings:

- `uninstall_helm_and_secrets` feeds `K8sTarget::prefix()` (`--context <name>`) to helm, whose flag is `--kube-context`. Pre-pivot, both helm steps fail, `capture_lossy` swallows it, and the code prints "Helm release not found - skipping" for releases that exist. Fails closed, but #247's "all Kubernetes and Helm operations consistently use the selected target" is not met for helm+kind.
- The local-talos UNKNOWN branches print abort text, then fall through to "No CAPI clusters found; nothing to release" / "Teardown complete." and exit 0. #248's "return a nonzero exit when teardown aborts" is unmet on this path while the aws path does `bail!`.

Suggestions:

- `suspend_flux` and `delete_capi_providers` still read listings via `capture_lossy`, so a failed query reads as "no Kustomizations / no providers" (same #248 class; does not weaken the deletion guard, fine as a follow-up issue).
- `probe()` has no `--request-timeout`; a blackholed API endpoint stalls teardown indefinitely even though the poll has a deadline.
- docs/bootstrap-cli.md's teardown section should describe the new query-failure semantics (failed query = unknown state = abort with nonzero exit, management cluster kept intact).

Looks good:

- The `K8sTarget` seam closes #247 for kubectl: every teardown mutation binds an explicit `--context` or `--kubeconfig`; no `kubectl_cmd(None, ...)` mutation path remains.
- `probe()` classifies and returns stdout from one invocation: no double query, no TOCTOU between probe and parse.
- The `Ok(false)` poll contract matches main's existing timeout path and the caller's `bail!`.
- Tests are behavioral (argv-recording stubs, scenario scripts); the two #247 tests assert the exact prefix flags.

shrinedogg (commented):
Verdict: all findings from both review rounds are fixed and verified; nothing blocking remains.

Verified locally at b74885c (fresh worktree): cargo fmt --check clean; cargo clippy --all-targets -D warnings clean; cargo test 69 passed / 0 failed; mise validate gates pass; all 8 CI checks green.

Round 1 (fmt/CI, helm --kube-context, talos nonzero exits, probed listings in suspend_flux and delete_capi_providers, probe --request-timeout=30s, docs query-failure semantics) and round 2 (absent-CRD reads as confirmed absence, live-verified; steps 5-8 skip when the sweep removed the management host) are each covered by a regression test or a live probe.

Non-blocking: discovery's get-contexts remains lossy (pre-existing, routes to the documented AWS-only path); the mgmt_swept condition mirrors the sweep's inclusion condition and should be kept in sync if either changes.

Issue #142, opened 2026-09-01
Cover Cluster topology versions and airgap component pins in Renovate


shrinedogg:
## Problem

PR #117/#124 bumped `kindest/node` to v1.37.0 in the local-host ClusterClasses, but the Cluster `topology.version` pins stayed at v1.35.0 and broke both local clusters (fixed in #141). Root cause is a coverage gap in `renovate.json5`:

- The `kindest/node` customManager (lines 266-276) matches only strings carrying the `kindest/node:` prefix. The `version: vX.Y.Z` topology pins in `mgmt/local-host/clusters/docker/cluster.yaml` and `mgmt/local-host/clusters/management/cluster.yaml` match nothing, so image and version drift apart silently.
- The `airgap/images.txt` manager (lines 278-287) tracks latest upstream docker tags, not what kubeadm actually deploys. Current drift, crictl-verified against the live v1.37.0 nodes: etcd pinned 3.7.1-0 vs deployed 3.7.0-0, [M] coredns v1.14.7 vs deployed v1.14.6, [M] kube-proxy v1.36.1 matching the stale v1.36.1 bootstrap node pending its recreation.
- `registry.k8s.io/pause:3.10.1` in `airgap/scripts/build-package.sh` (WORKLOAD_IMAGES) and `airgap/scripts/build-config-artifact.sh` (preLoadImages) is a v1.35.0-generation leftover: v1.37.0 kubeadm pulls pause:3.10.2, and no live v1.37.0 node holds 3.10.1.

## Goal

Every pin that must move with a `kindest/node` bump moves in the same Renovate PR, and the airgap inventory reflects what the nodes actually pull rather than latest upstream tags.

## Proposed scope

1. Extend the `kindest/node` customManager (or add a sibling regex manager) to cover `topology.version` in `mgmt/local-host/clusters/*/cluster.yaml` so image and version bump together.
2. Remove the vestigial pause:3.10.1 preload from `build-package.sh` and `build-config-artifact.sh`; confirm an offline run still reaches Cluster Available (v1.37.0 kubeadm requires 3.10.2, already inventoried).
3. Re-harvest `airgap/images.txt` [M] once the kind bootstrap node is recreated at v1.37.0 (KIND_NODE_IMAGE default in `airgap/scripts/stage-and-create-cluster.sh`); correct the etcd and coredns pins to the deployed versions.
[...]

PR #316, opened 2026-09-16
feat(renovate): group all Kubernetes-release-version pins into one PR

lmilbaum:
## Summary
- `kindest/node` (local-host node image + both `topology.version` pins, already grouped per #142) and `kubernetes/kubernetes` (kubectl in `mise.toml`, and the local-talos `TalosControlPlane.spec.version` annotation) both carry a literal Kubernetes release version rather than an independent versioning scheme.
- Replaces the old "Kubernetes node/cluster versions stay independent of each other" packageRule with one that groups both depNames (plus the seven registry.k8s.io kubeadm component depNames in images.txt) under `groupName: "kubernetes-version"`, so a Kubernetes version bump opens a single PR across every file that tracks it instead of kubectl/local-talos drifting in separate PRs.
- Adds `tests/test-renovate-kubernetes-version-grouping.py`, mirroring the existing `test-renovate-actions-grouping.py` pattern, asserting `kindest/node` and `kubernetes/kubernetes` join the group while `kubernetes-sigs/kind`, `kubernetes-sigs/cluster-api`, and `siderolabs/talos` do not.

Splitting this config into a separate file (to keep it grouped with related managers) was investigated but isn't currently possible without breaking this repo's offline `--platform=local` test harness — see [renovatebot/renovate#40970](https://github.com/renovatebot/renovate/discussions/40970#discussioncomment) for the upstream limitation.

## Test plan
- [x] `PATH="$(mise where node@24)/bin:$PATH" python3 tests/test-renovate-kubernetes-version-grouping.py -v`
- [x] `PATH="$(mise where node@24)/bin:$PATH" python3 tests/test-renovate-actions-grouping.py -v`
- [x] `PATH="$(mise where node@24)/bin:$PATH" python3 tests/test-renovate-coverage.py`
- [x] Verified end-to-end that `topology.version` in both local-host `cluster.yaml` files resolves to `depName: kindest/node` and lands in `groupName: kubernetes-version`


tests/test-renovate-kubernetes-version-grouping.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
```
shrinedogg, line 1: Warning: this test is not wired into CI or mise run validate. The renovate job in .github/workflows/validate.yml runs test-renovate-actions-grouping.py, airgap/tests/test-renovate-digest-pinning.py, tests/test-renovate-coverage.py, and tests/test-renovate-release-assets.py; grep finds no reference to kubernetes-version-grouping anywhere in .github/ or the mise configs. As pushed, the guard only runs on the author's machine, so a future edit that drops a depName or renames the group stays green in CI. Verified locally: it passes offline under Node 24 with renovate 44.50.1, same shape as the actions-grouping test, so a plain run step next to it needs no extra env. Suggest adding `python3 tests/test-renovate-kubernetes-version-grouping.py` to the renovate-digest-pinning job, and to mise run validate if that is where the offline tests belong.

docs/dependencies.md
```diff
@@ -45,11 +45,29 @@ Renovate discovers and updates versions in:
...
+The `kubernetes-version` group (#142) covers `kindest/node` (docker datasource:
+the local-host node image and both Cluster `topology.version` pins),
+`kubernetes/kubernetes` (github-releases datasource: the `kubectl` pin in
+`mise.toml` and the local-talos `TalosControlPlane.spec.version` annotation),
+and the `airgap/images.txt` images kubeadm itself deploys for that release
+(`kube-apiserver`, `kube-controller-manager`, `kube-proxy`, `kube-scheduler`,
+`coredns/coredns`, `etcd`, `pause`, all under the `registry.k8s.io` docker
+datasource). All of these carry a literal Kubernetes release version rather
+than an independent versioning scheme, so a Kubernetes bump opens a single PR
+across every file that tracks it, addressing #142's items 3-4 (the
```
shrinedogg: Warning: "addressing #142's items 3-4" overstates what grouping does, and the alternative item 4 explicitly offered ("document a re-harvest step in the kindest/node bump flow") was not added either. Grouping puts the images.txt component pins in the same PR, but Renovate still resolves each to its latest upstream tag, not the tag kubeadm deploys for the tracked release. Issue #142's own crictl data shows the drift this leaves behind (etcd pinned 3.7.1-0 vs deployed 3.7.0-0, coredns v1.14.7 vs deployed v1.14.6), and images.txt still carries those values unchanged, plus the [M] pause:3.10 vs [W] pause:3.10.2 split. A kubernetes-version PR will therefore propose wrong-by-default etcd/coredns tags unless the reviewer hand-corrects them. That is consistent with the repo's "Renovate opens, human completes" posture (KCC operator bundle, Talos images), but then the docs should say so: reword this sentence to state grouping gives same-PR visibility only, and add a step to the Update procedure requiring a crictl re-harvest/reconciliation of the images.txt component tags when a kubernetes-version PR opens.

shrinedogg (changes requested):
Request changes. The grouping rule itself is verified sound end to end; two gaps should land before merge: the new test is not wired into CI, and the docs overclaim #142 coverage while omitting the re-harvest step #142 asked for as the alternative.

Verified locally (worktree at PR head 95d5442):
- `python3 tests/test-renovate-kubernetes-version-grouping.py -v`: OK
- `python3 tests/test-renovate-actions-grouping.py -v`: OK
- `python3 tests/test-renovate-coverage.py` (tokened): OK, all managed pins detected
- Full dry-run per AGENTS.md (renovate 44.50.1, Node 24.20): exit 0, "Dependency extraction complete", zero "Failed to look up", empty skipReason histogram
- End-to-end grouping check: extracted 91 deps from images.txt, the three airgap scripts, mise.toml, and the three cluster.yaml files via the shared harness, then applied the repo's real packageRules with Renovate's applyPackageRules. All kindest/node occurrences (including both topology.version pins and the script pins), both kubernetes/kubernetes pins, and the seven registry.k8s.io component depNames join kubernetes-version. docker.io/kindest/{kindnetd,local-path-helper,local-path-provisioner}, kindest/haproxy, and kubernetes-sigs/kind stay ungrouped; kubernetes-sigs/cluster-api stays in cluster-api. No unintended members.
- `gh pr checks`: all pass; "Rendered Flux diff" skipping is expected for a config/docs-only diff.

Warnings:
1. New test not wired into CI or mise run validate (inline).
2. Docs overclaim #142 items 3-4; component tags still resolve to latest upstream, the existing etcd/coredns drift and pause tag split remain, and the documented re-harvest step is missing (inline).

Suggestions:
- The PR description says the rule groups "both depNames"; the rule lists nine (the seven registry.k8s.io components are also in scope). Updating the description would help reviewers see the full blast radius.

Looks good:
- Exact matchDepNames matching rather than registry-host matching is the right boundary, and the negative same-registry CAPI-image case guards it.
- Dropping the old kindest/node isolation rule is consistent with repo-level separateMajorMinor=false and the documented one-PR-at-newest policy.
- The pointer comment resolves to a real anchor (docs/dependencies.md#managed-surfaces).
- The "Intentional differences" rewrite is accurate: EKS versions are not Renovate-managed.

tests/test-renovate-kubernetes-version-grouping.py
lmilbaum: Fixed in 462d39c: added a run step for this test next to the actions-grouping test in the renovate job of validate.yml. Left mise run validate alone since the actions-grouping test isn't wired there either.

docs/dependencies.md
lmilbaum: Fixed in 462d39c: reworded to say grouping gives same-PR visibility only (etcd/coredns/pause can still drift), and added an update-procedure step to re-harvest with crictl and reconcile the images.txt component tags before merging a kubernetes-version PR.

shrinedogg (approved):
Re-review at 462d39c: both findings addressed.

Verified locally: new grouping test, actions-grouping test, and coverage test all pass (Node 24.20.0, renovate 44.90.3); the CI renovate job ran the new step on this head and is green.

- CI wiring: new step sits next to the actions-grouping test in the renovate job; confirmed running in CI.
- Docs: Managed surfaces now scopes the group to same-PR visibility with the drift caveat, the update procedure covers the re-harvest/reconcile step, and Intentional differences correctly drops the CAPD node image (now managed) and the EKS independence claim.

Looks good: exact depName matching keeps the CAPI/kubeadm provider images on registry.k8s.io out of the group, and the negative test case guards it.

shrinedogg (approved):
LGTM

Issue #72, opened 2026-08-26
Support for GCP resources via Config Connector


shrinedogg:
## Problem

knr-ops currently provides cloud resource provisioning and cluster lifecycle management for AWS and local-host profiles. Users deploying onto GCP have no first-class path: there are no GCP provider manifests, no managed resource definitions, and no documented workflow for reconciling Google Cloud resources through Flux the way the AWS profile does today.

## Goal

Add a GCP profile that provisions and manages Google Cloud resources using [Config Connector](https://docs.cloud.google.com/config-connector/docs/overview), driven by the same GitOps/Flux patterns as the existing profiles.

## Proposed scope

- Install and bootstrap Config Connector (with its required Workload Identity setup) in the management or workload cluster as part of the GCP profile.
- Define CNRM custom resources for the GCP equivalents of what the AWS profile provisions: projects/service enablement, networking, IAM/service accounts, and any workload-adjacent resources.
- Wire CNRM resources into the existing Flux/Kustomize layout with an environment overlay structure matching the current profiles.
- Integrate with knr-ops conventions where applicable:
  - dependency versions declared in their native files and discovered by Renovate (#74): new manifests under mgmt/ and workload/ are covered by the existing managerFilePatterns, and any nonstandard pin gets a custom manager plus a `tests/test-renovate-coverage.py` entry,
  - airgap packaging considerations for Config Connector images and CRDs where feasible,
  - `mise run validate` and the CI workflows (lint, kustomize build, rendered Flux diff, airgap inventory check) covering the new manifest paths.
- Document prerequisites (GCP project, Workload Identity federation, credentials) and the update procedure.

## Acceptance criteria

- A GCP profile exists alongside AWS and local-host and is selectable via the standard setup flow.
- GCP resources reconcile through Flux via Config Connector custom resources with no kubectl apply outside GitOps.
- Credentials are handled consistent with existing secret management (Workload Identity preferred, SOPS for remaining secrets).
- Validation covers the new manifests and passes in CI.
- Documentation explains the GCP setup path end to end.

## References

- Origin request: #70
- Related precedent: #74 (Renovate with native version declarations; superseded the central catalog from #62)

shrinedogg:
Dependency note (from the milestone planning pass):

1. #74 must land first: same rationale as #71, the scope references the central version catalog that #74 retires in favor of Renovate-managed native files.
[...]

PR #317, opened 2026-09-16
feat: add the gcp environment (CAPG + Config Connector management side)

shrinedogg:
## What

Adds the `gcp` environment's management side: `mgmt/gcp/` mirrors `mgmt/azure/` with CAPG v1.13.1 (GKE) replacing CAPZ and Config Connector 1.156.0 (pinned operator bundle) replacing the ASO bundle.

- `mgmt/gcp/`: gcp-vars, cert-manager, CAPI operator, CAPI core + CAPG providers, pinned KCC operator, KCC instance + WIF identity, GKE management + staging clusters (europe-north1), flux-apps addons for the workload cluster
- Zero secrets at rest: CAPG and management-side KCC authenticate through Workload Identity Federation (pool `krops`, providers `kind` + `mgmt`; no service-account keys exist)
- `mise.gcp.toml`: gcloud 585.0.0, `gcp-bootstrap` / `wif-federate` / `kubeconfigs`
- `bootstrap-rs`: one new config-driven knob, `pivot-manifest-vars`, applied before `${VAR}` substitution, plus `${VAR:=default}` support (no profile-name branching)
- `bootstrap.toml` `[environments.gcp]` with a manual teardown (the acceptance run establishes the orphan set; automation is a follow-up)
- toolbox image gains the gcp layer (gcloud); `toolbox-run.sh` passes `GCP_PROJECT`/`GCP_REGION`/`CLOUDSDK_CONFIG` through
- Renovate: CAPG provider version, gcloud pin, KCC operator version (annotated comment; the pin test turns bump PRs red until the ~3500-line bundle is replaced by hand)
- docs: `docs/gcp.md`, `extending.md` section, architecture diagrams, README, mkdocs nav, AGENTS.md

The workload side (`workload/gcp-base/`) is the stacked PR 2.

Refs #72

## Verification

- `mise run validate` green on this branch (96 kustomize builds, bootstrap cross-check, KCC pin, identity chains, airgap invariants)
- `cargo test -p krops-bootstrap`: 73 passed
- renovate coverage dry-run: all new surfaces resolve (CAPG github-releases, gcloud via docker datasource, KCC matchStrings)
- toolbox image build verified (gcloud 585.0.0 installs; the gcloud installer's provisioning probe needs an apt `python` on bookworm-slim, added to the image)
- konflate diff: new CRDs/providers only under `mgmt/gcp/`, zero diff under `mgmt/aws|azure` (verified with the git diff: no files outside the listed surfaces change)

## Known limitations (documented in docs/gcp.md)

- No live acceptance run yet; `GCP_PROJECT_NUMBER` and the GKE version are placeholders until `gcp-bootstrap` runs against a real project
- Zonal clusters (europe-north1-a); Arm pool omitted until machine-type availability is confirmed (spike S4, acceptance)
- Teardown is manual until the orphan set is established


shrinedogg (commented):
## Code Review Summary

**Verdict: Comment** (2 issues to fix before the first live run, 2 suggestions)

All CI checks pass. The diff is scoped exactly as described: new CRDs and providers only under `mgmt/gcp/`, zero render diff under `mgmt/aws|azure`.

Verified locally:
- `cargo test -p krops-bootstrap`: 73 passed
- kcc operator pin, bootstrap.toml cross-check, renovate coverage (Node 24 with tokens), `bash -n scripts/toolbox-run.sh`: all pass; `mise run validate`: 88/88 kustomize builds green
- Version pins resolve upstream: CAPG v1.13.1, KCC v1.156.0 (github-releases), gcloud 585.0.0 (gcr.io tag)
- `spec.patches` on InfrastructureProvider v1alpha2 exists in cluster-api-operator (generic provider spec, strategic merge supported), so the WIF volume swap is well-formed
- `e30=` decodes to `{}` as documented; azure counterparts match (flux-operator 0.58.0, capi-operator chart 0.28.0, cert-manager 1.21.1)

### Warning
- **mise.gcp.toml:80** - the `sed` pattern that extracts the kind issuer has doubled backslashes in the committed bytes (`\\(` instead of `\(`), so it matches nothing; `wif-federate` always dies at the `ISSUER` guard
- **mise.gcp.toml:112** - same problem in `kubeconfigs`: `-exec basename {} \\;` never terminates the command, so the loop is a bash syntax error

### Suggestion
- **bootstrap-rs/Dockerfile** is absent despite the PR body and commit 0510b82 describing the gcloud layer as committed and build-verified; `gcp-bootstrap` and `wif-federate` fail in the toolbox until the layer lands
- `scripts/toolbox-run.sh:163` appends `-e CLOUDSDK_CONFIG=...` after `"${PASS_ENV[@]}"`, so it silently overrides an operator-provided value; and the literal `krops-placeholder` appears in two files (`gcp-vars.yaml`, `flux-instance.yaml`), matching the azure precedent but worth a one-line comment pointing at each other

The WIF design itself (pool `krops`, per-surface providers, `${GCP_WIF_PROVIDER:=mgmt}` with the pivot override) reads clean and the ordering test guards the right contract.

mise.gcp.toml
```diff
@@ -0,0 +1,129 @@
...
+: "${GCP_PROJECT:?set GCP_PROJECT in .env or the shell}"
+KIND_CONTEXT="${KIND_CONTEXT:-kind-mgmt}"
+kubectl config use-context "$KIND_CONTEXT"
+gcloud config set project "$GCP_PROJECT" --quiet
+if ! gcloud auth print-access-token >/dev/null 2>&1; then
+  echo "==> no gcloud session; starting browser-less login"
+  gcloud auth login --no-launch-browser --quiet
+fi
+PROJECT_NUMBER=$(gcloud projects describe "$GCP_PROJECT" --format='value(projectNumber)')
+ISSUER=$(kubectl get --raw /.well-known/openid-configuration | sed -n 's/.*"issuer":"\\([^"]*\\)".*/\\1/p')
```
shrinedogg: Warning: the committed bytes have doubled backslashes in the sed pattern (`\\(`, `\\)`; verified at byte level, not from the diff view). This is a TOML literal string, so bash receives `\\(` and sed matches nothing: `ISSUER` ends up empty and the task always dies at the `: "${ISSUER:?...}"` guard. Reproduced with the exact extracted bytes: the single-backslash pattern captures the issuer, this one returns empty with exit 0. Fix with single backslashes, or sidestep BRE groups: `ISSUER=$(kubectl get --raw /.well-known/openid-configuration | jq -r .issuer)` (jq ships in the toolbox image).

mise.gcp.toml
```diff
@@ -0,0 +1,129 @@
...
+kubectl -n flux-system create configmap gcp-wif --from-literal=GCP_WIF_PROVIDER=kind \
+  --dry-run=client -o yaml | kubectl apply -f -
+'''
+
+[tasks.kubeconfigs]
+description = "Export the GKE workload-cluster user kubeconfig (clusterctl 1.14 has no --user flag, so read the <cluster>-user-kubeconfig Secret CAPG writes; falls back to the control-plane kubeconfig)"
+run = '''
+KUBECONFIG_FILE="${KUBECONFIG_FILE:-$HOME/.kube/krops-workloads.yaml}"
+fail=0
+for region in $(find mgmt/gcp/clusters -mindepth 1 -maxdepth 1 -type d -exec basename {} \\; | sort); do
```
shrinedogg: Warning: same doubled-backslash problem. The committed bytes are `-exec basename {} \\;`, which never terminates `-exec`, and the loop is a bash syntax error at parse time (`syntax error near unexpected token`, reproduced locally with the exact bytes). A tested alternative: `find mgmt/gcp/clusters -mindepth 1 -maxdepth 1 -type d | while read -r d; do basename "$d"; done | sort`.

scripts/toolbox-run.sh
```diff
@@ -157,5 +160,6 @@ exec "$CONTAINER_ENGINE" run --rm ${TTY_ARGS[@]+"${TTY_ARGS[@]}"} \
   -v "$REPO_ROOT/.kube:/root/.kube" \
   -e KUBECONFIG="$KUBECONFIG_IN" \
   "${PASS_ENV[@]}" \
+  -e CLOUDSDK_CONFIG=/workspace/.gcloud \
```
shrinedogg, line 169: Suggestion: this `-e CLOUDSDK_CONFIG=...` is appended after `"${PASS_ENV[@]}"`, so the container engine takes the last value and silently overrides an operator-provided `CLOUDSDK_CONFIG` (the PASS_ENV entry above never wins). Fine if canonical-by-design; a one-line comment saying so would prevent confusion.

mgmt/gcp/infrastructure/gcp-vars/gcp-vars.yaml
```diff
@@ -0,0 +1,20 @@
...
+# (${GCP_WIF_PROVIDER:=mgmt}); the kind bootstrap cluster overrides it to
+# `kind` through the gcp-wif ConfigMap the wif-federate task creates.
+# ══════════════════════════════════════════════════════════════════════════════
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: gcp-vars
+  namespace: flux-system
+data:
+  GCP_PROJECT: "krops-placeholder"
```
shrinedogg, line 21: Suggestion: the `krops-placeholder` / `000000000000` literals are committed here and again in `mgmt/gcp/addons/flux-apps/flux-instance.yaml` (`cluster-vars`). Matches the azure precedent, but the two must be updated together after `gcp-bootstrap` runs; a pointer in each file to the other (mirroring the docs' "Commit the identifiers" list) would help the next editor.

mise.gcp.toml
shrinedogg: Fixed in 6f4d2ac. Confirmed the root cause at the byte level: both `run` strings here are TOML literal strings, so the committed bytes reach bash verbatim — and these two lines had doubled backslashes where the working `mise.aws.toml` / `mise.azure.toml` counterparts have a single one. I reproduced both with the exact committed bytes: the single-backslash sed captures the issuer, this one returned empty with exit 0 (dying at the `ISSUER` guard); the single-backslash find lists the regions, this one hit the `|` syntax error and iterated zero. Both lines are now single-backslash and verified against a mock `kubectl` and a real cluster tree.

mise.gcp.toml
shrinedogg: Fixed in 6f4d2ac, same doubled-backslash root cause as the sed line above. The committed `\\;` never terminates `-exec`, so the command substitution is a bash syntax error and the loop body runs over nothing. Now `find mgmt/gcp/clusters -mindepth 1 -maxdepth 1 -type d -exec basename {} \; | sort`, matching the `aws`/`azure` tasks; verified it enumerates the region directories against a real tree.

scripts/toolbox-run.sh
shrinedogg: Agreed it is canonical-by-design: the container must always use the repo-local `.gcloud/` (inside the `/workspace` mount) so the host `mise -E gcp` env and the toolbox share one gcloud session. Added a comment above the line stating that the later `-e` deliberately wins over an operator-set `CLOUDSDK_CONFIG` forwarded in `PASS_ENV`, and to keep the two in sync if the mount path ever changes (819755b).

mgmt/gcp/infrastructure/gcp-vars/gcp-vars.yaml
shrinedogg: Added a pointer in each file to the other and to the docs "Commit the identifiers" list, so `GCP_PROJECT` / `GCP_PROJECT_NUMBER` are updated together after `gcp-bootstrap` runs (819755b).

shrinedogg:
Addressed all four points on top of `edd0c51` (three additive commits, CI running on the new head).

**Warning — doubled backslashes (`mise.gcp.toml:80` and `:112`)** — fixed in `6f4d2ac`. Root cause confirmed at the byte level: these `run` strings are TOML literal strings, so the committed bytes reach bash verbatim, and both lines had doubled backslashes where the working `mise.aws.toml` / `mise.azure.toml` tasks have a single one. I reproduced both with the exact committed bytes (sed returned empty at exit 0; the `find -exec` was a `|` syntax error over zero regions) and verified the single-backslash forms against a mock `kubectl` and a real cluster tree.

**Suggestion — `bootstrap-rs/Dockerfile` missing** — committed in `847c980`. You are right that the Dockerfile gcloud layer described in 0510b82 was never committed; the image now installs `gcloud` (via `mise -E gcp install -y gcloud`) plus `apt python-is-python3` for the installer's provisioning probe. Two things the byte-level work surfaced, both handled in the committed layer:

- The `gke-gcloud-auth-plugin` is **not** in the SDK tarball, so the docs' "bundles gke-gcloud-auth-plugin" claim was inaccurate. The layer now runs `gcloud components install gke-gcloud-auth-plugin` in the build, and `docs/gcp.md` was corrected to say so.
- The component install uses `mise -E gcp which gcloud` to locate the binary rather than `mise x`, because `mise x` auto-installs every tool in the env (including the dev-only zarf) and a zarf auto-install failure aborts the build.

Verified by a full `bootstrap-rs/Dockerfile` build on **both** `linux/arm64` and `linux/amd64` (the CI docker job is arm64): `gcloud --version` reports 585.0.0 and the plugin binary + shims are present in both images.

**Suggestion — `CLOUDSDK_CONFIG` override (`toolbox-run.sh:163`)** — commented in `819755b`. It is canonical-by-design: the container must always use the repo-local `.gcloud/` (inside the `/workspace` mount) so the host `mise -E gcp` env and the toolbox share one session. The comment states the later `-e` deliberately wins over an operator value in `PASS_ENV` and to keep the two in sync.

**Suggestion — placeholder cross-pointers** — added in `819755b`: each of `gcp-vars.yaml` and `flux-apps/flux-instance.yaml` now points at the other and at the docs "Commit the identifiers" list.

PR #321, opened 2026-09-16
ci(air-gapped): file/update a tracking issue when the scheduled run fails

lmilbaum:
## Summary
- The daily `air-gapped` workflow (`cron: "17 2 * * *"`) has been failing every day for at least the last 10 days, and nothing surfaces that beyond the Actions tab.
- Adds a `notify-on-failure` job that runs only when a *scheduled* run fails (not manual `workflow_dispatch`), using the built-in `GITHUB_TOKEN` (`issues: write`) to either open a new labeled issue (`bug`, `area:airgap`, `area:ci`, `priority:high`) or comment on the existing one, linking to the failed run — no new secrets needed.
- Context: this surfaced while following up on #142 — the daily job is designed to verify the offline deployment reaches `Cluster` `Available` with egress blocked, which is exactly what #142's acceptance criteria needed, but it's currently failing earlier, at a `cert-manager` Helm install timeout, unrelated to #142's `pause:3.10.1` removal. This PR only adds the notification; the underlying cert-manager timeout is tracked and fixed separately in #322/#332.
- Adds `airgap/tests/test-airgap-failure-notification.py`: a structural regression test guarding the job's trigger condition, permissions, and dependency on `build`/`isolated-deploy`, since its actual logic lives inside an `actions/github-script` block nothing here can execute offline. Wired into `validate.yml` and `mise run validate`.

## Test plan
- [x] YAML validated (`ruby -ryaml -e "YAML.load_file(...)"`)
- [x] `python3 airgap/tests/test-airgap-failure-notification.py` -- OK, confirmed it fails against `main` (job doesn't exist)
- [ ] First scheduled failure after merge confirms the issue gets filed/commented on as expected


.github/workflows/validate.yml
```diff
@@ -24,6 +24,8 @@ jobs:
         run: python3 airgap/tests/test-airgap-ownership.py
       - name: Verify digest token boundary (reject overlong)
         run: python3 airgap/tests/test-digest-regex.py
+      - name: Verify failure-notification job wiring (#321)
```
shrinedogg, line 27: Suggestion: The airgap section of AGENTS.md lists the gates that run in `mise run validate` and CI for this directory (`test-airgap-image-digests.py`, `test-airgap-ownership.py`), and #273 documented its new gate in both AGENTS.md and docs/airgap.md in the same PR. Consider adding a line for this wiring gate there (or a note that it is intentionally CI-only) so the inventory stays accurate. Not blocking.

shrinedogg (approved):
The failure-notification job works as described.

Verified locally:

- The wiring test passes on the PR head and fails against main, as the test plan claims; the CI `Air-gap image digest pins` job runs the same step green. Konflate skip is expected (no Flux-managed manifests touched).
- Executed the embedded github-script block against a mocked `github`/`context`: existing issue -> `createComment` with the run URL; no match -> `create` with the title and all 4 labels (all present in the repo). Issue #322 is already open with the exact dedup title and label set, so the first scheduled failure will comment on it instead of creating a duplicate.
- `failure()` covers a `build` failure: both deploy jobs depend on `build`, and `failure()` is true when any ancestor job fails, so the notification still fires and the message stays accurate.
- `actions/github-script@v9.0.0` tag exists; job-level `permissions: issues: write` is least privilege for the step.

Suggestions (non-blocking):

- The airgap section of AGENTS.md lists the validate/CI gates for this directory and #273 documented its gate in AGENTS.md and docs/airgap.md in the same PR; consider a line for this wiring gate there (see inline comment).
- Corner case on the dedup: the issue lookup ANDs all 4 labels, so a tracking issue that keeps its title but loses a label would stop matching and a second same-title issue would be created. Unlikely given the label combination; noting for awareness.

Looks good otherwise. The one open test-plan item (live confirmation on the first scheduled failure after merge) is the author's to watch, but the logic itself is verified above.

PR #323, opened 2026-09-16
fix: downgrade Kubernetes to v1.36.4 for cert-manager compatibility

lmilbaum:
## Summary

Root cause of #322 (the daily `air-gapped` workflow failing every scheduled run since 2026-09-01): cert-manager v1.21.x — pinned repo-wide, including `airgap/zarf.yaml` and `mgmt/local-host/infrastructure/cert-manager/helmrelease.yaml` — [officially supports only Kubernetes 1.33-1.36](https://cert-manager.io/docs/releases/). The v1.37.0 Kubernetes bump (#117-#128) pushed the local-host/local-talos management clusters outside that support window on exactly the day the daily job started failing. No cert-manager release supporting 1.37 exists yet (latest is v1.21.2, still 1.33-1.36 only).

- Downgrades every pin in the `kubernetes-version` Renovate group (#142) — `kindest/node` image + digest (both `cluster-class.yaml` files), both `topology.version` pins, `kubectl` in `mise.toml`, and the local-talos `TalosControlPlane.spec.version` — to v1.36.4, using kind v0.33.0's own published node image digest (`sha256:099e04...`, from the [kind v0.33.0 release notes](https://github.com/kubernetes-sigs/kind/releases/tag/v0.33.0)).
- Refreshes `airgap/images.txt`'s kubeadm-deployed component versions (kube-apiserver/controller-manager/scheduler/proxy, coredns, etcd, pause) to kubeadm v1.36.4's own image list — obtained by running the official `kubeadm` v1.36.4 binary (`kubeadm config images list --kubernetes-version v1.36.4`) in a container, not guessed — rather than carrying forward the v1.37.0 generation's already-known coredns/etcd drift (#142 items 3-4 were never fixed for v1.37.0).
- Updates matching references in `airgap/scripts/{build-package,build-config-artifact,stage-and-create-cluster}.sh` and `docs/{airgap,operations}.md`.

## Caveats

- **Not yet verified against a live offline run.** #322 tracks the daily workflow's current failure; this PR should turn it green once merged, but that's the actual verification, not this diff alone.
- The `airgap/images.txt` k8s-component digests come from kubeadm's own compiled image list, not a live `crictl` harvest against a running v1.36.4 node — matching the caveat #142 already raised for the v1.37.0 generation. Worth a real harvest once an operator has a live v1.36.4 environment.
[...]

PR #325, opened 2026-09-16
test(airgap): verify images.txt k8s component pins against real kubeadm

lmilbaum:
## Summary

`airgap/images.txt`'s `kube-apiserver`/`kube-controller-manager`/`kube-proxy`/`kube-scheduler`, `coredns`, `etcd`, and `pause` pins are meant to track what kubeadm actually deploys for the pinned `kindest/node` version, not just "latest upstream tag" — `docs/dependencies.md` has said *"there is no automated completeness check between manifests and the inventory, so verify the pairing during review"* ever since #142 was filed, and items 3-4 of that issue were never closed.

This adds `airgap/tests/test-airgap-kubeadm-images.py`:
- Reads the target Kubernetes version from the `kindest/node` `topology.version` pin.
- Downloads and runs the **real** `kubeadm` binary for that exact version (`kubeadm config images list --kubernetes-version vX.Y.Z`) — in a container on non-Linux hosts, since kubeadm has no macOS build; native on Linux CI runners.
- Fails if any `images.txt` pin disagrees with kubeadm's own authoritative list.

Wired into `validate.yml`'s `airgap-image-pins` job (needs real network access to `dl.k8s.io`, so it's CI-only, matching how the existing Renovate tests are kept out of `mise run validate` for the same reason — they need Node 24 + network too).

## Verification

I ran this against two known states rather than trusting it blind:
- **main's current v1.37.0 pins**: fails with exactly the drift #142 already documented — `coredns:v1.14.7` vs kubeadm's actual `v1.14.6`, `etcd:3.7.1-0` vs kubeadm's actual `3.7.0-0`, plus a `pause:3.10` vs `3.10.2` inconsistency between images.txt's two sections.
- **PR #323's v1.36.4 downgrade** (kubeadm-verified pins): passes cleanly.

## Test plan
- [x] `python3 airgap/tests/test-airgap-kubeadm-images.py` against main — fails with the expected, documented drift
- [x] Same script against PR #323's branch — passes
- [x] `mise run validate` — unaffected (this check intentionally isn't part of it)
- [x] `ruby -ryaml -e "YAML.load_file('.github/workflows/validate.yml')"` — valid

Refs: #142


airgap/tests/test-airgap-kubeadm-images.py
```diff
@@ -0,0 +1,159 @@
...
+"""
+
+import platform
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+NODE_IMAGE_SOURCE = REPO_ROOT / "mgmt/local-host/clusters/docker/cluster.yaml"
```
shrinedogg, line 25: Suggestion: the version source covers only the workload cluster. `clusters/management/cluster.yaml` carries an identical `topology.version` pin for the cluster the [M] section images actually serve. Today both are v1.37.0 and Renovate bumps them in one grouped PR (#270), but nothing asserts they agree; if they ever diverge, the [M] pins get validated against the wrong kubeadm version. Reading both files and failing on mismatch would turn the grouping assumption into an invariant.

.github/workflows/validate.yml
```diff
@@ -24,6 +24,8 @@ jobs:
         run: python3 airgap/tests/test-airgap-ownership.py
       - name: Verify digest token boundary (reject overlong)
         run: python3 airgap/tests/test-digest-regex.py
+      - name: Verify images.txt k8s component pins match kubeadm (#142)
+        run: python3 airgap/tests/test-airgap-kubeadm-images.py
```
shrinedogg, line 30: Suggestion: AGENTS.md's airgap section enumerates the repo gates and their mise/CI split (ownership test, digest gate, the CI-only Renovate gates) but doesn't mention the new CI-only kubeadm gate. Per this repo's own AGENTS.md rule (update it in the same PR for test-procedure changes), a one-line addition alongside the ownership-test sentence would match the established pattern.

shrinedogg (approved):
Verified locally before approving:

- `python3 airgap/tests/test-airgap-kubeadm-images.py` on this branch: passes (7 components checked against kubeadm v1.37.0), exercising the macOS docker path with the real kubeadm binary from dl.k8s.io.
- Failure path: ran the same test against a fresh origin/main worktree: fails with exactly the drift the PR describes (coredns v1.14.7 vs v1.14.6, etcd 3.7.1-0 vs 3.7.0-0, pause 3.10 vs 3.10.2), exit 1. Matches issue #142's crictl-verified harvest.
- New digests verified against the live registry (the repo's digest gate does no lookups): coredns:v1.14.6 -> 900f9c10..., etcd:3.7.0-0 -> 6ecefbe2..., pause:3.10.2 -> f548e0e8... All match the pins here.
- `test-airgap-image-digests.py` and `test-airgap-ownership.py` pass on the branch; the new CI step ran green on this head SHA; `mise run validate` deliberately excludes it, consistent with the other network-only gates.
- Cross-references check out: #326 and renovatebot/renovate#46270 exist; the rewritten images.txt NOTE matches cluster-class.yaml, KIND_NODE_IMAGE, and build-package.sh (all three node types on kindest/node v1.37.0).

Two non-blocking suggestions inline.

Looks good:

- The gate checks every occurrence of each component, so the old [M] vs [W] pause inconsistency (3.10 vs 3.10.2) is structurally caught, not just this instance.
- Failure output names images.txt:<line> with kubeadm's expected tag; download failures surface loudly via set -eu + CalledProcessError.
- The docs paragraph on the postUpgradeTasks fallback is accurate, honest about the residual gap (Renovate can still propose a stale tag), and properly parked in #326.

airgap/tests/test-airgap-kubeadm-images.py
lmilbaum: Done in the latest push.

.github/workflows/validate.yml
lmilbaum: Done in the latest push.

Issue #322, opened 2026-09-16
air-gapped: scheduled workflow is failing


lmilbaum:
## Problem

The daily `air-gapped` workflow (`.github/workflows/air-gapped.yml`, `cron: "17 2 * * *"`) has failed every scheduled run for at least the last 10 days (2026-09-07 through 2026-09-16 checked).

Both the `isolated-deploy` ("Deploy without external egress") and `monitored-deploy` ("Deploy while monitoring public traffic") jobs fail at the same step: the `cert-manager` Helm install times out mid-deploy.

Latest failure (run [35048485941](https://github.com/polarsquad/krops/actions/runs/35048485941), 2026-09-16):

```
2026-09-16 02:41:14 INF performing Helm install chart=cert-manager
2026-09-16 02:56:14 ERR failed to deploy package: unable to deploy component "cert-manager":
                        unable to install chart context deadline exceeded: if you need to remove the failed chart, use `zarf package remove`
```

The `build` job (bundling the air-gap package) succeeds; the failure is specifically in the offline deploy, at the cert-manager component, roughly 15 minutes after the Helm install starts.

## Impact

`airgap/scripts/offline-run.sh` never gets far enough to reach step 8 ("verify CAPD workload cluster provisions offline" / `Cluster` `Available` check), so the workflow currently provides no live signal on whether an air-gapped deployment reaches `Cluster Available`. This is also why #142's acceptance criterion ("confirm an offline run still reaches Cluster Available") could not be verified after the `pause:3.10.1` removal in #270 — the daily job that was meant to catch this has been red since before and after that change, for an apparently unrelated reason.

## Scope

- Determine why the `cert-manager` Helm install times out in the isolated/monitored deploy jobs (resource constraints on the `ubuntu-24.04-arm` runner? a missing/incorrect wait condition? an image pull issue despite images being pre-pushed to the local registry?).
- Fix it so the daily workflow reaches a green run again.
- Once green, use it (or a manual `workflow_dispatch` run) to confirm the offline deployment reaches `Cluster` `Available`, closing the loop on #142's outstanding verification step.

## References

- #142 (origin: the acceptance criterion this blocks)
- #270 (the pause:3.10.1 removal whose live verification is blocked by this)
- PR #321 (adds failure notification for this workflow, separate from fixing the underlying failure)

lmilbaum:
## Update: the v1.36.4 downgrade does not fix this

[...]

PR #327, opened 2026-09-16
test: verify pinned cert-manager version supports pinned Kubernetes (#322)

lmilbaum:
## Summary
- Adds `tests/test-cert-manager-kubernetes-support.py`, a CI gate that checks the `bootstrap.toml` cert-manager pin against the Kubernetes version pinned for the CAPD management cluster, using cert-manager's own "Currently supported releases" table.
- Only runs the network check when the diff actually touches the cert-manager pin or the Kubernetes pin (`--changed-from`), matching the existing digest-pinning gate's pattern.
- Wired into `validate.yml`'s `airgap-image-pins` job, which already has `fetch-depth: 0` for this kind of diff.
- Documents the check, and the upstream tracking, in `docs/dependencies.md`.

## Why
cert-manager and Kubernetes are pinned independently and bumped by separate Renovate PRs. Nothing previously caught a Kubernetes bump landing outside cert-manager's supported window: cert-manager v1.21.x (supports v1.33 → v1.36) was left paired with a Renovate-bumped Kubernetes v1.37.0, and the daily air-gapped deployment workflow failed silently for two weeks before it was root-caused (#322).

There's no structured/versioned feed for cert-manager's supported-version window, so this fetches and parses the same Markdown source that renders https://cert-manager.io/docs/releases/, rather than embedding a compatibility table in this repo that would go stale the same way. That's a known fragile workaround, not a real fix.

Checked cert-manager/cert-manager for a prior request to give the Helm chart's `kubeVersion` field a real upper bound (which would let this rely on chart metadata instead of scraping docs): found [#4132](https://github.com/cert-manager/cert-manager/issues/4132) (2021, closed by the stale bot without resolution) and its reopening as [#9123](https://github.com/cert-manager/cert-manager/issues/9123) (currently open). Left a comment on #9123 with this incident as concrete motivation instead of filing a duplicate.

## Test plan
- [x] `mise exec -- python3 tests/test-cert-manager-kubernetes-support.py --changed-from main` on this branch: skips (neither pin changed by this PR itself).
- [x] Verified against real pins: cert-manager 1.21.1 + Kubernetes v1.37.0 (current `main`) → correctly fails, matching the real #322 incident.
- [x] Verified against PR #323's downgrade target (Kubernetes v1.36.4) → correctly passes.
- [x] `mise run validate` passes.


shrinedogg:
Merged mid-batch (squash 7cbb67967), after the toolbox/renovate refactors so its validate.yml wiring absorbed their changes via a single main-merge refresh (clean, no conflicts). One flaky pass on the Zarf mise asset pin job, green on rerun. Note for Renovate follow-ups: this gate is the enforcement half of the pairing that #332 fixes and #331 groups.

PR #328, opened 2026-09-16
ci(air-gapped): allow workflow_dispatch to verify its own branch

lmilbaum:
## Summary
- `airgap/scripts/offline-run.sh` hardcoded `--certificate-identity` to `refs/heads/main`, so `zarf package verify` always failed for a `workflow_dispatch` run off any other branch, since cosign's keyless SAN reflects the actual triggering ref.
- Pass `github.ref` through as `AIRGAP_VERIFY_REF` and use it in place of the hardcoded value. The build and deploy jobs run in the same workflow execution, so this still only trusts a package signed by this repo's `air-gapped.yml` on the exact ref currently being verified, it just stops assuming that ref is always `main`.
- Let `workflow_dispatch` trigger the `build` job on any branch, not only `main`, so a PR's changes can be validated with a real offline deploy before merge.
- Adds `airgap/tests/test-airgap-verify-identity.py`: a regression test asserting `offline-run.sh`'s certificate-identity always uses the `${AIRGAP_VERIFY_REF:-refs/heads/main}` fallback (never a hardcoded ref), and that every workflow step invoking `offline-run.sh` wires `AIRGAP_VERIFY_REF` from `github.ref`. Wired into `validate.yml` and `mise run validate`.

## Why
Discovered while trying to verify PR #323 (the v1.36.4 downgrade) live: a manual `workflow_dispatch` run against that branch got past the build step but failed at signature verification with `expected SAN value ".../air-gapped.yml@refs/heads/main", got ".../air-gapped.yml@refs/heads/fix/downgrade-kubernetes-1.36-cert-manager-compat"` (run [35118716391](https://github.com/polarsquad/krops/actions/runs/35118716391)). That's a CI plumbing gap unrelated to the actual fix being tested, and it blocked getting a real before/after answer on the cert-manager timeout investigated in #322.

## Test plan
- [x] `python3 airgap/tests/test-airgap-verify-identity.py` -- OK, confirmed it fails against `main`'s prior hardcoded-ref state
- [x] `workflow_dispatch` off a branch reaches the actual deploy steps instead of failing at signature verification (verified live: runs 35154324226, 35156891686)
- [x] Scheduled/main runs still verify successfully (identity still resolves to `refs/heads/main`)


shrinedogg:
Merged after #330 in the batch (squash 8c32de4c5). This branch predated #330's job deletion, so the refresh resolved the air-gapped.yml overlap by taking #330's side (monitored-deploy stays gone); #328's AIRGAP_VERIFY_REF wiring in offline-run.sh and its regression test survive intact and pass on the refreshed branch.

PR #329, opened 2026-09-16
ci(air-gapped): capture cert-manager state when zarf package deploy fails

lmilbaum:
## Summary
- When `zarf package deploy` fails in `offline-run.sh`, dump cert-manager's pods, pod descriptions, pod logs, cluster-wide events, and node status to `/tmp/airgap-cert-manager-debug.txt` before the job ends.
- Upload that file alongside the existing deployment evidence in both the isolated-deploy and monitored-deploy jobs.
- Adds `airgap/tests/test-airgap-debug-capture.py`: a regression test asserting `offline-run.sh` still writes the debug file and runs the pod/describe/events probes, and that every deployment-evidence upload step still lists it. Wired into `validate.yml` and `mise run validate`.

## Why
Zarf's Helm `--wait` blocks for the whole timeout with no progress output, so the existing failure ("unable to install chart context deadline exceeded") gives no clue what's actually stuck. The kind cluster and its pods are still up on the runner right after zarf gives up, so this is a cheap place to capture real diagnostic state instead of guessing.

Needed because the Kubernetes/cert-manager version-mismatch theory in #322 turned out to be wrong, verified live via PR #328 plus a `workflow_dispatch` run: cert-manager's Helm install hangs at this exact step even on a Kubernetes version inside its documented support window (v1.36.4). This capture is what actually found the real root cause (`ImagePullBackOff` from a chart/image version mismatch, now fixed in #332).

## Test plan
- [x] `python3 airgap/tests/test-airgap-debug-capture.py` -- OK, confirmed it fails against `main`'s prior state (no debug capture at all)
- [x] Triggered a real `workflow_dispatch` run: `/tmp/airgap-cert-manager-debug.txt` was uploaded with the `ImagePullBackOff` pod state that led to #332's fix


shrinedogg:
Merged last in the air-gapped family (squash ef8e9df86). The branch was rebuilt by its author onto the post-#328 main mid-review, so the final state is that head plus a clean merge of current main; both this PR's debug-capture regression test and #328's verify-identity test coexist and pass on the merged tree, and the monitored-deploy removal from #330 is preserved.

PR #331, opened 2026-09-16
renovate: group cert-manager's chart pin with its images.txt/zarf.yaml tags

lmilbaum:
## Summary
- Adds cert-manager's four `airgap/images.txt`/`airgap/zarf.yaml` image depNames (`quay.io/jetstack/cert-manager-{controller,cainjector,webhook,startupapicheck}`) to the existing `platform-charts` Renovate group, alongside its Helm chart pin.
- Adds a new annotation-driven custom manager for `airgap/zarf.yaml`'s own embedded cert-manager chart-version line, which nothing tracked before at all (confirmed with a real `renovate --dry-run=full` run against this repo: it returned zero dependencies for that line). It now resolves to `depName=cert-manager, datasource=helm` and joins `platform-charts` too.
- Checked every other file PR #332 had to touch by hand: `bootstrap.toml`, `pivot.sh`, and the five `HelmRelease` manifests (aws, local-host, local-talos, azure, workload/azure-base) already resolved to `depName=cert-manager/datasource=helm` via Renovate's built-in managers, so the pre-existing Helm-chart packageRule already covered them, no change needed there.
- Adds `tests/test-renovate-cert-manager-grouping.py`: one test proves the grouping boundaries (mirrors `tests/test-renovate-actions-grouping.py`'s harness pattern), the other proves the new `zarf.yaml` manager actually extracts the dependency, not just that the rule would group it if it did. Wires both into `validate.yml`.

## Why
cert-manager's chart version and its image tags were tracked as separate, ungrouped dependencies. Four separate Renovate PRs (#281-#284) bumped only the image tags to v1.21.2, while nothing bumped any chart-version pin, still at 1.21.1. Since `values/cert-manager.yaml` has no image-tag override, the chart deploys pods expecting v1.21.1 images by default, but Zarf had only mirrored v1.21.2, so every cert-manager pod sat in `ImagePullBackOff` and Helm's install wait ran out its 15-minute timeout (#322), initially misdiagnosed as a Kubernetes-version incompatibility (see the retraction on cert-manager/cert-manager#9123).

Same class of fix as the `kubernetes-version` group (#142), just for cert-manager instead of Kubernetes: keep the chart version and the images it actually needs from drifting apart, and make sure every location that pin lives in is actually tracked in the first place.

## Test plan
- [x] `python3 tests/test-renovate-cert-manager-grouping.py` -- green against the real Renovate engine (both grouping and extraction assertions)
- [x] `python3 tests/test-renovate-actions-grouping.py` -- unaffected
- [x] `npx renovate-config-validator renovate.json5` -- valid
- [x] `python3 tests/test-renovate-coverage.py` -- unaffected, including the repo-level dry-run's duplicate-registry-URL check


shrinedogg:
Merged after #327 and before #332 in the cert-manager chain (squash 2101185e6). Refresh conflict in docs/dependencies.md resolved keep-both: #331's platform-charts grouping prose retained alongside #354's CAPI-manager bullet, and the Dockerfile bullet superseded by main's #337 wording (the kept HEAD text still described the removed inline uv pin). zarf.yaml needed no resolution here; #332's bump lands on top next and carries the annotation-line/v1.21.2 splice.

PR #332, opened 2026-09-16
fix: bump cert-manager chart to 1.21.2 to match images.txt (#322)

lmilbaum:
## Summary
- Bumps every cert-manager Helm chart-version pin from 1.21.1 to 1.21.2: `bootstrap.toml`, `pivot.sh`, `airgap/zarf.yaml`'s chart entry, and the five `HelmRelease` manifests `test-bootstrap-config.py` cross-checks against `bootstrap.toml` (aws, local-host, local-talos, azure, workload/azure-base).
- Adds `tests/test-cert-manager-version-consistency.py`: a permanent regression test cross-checking `bootstrap.toml`'s cert-manager version against `pivot.sh`, `airgap/zarf.yaml`'s embedded chart version, and the actual image tags in `airgap/images.txt`/`airgap/zarf.yaml`. Wired into `validate.yml` and `mise run validate` (pure local file check, no network/docker needed).

## Why
`airgap/images.txt` and `airgap/zarf.yaml`'s image list already pin cert-manager's container images at v1.21.2 (four separate Renovate PRs, #281-#284), but the chart version everywhere else stayed at 1.21.1. `values/cert-manager.yaml` has no image-tag override, so the chart deploys pods expecting v1.21.1 images by default; Zarf had only mirrored v1.21.2 into the air-gapped registry, leaving every cert-manager pod in `ImagePullBackOff` and Helm's install wait running out its 15-minute timeout (#322). Confirmed live via a debug capture on a `workflow_dispatch` run (#329):

```
cert-manager-controller: Image: 127.0.0.1:31999/jetstack/cert-manager-controller:v1.21.1-zarf-...
  zarf.dev/original-image-cert-manager-controller: quay.io/jetstack/cert-manager-controller:v1.21.1
Status: Pending, State: Waiting, Reason: ImagePullBackOff
```

Nothing checked any of these three places before (the five `HelmRelease` manifests were the only cert-manager-version consumers `test-bootstrap-config.py` covered), so nothing would have caught this drift, or would catch it recurring. Verified the new test fails with the exact real mismatch when pointed at `main`'s pre-fix state, and passes once every pin here is aligned. This is a belt-and-suspenders backstop alongside the `platform-charts` Renovate group added in #331: the group stops future Renovate-driven drift, this catches drift regardless of how it happens (manual edit, partial merge, a rebase dropping one file's change, ...).

This was originally misdiagnosed as a Kubernetes-version incompatibility in #322 and PR #323, and reported upstream as such before being retracted (cert-manager/cert-manager#9123). It has nothing to do with Kubernetes version support.

## Test plan
- [x] `python3 tests/test-bootstrap-config.py` -- OK (3 charts, 4 environments)
- [x] `python3 tests/test-cert-manager-version-consistency.py` -- OK, and confirmed it fails with the real historical mismatch when pointed at `main`'s pre-fix state
- [ ] Live `workflow_dispatch` run reaches past the cert-manager component without `ImagePullBackOff`

Refs: #322, #331


shrinedogg:
Merged after #331 so the bump absorbed the new zarf.yaml chart annotation (squash 555e2674a). Refresh conflict resolved keeping that annotation line together with v1.21.2. Verified on the refreshed branch: the consistency test passes (chart and images both 1.21.2 across bootstrap.toml, pivot.sh, zarf.yaml, images.txt, and the five HelmReleases). This closes the ImagePullBackOff root cause of #322.

PR #333, opened 2026-09-17
ci: register the k3s management-cluster prototype workflow

lmilbaum:
## Summary
Adds `prototype-k3s-mgmt.yml`, a dormant, `workflow_dispatch`-only workflow, to `main`.

## Why
GitHub only allows dispatching a `workflow_dispatch` workflow once its file exists on the default branch, even though the run itself then executes whichever ref is targeted. This registers the file so it can be dispatched against `prototype/k3s-management-cluster`, which has the actual script changes (`MGMT_CLUSTER_ENGINE=k3s` support) it exercises.

This workflow has no trigger except manual dispatch, doesn't touch `air-gapped.yml` or anything on the daily schedule, and does nothing until someone runs it.


shrinedogg:
Merged in this PR-batch triage (squash fa9c6a798). Registers the dormant dispatch-only k3s prototype workflow; single new file, disjoint from everything else in the batch, no overlaps.

PR #338, opened 2026-09-17
ci: add ruff lint + fix stale toolbox docs from PR #269

lmilbaum:
## Summary
- Adds `ruff` (pinned, scoped to `F401` + `PLW1510`) as a dev dependency and wires it into `mise run validate` and a new `python-lint` CI job.
- `F401` (unused imports): catches the stray `stat` import left over from PR #269 in `tests/test-toolbox-entrypoint-podman.py`.
- `PLW1510` (`subprocess.run` without explicit `check=`): the default `check=False` silently swallows a non-zero exit code.
  - `tests/test-toolbox-entrypoint-podman.py` discarded its `subprocess.run()` result entirely — a broken entrypoint script could still pass the test. Now captures the result and asserts `returncode == 0`, printing stderr on failure.
  - `airgap/tests/test-digest-regex.py` already inspects the returned `CompletedProcess` deliberately, so it gets an explicit `check=False` to document that intent.
- Neither rule is a general style linter for this repo — scope is intentionally narrow.
- `tests/test-toolbox-entrypoint-podman.py` only ever ran with `CONTAINER_HOST` unset, so it couldn't distinguish the fix's conditional default (`${CONTAINER_HOST:-...}`) from an unconditional override. Added a second case that pre-sets `CONTAINER_HOST` and asserts the entrypoint preserves a user-supplied value.
- Auditing PR #269 further turned up two stale docs/comments it left behind, fixed here:
  - `bootstrap-rs/toolbox-entrypoint.sh` and `scripts/toolbox-run.sh` both claimed a raw `podman run` command is documented in `docs/operations.md`; it only documents raw `docker run` and points Podman users at `scripts/toolbox-run.sh` instead.
  - `docs/bootstrap-cli.md`'s "toolbox runtime contracts" list omitted `CONTAINER_HOST`, which the entrypoint now sets as a real runtime contract for the podman case.

## Test plan
- [x] `uv run ruff check .` — all checks pass
- [x] `mise run validate` — green end to end, including the new "Python lint (ruff: unused imports, subprocess.run check=)" step
- [x] `uv sync --locked && uv run ruff check .` (the new CI job's exact commands) succeeds from a clean `.venv`
- [x] Manually verified the toolbox test's new returncode check actually catches a broken entrypoint (simulated a failing `podman info`, confirmed the test now surfaces `returncode: 1` / `ERROR: Podman not reachable` instead of passing silently)
- [x] Manually verified the new override-case test catches a regression: temporarily changed the entrypoint's `${CONTAINER_HOST:-...}` to an unconditional `export CONTAINER_HOST=...` and confirmed the test fails with `entrypoint overrode a user-supplied CONTAINER_HOST`; reverted after confirming
- [x] `bash -n` on both edited shell scripts


shrinedogg:
Merged after #341 and #342 (squash 9e4ff7631). The author's rebuilt head already carried the PLW1510 check=False fixes for the two test files #342 added, and the merge of current main was clean. The delete-vs-modify overlap on tests/test-toolbox-entrypoint-podman.py resolved with #341's deletion, since CONTAINER_HOST defaulting now lives in bootstrap-rs/src/engine.rs; this PR's ruff gate continues to cover the Python side. Ruff is clean on the merged tree.

PR #342, opened 2026-09-17
fix(scripts): resolve toolbox environment before engine selection

### LINKED ISSUE #257: refactor: resolve toolbox environment before engine selection
## Problem

In the wrapper introduced by #148, configuration is consumed before it is loaded: `scripts/toolbox-run.sh` selects `TOOLBOX_IMAGE`, detects the engine, and resolves its socket before parsing `.env`. The parser can then replace `CONTAINER_ENGINE` or `ENGINE_SOCK` after those decisions.

For example, invoke the wrapper directly on a host with Docker and Podman available and `CONTAINER_ENGINE="podman"` in `.env`. Detection can select Docker and its socket first; loading `.env` then changes the final executable to Podman while retaining the Docker-derived socket. Running through mise adds another `.env` loader and can hide this ordering problem.

The bespoke parser also claims to accept valid identifiers but allows leading digits (`1BAD=value` reaches `export` and fails under `set -e`). A line without `=` can become a key/value pair rather than being rejected. These are signs that loading, validation, precedence, and execution need a simpler boundary.

## Proposed scope

- Load and validate supported configuration once, before resolving the image, engine, or socket.
- Define a consistent precedence for process environment versus `.env`, and make direct wrapper and mise invocation agree. Document any intentional compatibility change.
- Prefer an existing suitable loader or a small explicitly defined assignment format over growing an ad hoc dotenv interpreter. Do not introduce a new host runtime dependency merely to parse this file; do not silently execute arbitrary file content as a shortcut.
- Resolve dependent values only after the final configuration is known.
- Preserve deliberate container environment forwarding. Audit the existing allowlist against documented lifecycle overrides (for example MGMT_KUBECONFIG and timeout knobs), making unsupported overrides explicit instead of silently dropping them. Coordinate canonical settings/default ownership with #150 rather than creating a second config registry.

## Acceptance criteria

- Direct and mise-based launches resolve the same engine/socket/image for the same effective configuration.
- Tests use stub engine commands and synthetic, non-secret assignments to cover precedence, quoted values, empty values, malformed identifiers, and engine/socket consistency without starting infrastructure.
- No secret values are logged; forwarded configuration retains intended quoting and values.
- The loader has a documented limited contract or uses a maintained implementation, with less duplicated configuration logic.
- Run relevant shell checks and `mise run validate`; update runtime docs and AGENTS.md if the invocation contract changes.

Related: #148, #150, #255. This is separate from the raw AWS quickstart documentation problem already fixed by #154.

lmilbaum:
## Summary

Closes #257.

`scripts/toolbox-run.sh` selected `TOOLBOX_IMAGE`, detected the container engine, and resolved its socket before parsing `.env`. A `.env`-selected `CONTAINER_ENGINE` or `ENGINE_SOCK` could then override those decisions after the fact, so a Docker-detected socket could end up paired with the Podman binary `.env` asked for. Direct and mise-based invocations could also disagree, since mise's own `env_file` loading races the wrapper's separate parser.

- Load `.env` first, before any detection or resolution runs.
- An already-exported variable (set on the command line, or by mise's own `env_file` loading before the script runs) is left alone: process environment wins over `.env` - the same precedence mise itself uses - so direct and mise-based invocations resolve identically for the same effective configuration.
- The parser also silently accepted invalid input: a line with no `=` fell through with key and value both set to the whole line, and a leading-digit key like `1BAD=value` passed the identifier check and reached `export`, which fails under `set -e`. Both are now rejected instead of exported.
- Add `tests/test-toolbox-run-env-precedence.py` (stub `docker`/`podman`, synthetic `.env`) covering precedence, quoting, empty values, malformed identifiers, and engine/socket consistency, without starting any real infrastructure. Wired into `mise run validate`, which also now syntax-checks the script itself.

## Test plan

- [x] `bash -n scripts/toolbox-run.sh`
- [x] `python3 tests/test-toolbox-run-env-precedence.py` (8/8 passing)
- [x] `uv run ruff check tests/test-toolbox-run-env-precedence.py`


shrinedogg:
Merged early in the batch (squash 119721dc7), before #341 and #338 were refreshed. Env-first precedence matches the documented repo-local .env TOOLBOX_IMAGE path in the e2e runbook; malformed-line rejection and process-env-wins semantics reviewed against the new precedence test.

Issue #267, opened 2026-09-13
bug: airgap digest gate accepts overlong SHA-256 digest tokens


lmilbaum:
## Problem

The offline air-gap image digest gate accepts malformed digest tokens whose first 64 characters are valid lowercase hexadecimal.

In `airgap/tests/test-airgap-image-digests.py`, `IMAGE_REF` captures `sha256:[a-f0-9]{64}` without requiring the digest token to end. A reference with 65 hexadecimal characters is truncated by the match and treated as a valid pin, so CI can approve an invalid image reference.

Source reviewed: https://github.com/polarsquad/krops/blob/37f25484acf5b9616464971a5a82d12877dc2319/airgap/tests/test-airgap-image-digests.py#L31

## Reproduction and limitations

Load the exact final #166 gate into an isolated temporary fixture. Set `airgap/images.txt` to `ghcr.io/example/app:v1@sha256:` followed by 65 lowercase `a` characters, and provide a valid pinned image in the fixture's `airgap/zarf.yaml`. Invoke the gate's `--all` path: it reports success and returns 0 because it matches only the first 64 digest characters.

This was reproduced through the gate's `main()` with a temporary repository root. It is a local parser finding, not a registry lookup or live deployment failure. No claim is made that current committed inventories contain such a malformed digest. The gate is intentionally offline and need not establish whether a digest exists or matches a tag upstream.

## Acceptance criteria

- Validate the complete digest token: require exactly 64 lowercase hexadecimal characters after `sha256:` and reject extra token characters.
- Reject malformed references explicitly rather than silently skipping them when tightening the regex.
- Preserve valid reference parsing in inventories, Zarf image lists, shell quotes, and shell parameter defaults.
- Add isolated regression cases for 63-, 64-, and 65-character digests and trailing non-hex characters; exercise tagged and digest-only references, preserving the existing readable-tag requirement.
- `mise run validate` and relevant CI checks pass.

Related: #166 (introduces the gate), #165 (immutable image enforcement), #170 (separate source-discovery gaps).

PR #343, opened 2026-09-17
test(airgap): verify pinned digests exist and tighten digest-length regression

lmilbaum:
Follow-up to #272 / #267.

## What
- Adds `airgap/tests/test-airgap-image-existence.py`: a live gate that queries every digest pinned in `airgap/images.txt` against its registry with `docker buildx imagetools inspect`, confirming the manifest actually exists. The existing offline gates only check that a digest is *shaped* like a valid sha256 (64 hex chars); they can't tell a real digest from a typo'd-but-well-formed one. Wired into a new `airgap-image-existence` job in `.github/workflows/validate.yml`, kept separate from `mise run validate` (needs network/registry access, same reasoning as the existing `renovate-digest-pinning` job). It only queries pins that already parse as shape-valid; a malformed digest is never queried.
- Extends `airgap/tests/test-digest-regex.py` (the #267 regression test) to also cover:
  - the digest-only reference form (`name@sha256:...`, no tag), at both too-long (65) and too-short (63) lengths, plus the valid-64 case (rejected for an unrelated reason: no readable tag)
  - the too-short (63-char) case for the tagged form, alongside the existing too-long (65) case
- Improves `airgap/tests/test-airgap-image-digests.py`'s failure message for the digest-only fallback case: an out-of-length digest-only reference doesn't match the `digest_only` group, so the regex re-parses `sha256:<hex>` itself as a bare name:tag reference. The failure now names the real offending reference (including the full original image name) instead of just the synthetic `sha256:<hex>` fragment.
- Adds a short section to `docs/airgap.md` documenting the digest-length boundary coverage and the live existence check, moving detail out of source comments.

## Verification
- `python3 airgap/tests/test-airgap-image-existence.py` — all 45 currently-pinned digests confirmed pullable
- Confirmed the failure path against a deliberately bogus digest (registry correctly rejects, exit 1)
- `python3 airgap/tests/test-digest-regex.py` — green, covers both reference forms and both length-boundary directions
- `python3 airgap/tests/test-airgap-image-digests.py --all` and `test-airgap-ownership.py` — green
- `uv run ruff check --select F401,PLW1510` — clean (matches this repo's own selected ruff rules)


shrinedogg:
Merged after #338 (squash 0f655c25a). The refresh conflict on test-digest-regex.py was formatting-only: both sides had already added check=False to the same subprocess.run call, so the PR's multi-line form was kept. The author's rebuilt head (based on the post-#329 main) was adopted, plus a clean merge of current main.

PR #240, opened 2026-09-10
chore(renovate): track staged CAAPH release assets

lmilbaum:
## Summary

- teach Renovate to update CAAPH release URLs, SHA-256 checksums, staged target paths, the deploy action, and clusterctl config together
- use JSON5/Handlebars-safe replacement templates with real newlines and quotes
- add release-asset coverage and retain digest-pinning scope for Docker-only dependencies

## Validation

- 
- Renovate fixture tests attempted locally; live GitHub release lookup requires CI token/network access

## Depends on

- #226. This stacked PR is based on its focused render-refactor branch and should be merged after #226.

PR #344, opened 2026-09-17
chore: automerge patch/minor bumps of the CI-only Renovate CLI pin

lmilbaum:
## Summary
- Adds a packageRule scoping `automerge: true` (patch/minor only) to the `renovate` npm dependency that pins the CLI version used by the `renovate-digest-pinning` CI job in `.github/workflows/validate.yml`.
- The global "no automerge anywhere" rule (issue #74) stays in place for everything else; this carves out one exception for a dependency whose own bump is validated by the CI job that installs it and re-runs the full pinning test suite against it.

## Why
Renovate ships patch releases several times a week, and each one previously needed a manual merge click even though nothing about the change is unreviewed: the CI run reinstalls that exact version and exercises it against the real digest-pinning tests. Issue #74's operating model explicitly leaves room for automerge on "demonstrably low-risk patch updates with reliable validation," which this is.

## Test plan
- [x] `RENOVATE_CONFIG_FILE=renovate.json5 renovate-config-validator` passes
- [x] `python3 tests/test-renovate-actions-grouping.py`
- [x] `python3 airgap/tests/test-renovate-digest-pinning.py` (under Node 24, with a GitHub token)
- [x] `python3 tests/test-renovate-coverage.py`
- [x] `python3 tests/test-renovate-release-assets.py`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #345, opened 2026-09-17
fix: harden PR #240's Zarf renovate matchers and dedupe test cost

lmilbaum:
## Summary
Follow-up review fixes for #240 (CAAPH/clusterctl Zarf renovate matchers):

- The five `"$clusterctl" generate provider ...` regex matchers hardcoded 14 literal spaces of indentation in `autoReplaceStringTemplate`, unlike the sibling `- source:`/`shasum:`/`target:` matchers, which already capture `(?<indent>\s*)` dynamically. If that section of `airgap/zarf.yaml` is ever reindented, these five matchers would silently emit a mis-indented line. All five now capture and replay the indent.
- The `shasum:` capture regex only accepted lowercase hex (`[a-f0-9]{64}`). Nothing else in the repo validates that field stays lowercase, so an uppercase/mixed-case SHA-256 would silently fail to match. Broadened to `[a-fA-F0-9]{64}` across all six release-file matchers that use it.
- `tests/test-renovate-release-assets.py`, `tests/test-renovate-coverage.py`, and `airgap/tests/test-renovate-digest-pinning.py` each spin up an independent Renovate subprocess with live GitHub API lookups against overlapping depNames (`kubernetes-sigs/cluster-api*`). They run sequentially in the same CI job, so `tests/renovate_harness.py` now points them at a shared, persistent `RENOVATE_CACHE_DIR` to let repeat datasource lookups hit Renovate's on-disk cache instead of GitHub again.

## Test plan
- [x] `renovate-config-validator renovate.json5` passes
- [x] Directly matched all five reworked regexes and the reworked shasum regex against the real `airgap/zarf.yaml` content (Node) to confirm correct capture groups
- [x] `python3 -m py_compile` on the three changed Python files
- [ ] CI's `renovate-digest-pinning` job (live GitHub lookups; couldn't authenticate locally to run `test-renovate-release-assets.py` / `test-renovate-coverage.py` end-to-end, same limitation noted in #240 itself)


shrinedogg:
Merged after #344 and before #354 in the batch (squash b5144e586). The #354 branch predates this PR, so its refresh resolved the renovate.json5 conflict by porting the indent-capture and case-insensitive shasum hardening into the consolidated alternation managers. Verified on the refreshed branch: renovate-config-validator green, coverage test extracts 8+4+4 deps covering all four providers.

Issue #313, opened 2026-09-15
feat: example workflow delivering Crossplane as a workload app on top of krops-managed resources


shrinedogg:
## Context

The README positions Crossplane as the closest comparison to krops and then
draws the line: krops introduces no krops-specific CRD or controller, it
combines Cluster API, ACK, and Flux and does not wrap resource APIs that
already say what you mean. Readers take that as krops versus Crossplane.
It is not. krops is a delivery and lifecycle pattern (Flux reconciles
controllers and their CRs from Git, CAPI owns the clusters); Crossplane is a
composition layer that turns resource APIs into higher-level platform APIs.
Crossplane v2 makes the fit explicit: a namespaced XR can compose any
Kubernetes resource, including a third-party CR such as the ACK `Bucket` this
repo already declares in `workload/base/s3-buckets/bucket.yaml`.

Nothing in the repo shows that. #224 (Argo CD) settled that krops does not add
a second GitOps engine; the same reasoning says Crossplane is not a second
engine at all, it is a workload app delivered by the engine krops already
has, exactly like the ACK controllers in `workload/base/aws-operators/` and
Podinfo in `workload/local-host/podinfo/`.

## Proposal

Add Crossplane as a workload app, following the `aws-operators` /
`s3-buckets` pattern from `docs/extending.md`, and document the resulting
workflow as the worked example of krops and Crossplane composing rather than
competing:

- Flux (krops) delivers Crossplane, its composition function, the XRD, the
  Composition, and the XR instances from Git.
- Crossplane reconciles the XR into composed resources.
- The existing controller (ACK on AWS, plain Kubernetes on local-host)
  reconciles the composed resources. No Crossplane provider replaces it.

Two tiers, mirroring how the repo already splits cloud-free and cloud paths:

1. `local-host` (cloud-free, exercised by the bootstrap and CI):
   `workload/local-host/crossplane/` installs Crossplane and a function;
   `workload/local-host/crossplane-example/` ships an XRD, a Composition, and
   one XR whose composed resources are ordinary Kubernetes objects that
   reach Ready without any cloud. Same two-Kustomization ordering as
   `aws-operators` (wait: true) then `s3-buckets` (dependsOn).
2. `aws` (CI-rendered, live-verified when a run is possible):
   `workload/base/crossplane/` and `workload/base/crossplane-example/`, where
   the Composition composes the ACK `Bucket`
   (`s3.services.k8s.aws/v1alpha1`) with the same security posture as
[...]

PR #347, opened 2026-09-18
feat(aws): run ACK controllers on the management cluster only

### LINKED ISSUE #346: arch: run ACK controllers on the management cluster only (AWS)
## Context

Every ACK controller in this repo runs twice today: once on the management
cluster (`mgmt/aws/infrastructure/ack-controllers/`, IAM + EKS only, static
SOPS credentials) and once per workload cluster
(`workload/base/aws-operators/`, S3 + RDS + IAM, authenticated via EKS Pod
Identity). The management cluster's IAM/EKS controllers exist only to create
the `PodIdentityAssociation` objects the workload-cluster controllers need to
authenticate (`mgmt/aws/infrastructure/ack-pod-identity/`).

ACK controllers talk to the AWS API directly; they do not need to run inside
the account/cluster whose resources they manage. The management cluster
already owns every `Cluster` object (CAPA/CAPI) and is the only cluster with
static AWS credentials in Git. Running ACK there exclusively removes an
entire authentication path (EKS Pod Identity, the three
`krops-ack-*-controller` IAM roles, the per-cluster `PodIdentityAssociation`
objects) and one controller install per workload cluster, without losing
anything: the same `Bucket` / `DBInstance` / `Role` CRs get created, tagged
per cluster exactly as before.

This mirrors the direction #313 is already pushing (a single controller
instance is the authoring identity for resources scoped to many clusters —
see the GCP Config Connector precedent in
`docs/workload-resources.md`: "the management cluster's KCC is the authoring
identity") and sets up a template the follow-up Azure (ASO) and GCP (KCC)
sub-issues will reuse once this AWS pass lands.

## Proposal

Move the ACK S3, RDS, and IAM controllers from every workload cluster onto
the management cluster only:

- Extend `mgmt/aws/infrastructure/ack-controllers/` with the S3 and RDS
  HelmReleases (same static SOPS credentials already used there for IAM).
  Drop the EKS controller/chart — its only consumer
  (`PodIdentityAssociation`) goes away with this change.
- Delete `mgmt/aws/infrastructure/ack-pod-identity/` (the three
  `krops-ack-*-controller` roles and their `PodIdentityAssociation`
  objects) and `workload/base/aws-operators/`.
- Move the `Bucket`, `DBInstance`, and reader `Role` CRs from
  `workload/base/{s3-buckets,rds-instances,iam-roles}/` (Flux
  `postBuild.substituteFrom: cluster-vars`, one instance per workload
  cluster reconciling its own) to a new
  `mgmt/aws/infrastructure/workload-resources/` (one Flux `Kustomization`,
  `dependsOn: ack-controllers`, explicit per-cluster values — there is no
  `cluster-vars` ConfigMap on the management cluster, same reason
  `aws-global-iam/reader-user.yaml` already wildcards the account ID).
  Multi-region resources (eu-west-1) use the `services.k8s.aws/region`
  annotation, the same mechanism `ack-pod-identity` already uses today.
- Remove the `eks-pod-identity-agent` EKS addon from both workload cluster
  definitions; nothing on the workload cluster authenticates via Pod
  Identity anymore.
- `workload/base/` becomes an empty overlay (no AWS resource-provisioning
  apps left). Workload clusters keep their own Flux instance — ready for a
  future application workload (the local-host Podinfo pattern) — but
  reconcile nothing today.

## Pitfalls to handle

- **Static credentials gain scope.** Today the SOPS `aws-credentials`
  principal only needs IAM/EKS-management permissions; the `s3:*` /
  `rds:*` / `secretsmanager:*` (managed master password) / `kms:CreateGrant`
  permissions live on the narrowly-scoped, ACK-created pod-identity roles
  instead. Once S3 and RDS controllers run on the management cluster with
  the shared static credential, that credential needs the union of all
  three former role policies directly. This is a real least-privilege
  trade-off, not just a wiring change — document it plainly in
  `docs/aws-iam.md` and `docs/operations.md`'s prerequisites list.
- **CRD + controller readiness ordering.** `workload-resources` must
  `dependsOn: ack-controllers` with `wait: true` on `ack-controllers` (already
  true) so the `Bucket`/`DBInstance`/`Role` CRDs exist first.
- **Bucket global uniqueness without substitution.** `bucket.yaml` relies on
  `${AWS_ACCOUNT_ID}`/`${CLUSTER_NAME}` today; the management-side version
  needs literal values (the account ID is already a literal elsewhere in
  `mgmt/aws/addons/flux-apps/flux-instance.yaml`, so this isn't a new
  pattern).
- **`bootstrap.toml` / `teardown.sh` drift.** `teardown.toml`'s
  `global-iam-roles` list and `teardown.sh`'s `GLOBAL_IAM_ROLES` still name
  the three deleted `krops-ack-*-controller` roles; the pod-identity-association
  sweep step becomes dead code. `tests/test-bootstrap-config.py` hard-codes
  the same three role names — update all three together or CI catches the
  drift on the role names but not the dead sweep code.
- **Diagram doc debt.** `docs/aws-infra.svg` is a hand-authored diagram, not
  generated from the mermaid block in `docs/architecture.md`; there's no
  regeneration tooling in this repo. The mermaid source (GitHub- and
  mkdocs-rendered) is the change of record; the SVG is a known follow-up.

## Explicitly out of scope

- Azure (ASO) and GCP (KCC) get their own follow-up sub-issues once this
  pattern is proven here; this issue is AWS/ACK only.
- Any application workload for the now-empty `workload/base` AWS overlay.
- Regenerating `docs/aws-infra.svg` (see Pitfalls).

## Acceptance criteria

- `mise run validate` passes: every kustomize overlay builds, `bash -n`
  passes for the touched shell scripts, and `tests/test-bootstrap-config.py`
  cross-checks cleanly against the updated `bootstrap.toml`.
- No `Bucket`, `DBInstance`, or ACK-controller-authored `Role` CR remains
  under `workload/`; all three exist once each per workload cluster under
  `mgmt/aws/infrastructure/workload-resources/`.
- No `PodIdentityAssociation`, no `krops-ack-*-controller` IAM `Role`, and no
  `eks-pod-identity-agent` addon remain anywhere in `mgmt/aws/`.
- `docs/architecture.md` (prose + mermaid diagram + both reconciliation-order
  sections), `docs/aws-iam.md`, `docs/workload-resources.md`,
  `docs/extending.md`, `docs/operations.md`, and `README.md` (docs table +
  repository layout) reflect the new topology.

## Verification

CI (`mise run validate`, the Renovate coverage tests, konflate's rendered
diff) plus a live AWS bootstrap/teardown run when one is possible, the same
CI-first-live-later gating #313 already uses (#236 precedent).

Related: #313 (parent effort), #224 (one engine per object, the precedent
this generalizes: one controller instance per resource type, not one per
cluster).

lmilbaum:
## Summary

- ACK's S3, RDS, and IAM controllers now run once, on the management
  cluster (`mgmt/aws/infrastructure/ack-controllers/`), and manage every
  workload cluster's resources directly through the AWS API. No workload
  cluster runs an ACK controller or holds AWS credentials.
- Deletes `mgmt/aws/infrastructure/ack-pod-identity/` (the three
  `krops-ack-*-controller` IAM roles and their `PodIdentityAssociation`
  objects) and `workload/base/aws-operators/`; the `Bucket`, `DBInstance`,
  and reader `Role` CRs move from `workload/base/{s3-buckets,rds-instances,
  iam-roles}/` to the new `mgmt/aws/infrastructure/workload-resources/`
  (literal per-cluster values, no `cluster-vars` substitution on the
  management cluster; `services.k8s.aws/region` for eu-west-1).
- Drops the now-unused `eks-pod-identity-agent` EKS addon and the
  `cluster-vars` ConfigMap from both workload clusters; `workload/base/`
  is an empty overlay.
- Updates `bootstrap.toml`, `teardown.sh`, and
  `tests/test-bootstrap-config.py` for the removed IAM role names, and
  `docs/architecture.md`, `docs/aws-iam.md`, `docs/workload-resources.md`,
  `docs/extending.md`, `docs/operations.md`, `README.md`, and `AGENTS.md`
  for the new topology.

Documents the real trade-off in `docs/aws-iam.md`: the shared static
credential now needs the union of the S3/RDS/IAM permissions that used to
live on narrowly-scoped, per-controller Pod Identity roles.

Fixes #346 (sub-issue of #313).

## Test plan

- [x] `mise run validate` (every kustomize overlay builds, `bash -n` on the
      touched scripts, `bootstrap.toml` cross-check)
- [ ] Live AWS bootstrap/teardown run. Tick only with kubectl output showing
      `ACK.ResourceSynced=True` for all six workload-resources CRs on the
      management cluster (not only Flux Ready; the Kustomization has
      `wait: false`).

## Rollout gate

Merge only after the live AWS environment is torn down on main: main's
teardown still sweeps the three `krops-ack-*` controller roles, so tearing
down after merge would orphan them.

Out of scope: regenerating `docs/aws-infra.svg` (the mermaid block is the
record); follow-up to be filed.


docs/aws-iam.md
```diff
@@ -1,46 +1,89 @@
...
+buckets and RDS instances directly — work that used to be scoped to the
+narrow, ACK-created Pod Identity roles above — that shared static credential
+needs a wider set of permissions than before:
+
+- `iam:CreateRole`/`DeleteRole`/`GetRole`/`UpdateRole`/`PutRolePolicy`/
+  `DeleteRolePolicy`/`TagRole` scoped to `krops-*` roles (the per-cluster
+  reader roles)
+- `iam:CreateUser`/`PutUserPolicy`/`GetUser`/`GetUserPolicy`/`TagUser` (the
+  `krops-reader` console user)
+- `s3:CreateBucket`/`DeleteBucket`/`GetBucket*`/`PutBucket*`/`ListBucket`
```
shrinedogg: Required: the branch conflicts with main. #353 (028522d) added s3:TagResource, s3:UntagResource, s3:DeleteBucketTagging and s3:ListTagsForResource to mgmt/aws/infrastructure/ack-pod-identity/role.yaml, which this PR deletes (git merge-tree: one modify/delete conflict, nothing else). Resolve by keeping the deletion, then carry those four actions into the permission list this PR now treats as the record of what the static credential needs (see R2). Without them a fresh bootstrap reproduces #352 against the static credential instead of the pod-identity role.

Required: this list is narrower than the policies it replaces. The deleted role.yaml / rds-role.yaml / iam-role.yaml listed s3:GetEncryptionConfiguration, s3:PutEncryptionConfiguration, s3:Get/PutLifecycleConfiguration and the other non-PutBucket* pairs explicitly because the PutBucket* glob does not cover them; the RDS role also had rds:AddTagsToResource / rds:RemoveTagsFromResource (the DBInstance declares tags); and main's role.yaml now has the four tagging actions from #353. Since these manifests were the record of least privilege, please reproduce the three inline policy documents verbatim (post-#353) under a "Minimum IAM policy for the ACK credential" heading here, and have docs/operations.md point at that section instead of paraphrasing (R1, R5).

docs/operations.md
```diff
@@ -166,17 +166,21 @@ You also need:
...
-  (managed master passwords) used by the workload clusters' ACK RDS
-  controllers are granted through the Git-declared
-  `krops-ack-rds-controller` pod-identity role; no extra static
-  credentials are required for them.
+  The ACK S3, RDS, and IAM controllers run only on the management cluster and
+  share this same static credential (no EKS Pod Identity, no per-cluster IAM
+  role), so the principal additionally needs the union of what those
+  controllers do directly: `iam:CreateRole`/`PutRolePolicy`/`GetRole`/`TagRole`,
+  `iam:CreateUser`/`PutUserPolicy`/`GetUser`/`GetUserPolicy`/`TagUser` (for the
+  `krops-reader` console user), `s3:CreateBucket`/`PutBucket*`/`GetBucket*`
```
shrinedogg: Required: same gap as R2 in docs/aws-iam.md: this summary omits the non-PutBucket* S3 actions, rds:AddTagsToResource/RemoveTagsFromResource and the #353 tagging actions. Replace the inline action list with a one-line pointer to the verbatim policy section in docs/aws-iam.md so there is one place to keep current.

mgmt/aws/infrastructure/workload-resources/buckets.yaml
```diff
@@ -0,0 +1,125 @@
...
+# Security posture (unchanged from the pre-move workload/base/s3-buckets/):
+#   - All public access blocked
+#   - Server-side encryption enforced (SSE-S3 / AES256, bucket keys enabled)
+#   - Versioning enabled
+#   - ACLs disabled (BucketOwnerEnforced)
+#   - Bucket policy denies any request not using TLS
+# ══════════════════════════════════════════════════════════════════════════════
+apiVersion: s3.services.k8s.aws/v1alpha1
+kind: Bucket
+metadata:
```
shrinedogg, line 27: Required: every CR in this directory (and aws-global-iam/reader-user.yaml) will be created twice per bootstrap. The kind bootstrap cluster reconciles mgmt/aws/infrastructure before the pivot (ack-controllers and ack-pod-identity are in the suspend list the pivot prints), its controllers create the AWS resources, clusterctl move does not carry ACK CRs, and the management cluster's Flux then applies the same CRs to controllers that find the resources already there. Live evidence on the current management cluster (028522d): Role/ack-iam-controller, Role/ack-rds-controller, Role/ack-s3-controller and User/reader-user are all ACK.Terminal=True "Resource already exists", ACK.ResourceSynced=False. After this PR the same happens to both Buckets, both DBInstances (a 10-minute create the kind side starts and abandons) and both reader Roles, so the acceptance check "kubectl get dbinstances -n ack-system" would show Terminal on a fresh bootstrap.

Fix: add to metadata.annotations of every Bucket, DBInstance and Role here and to the User in aws-global-iam/reader-user.yaml:

    services.k8s.aws/adoption-policy: adopt-or-create

All four kinds look the resource up by a spec field (spec.name / spec.dbInstanceIdentifier), so no adoption-fields annotation is needed. Supported by all controllers since aws-controllers-k8s/community#2406; the pinned charts postdate it. Please also add a sentence to docs/aws-iam.md explaining why the annotation is there (the pivot), and note in the PR description that the live run must check ACK.ResourceSynced=True, not only Flux Ready (workload-resources has wait: false).

mgmt/aws/addons/flux-apps/flux-instance.yaml
```diff
@@ -5,11 +5,9 @@
 #
 # Each workload cluster runs its own Flux instance syncing the workload/ overlay
 # for its region (workload/eu-north-01, workload/eu-west-01). Those overlays pull in
-# workload/base, which installs the ACK S3 operator and the Bucket CRs.
-#
-# Each ConfigMap also ships a `cluster-vars` ConfigMap into flux-system on the
-# workload cluster; the aws-operators / s3-buckets Flux Kustomizations use it
-# for postBuild variable substitution (region, cluster name, account ID).
+# workload/base, which is empty today: ACK now runs only on the management
```
shrinedogg, line 8: Fix: two files this PR keeps still describe the cluster-vars ConfigMap this hunk removes: bootstrap.toml:120 and teardown.sh:224-225 (see R5). Separate question (S2): Azure and GCP keep cluster-vars for workload apps; keeping the AWS one costs nothing and preserves the pattern for the future Podinfo style workload the PR description mentions. Either way is fine, but the two comments must follow the choice.

docs/architecture.md
```diff
@@ -88,54 +86,29 @@ flowchart TD
...
     AWSIAM -->|creates| RUSER
     RUSER -.->|sts:AssumeRole| RD1
     RUSER -.->|sts:AssumeRole| RD2
-    ACKPI -->|binds SAs to roles| EKS1
-    ACKPI -->|binds SAs to roles| EKS2
-
-    FA -->|"HelmChartProxy: flux-operator<br/>CRS: FluxInstance + cluster-vars + pull secret"| WF1
+    WR -->|Bucket CR reconciled| B1
+    WR -->|Bucket CR reconciled| B2
+    WR -->|DBInstance CR reconciled, eu-west-1 via<br/>services.k8s.aws/region annotation| DB1
```
shrinedogg: Fix: DB1 is the eu-north-1 instance and uses the controller's default region; only DB2 (and B2) carry services.k8s.aws/region: eu-west-1. Label the DB1 edge "DBInstance CR reconciled" and keep the annotation note on DB2 only (the B1/B2 edges above already follow this split).

mgmt/aws/infrastructure/workload-resources/dbinstances.yaml
```diff
@@ -1,8 +1,10 @@
 ---
 # ══════════════════════════════════════════════════════════════════════════════
-# Smallest-footprint PostgreSQL instance, reconciled by the ACK RDS controller
-# running on this workload cluster. The controller's region is ${AWS_REGION}
-# (cluster-vars ConfigMap), so each cluster gets a DB in its own region.
+# Smallest-footprint PostgreSQL instance per workload cluster, reconciled by
+# the ACK RDS controller on the MANAGEMENT cluster
+# (mgmt/aws/infrastructure/ack-controllers/). The controller's default region
+# is eu-north-1; the eu-west-1 instance overrides it with the
+# services.k8s.aws/region annotation (see buckets.yaml for why).
```
shrinedogg, line 7: Fix: line 16 still says "Workload clusters have no SOPS key, so an in-Git password secret is not an option". The CR now lives on the management cluster, which does have the age key, so the reason is gone; the choice is still right for a different reason. Suggested wording: "Master password: manageMasterUserPassword: RDS generates the password and stores it in Secrets Manager (prefix rds!), so it never enters Git or the cluster, the same posture as Azure (passwordAuth: Disabled) and GCP (IAM-only auth), see docs/workload-resources.md."

docs/workload-resources.md
```diff
@@ -17,9 +18,11 @@ see [AWS authentication & IAM](./aws-iam.md).
 
 ## RDS instances
 
-`workload/base/rds-instances/dbinstance.yaml` creates one PostgreSQL 17 instance
-per cluster (`krops-<cluster>-db`) in that cluster's own region; the ACK
-RDS controller runs with `aws.region: ${AWS_REGION}`:
+`mgmt/aws/infrastructure/workload-resources/dbinstances.yaml` creates one
+PostgreSQL 17 instance per cluster (`krops-<cluster>-db`) in that cluster's
```
shrinedogg, line 22: Fix: the next bullet (new file line 29, unchanged context just below this hunk) still says "workload clusters have no SOPS key, so an in-Git password secret is not an option"; same as R8 in dbinstances.yaml, that no longer applies on the management cluster. Reword both to the posture argument (password never enters Git or the cluster; matches Azure and GCP).

teardown.sh
```diff
@@ -377,11 +357,13 @@ _cleanup_rds_instance() {
     2>/dev/null || warn "  Failed to delete RDS instance $_rds_db"
 }
 
-# ── S3 buckets (created by the ACK S3 controller on workload clusters) ─────────
-# Like the RDS instances, the Bucket CRs live on the workload clusters, so the
-# buckets are orphaned when the clusters are deleted. Buckets are versioned
-# (see workload/base/s3-buckets/bucket.yaml), so every object version AND delete
-# marker must be purged before the bucket itself can be deleted.
+# ── S3 buckets (created by the ACK S3 controller on the management cluster) ────
+# Like the RDS instances, this sweep bypasses Flux/ACK entirely: step 1
```
shrinedogg, line 361: Warning: this comment says the sweep is safe because "step 1 suspends Flux", but suspending Kustomizations does not stop the ACK controllers, and teardown.sh (unlike the Rust CLI, which deletes the management EKS cluster in 4c before sweeping RDS/S3/IAM) never removes the self-managed management cluster. Before this PR the resource-owning controllers died with the workload clusters in step 2; now they keep running on the management cluster with the Bucket/DBInstance/Role CRs still present, so on the next ACK resync they recreate whatever 4c/4e/4f deleted (RDS included). Minimal fix for the shell path: in the non-AWS_ONLY branch after step 1, scale the controllers to zero when the namespace exists:

    if kubectl get namespace ack-system >/dev/null 2>&1; then
      kubectl scale deployment --all -n ack-system --replicas=0 \
        || warn "Could not scale ACK controllers down - continuing anyway"
    fi

helm-controller does not revert replicas (drift detection is off), and the Kustomizations are already suspended. The management-cluster gap itself is pre-existing (bootstrap.toml has mgmt-eks-cluster-name only for the CLI) and out of scope here; this note is about not extending it to RDS and S3.

tests/test-bootstrap-config.py
```diff
@@ -199,9 +199,8 @@ def main() -> int:
     teardown = config.get("teardown", {})
     if teardown:
         expected_roles = [
-            "krops-ack-s3-controller",
-            "krops-ack-rds-controller",
-            "krops-ack-iam-controller",
+            "krops-eu-north-1-workload-reader",
+            "krops-eu-west-1-workload-reader",
```
shrinedogg: Suggestion: this keeps the test green by hard-coding the same two names on both sides, which is what the test exists to avoid ("cross-checks them against the Git manifests that define the same names", bootstrap.toml:43). Now that the names are literals in one directory, read them from there (yaml is importable in the interpreter this task runs):

    import yaml

    WORKLOAD_RESOURCES = REPO_ROOT / "mgmt/aws/infrastructure/workload-resources"

    def ack_spec_values(filename: str, kind: str, field: str) -> set[str]:
        docs = yaml.safe_load_all((WORKLOAD_RESOURCES / filename).read_text())
        return {d["spec"][field] for d in docs if d and d.get("kind") == kind}

Inside the teardown loop (around new lines 176-184 and 203-208) replace the "krops-{cluster}-db" derivation and the hard-coded expected_roles with:

    reader_roles = ack_spec_values("roles.yaml", "Role", "name")
    ...
    rds_ids = ack_spec_values("dbinstances.yaml", "DBInstance", "dbInstanceIdentifier")
    for workload in td.get("aws-workloads", []):
        if workload["rds-instance"] not in rds_ids:
            failures.append(f"teardown rds-instance {workload['rds-instance']!r} not in dbinstances.yaml")
    for role in sorted(reader_roles - roles):
        failures.append(f"teardown.global-iam-roles missing {role} (roles.yaml)")
    for role in sorted(roles - reader_roles):
        failures.append(f"teardown.global-iam-roles lists {role}, not defined in roles.yaml")

If yaml is not acceptable in this python3-only test, keep the set logic and parse with a regex on the "name: krops-" / "dbInstanceIdentifier:" lines. mise run validate must still print "bootstrap.toml cross-check OK".

shrinedogg (changes requested):
Requesting changes: the topology and file layout are sound and match #346, but there is one merge conflict with main, three correctness gaps that only surface on a live run (pre-pivot double creation, a permission list that drops the #353 tagging actions, and a shell teardown that races the now-management-cluster controllers), the Rust CLI left out of the sweep removal, and a reference sweep that stopped short. Every Required item below blocks merge.

## Verified locally (on a961fc8, worktree; origin/main 028522d)
- git merge-tree origin/main pr-347: exactly one conflict, modify/delete on mgmt/aws/infrastructure/ack-pod-identity/role.yaml (R1). GitHub reports mergeable: CONFLICTING.
- CI on the PR head: kustomize build, yamllint, air-gap digest pins, docs site build, renovate digest pinning all pass. Rendered Flux diff (konflate) is skipped by design for fork PRs (konflate.yml fork condition), so the rendered diffs were produced locally instead.
- Rendered ack-controllers on the PR head: eks-chart removed, iam 1.9.0 / s3 1.11.0 / rds 1.12.0 present (56 changed lines vs main).
- eks-pod-identity-agent: 0 references in both rendered AWSManagedControlPlane specs (eu-north-1, eu-west-1).
- workload/eu-north-01 and workload/eu-west-01 render to 0 bytes (empty overlay); kubectl kustomize and flux build --dry-run exit 0.
- kubectl kustomize mgmt/aws/infrastructure/workload-resources: 6 objects (2 Bucket, 2 DBInstance, 2 Role), eu-west-1 objects annotated services.k8s.aws/region: eu-west-1; names equal bootstrap.toml rds-instance / s3-bucket-pattern / global-iam-roles.
- mise run validate: bash -n, bootstrap.toml cross-check OK (3 charts, 5 environments), all mgmt/** and workload/** kustomize builds pass.
- No PodIdentityAssociation, krops-ack-* Role, or pod-identity reference remains under mgmt/aws/ except the two reader-user.yaml comments (R5) and one buckets.yaml comment (R5).

## Required
- R1 (inline, docs/aws-iam.md:34): rebase and resolve the modify/delete conflict by keeping the deletion and carrying #353's four tagging actions into the new permission list.
- R2 (inline, docs/aws-iam.md:34 and docs/operations.md:174): the paraphrased permission list is narrower than the deleted role policies; reproduce the three inline policy documents verbatim (post-#353) and point operations.md at that section.
- R3 (inline, workload-resources/buckets.yaml:27): add services.k8s.aws/adoption-policy: adopt-or-create to every Bucket, DBInstance, Role here and to the reader User, or every bootstrap double-creates (live management cluster already shows Terminal "Resource already exists" on the current four IAM CRs).
- R4 (Rust CLI parity): bootstrap-rs/src/teardown.rs still runs cleanup_pod_identity_associations (706-744, call at 2230) and the 4a..4h numbering, while teardown.sh drops it. Remove the function and call, renumber, and update the doc comments in bootstrap-rs/src/config.rs:123-124 and :131 (they cite ack-pod-identity, workload/base/iam-roles/role.yaml and workload/base/s3-buckets/bucket.yaml) and docs/bootstrap-cli.md:225 ("covers pod identity associations"). AGENTS.md defines the CLI as a behavioral port of the scripts, so the two paths must agree. Verify with cd bootstrap-rs && cargo fmt --check && cargo clippy --locked --all-targets -- -D warnings && cargo test --locked.
- R5 (stale references, inline at flux-instance.yaml:8 plus this list): each line still describes the old topology and needs updating:
  - AGENTS.md:16 "ACK controllers, pod-identity roles, account-global IAM" -> "ACK controllers (S3/RDS/IAM), workload-resources (per-cluster Bucket, DBInstance, reader Role), account-global IAM".
  - AGENTS.md:87 "`base/`: ACK controllers and S3/RDS/IAM custom resources." -> "`base/`: empty overlay; the AWS resources live in mgmt/aws/infrastructure/workload-resources/ (#346)".
  - AGENTS.md:288 "EKS Pod Identity, ACK controller roles, reader user" -> "ACK on the management cluster only, static credential permissions, reader roles and user".
  - SECURITY.md:84-86 "Workload clusters use EKS Pod Identity; no static AWS keys are present on them" -> "Workload clusters hold no AWS credentials of any kind. The management cluster holds the CAPA and ACK credentials as SOPS-encrypted secrets; the ACK credential's permissions are listed in docs/aws-iam.md." (this is the security posture statement, so it must carry the trade-off too).
  - CONTRIBUTING.md:130 "Per-cluster values come from postBuild.substituteFrom: cluster-vars, not from hardcoded strings." -> mirror the AGENTS.md:196-201 wording this PR already added (management-side per-cluster manifests use literals).
  - docs/azure.md:11-12 and docs/gcp.md:11-12 comparison tables: the AWS column says "ACK controllers on workload clusters" / "EKS Pod Identity" -> "ACK controllers on the management cluster" / "static SOPS credential (no per-cluster identity)".
  - docs/secrets.md:19 "ACK IAM/EKS controller AWS credentials" -> "ACK IAM/S3/RDS controller AWS credentials".
  - bootstrap.toml:120 and teardown.sh:224-225 cite "the cluster-vars substitutions in mgmt/aws/addons/flux-apps/flux-instance.yaml", which this PR removes from that file -> cite the mgmt/aws/infrastructure/workload-resources/ literals instead.
  - tests/test-bootstrap-config.py:179 comment cites workload/base/rds-instances/dbinstance.yaml -> dbinstances.yaml (see S1).
  - mgmt/aws/infrastructure/aws-global-iam/reader-user.yaml:5 and :12 cite workload/base/iam-roles/role.yaml and ../ack-pod-identity/ -> ../workload-resources/roles.yaml (also gets the R3 annotation).
  - mgmt/aws/infrastructure/workload-resources/buckets.yaml:16 comment cites "the deleted ack-pod-identity/pod-identity-associations.yaml" -> drop the reference (the mechanism note is otherwise fine).
  - mgmt/azure/infrastructure/aso-workload-identity/swedencentral-workload.yaml:14 "the same way ACK retries PodIdentityAssociations" (identity.yaml:4 was updated, this sibling was not) -> "the same way ACK retries a Bucket whose controller is not yet Ready" or drop the comparison.
- R7 (inline, docs/architecture.md:94): the DB1 edge carries the eu-west-1 region-annotation label but DB1 is the eu-north-1 instance; only DB2 (and B2) override the region.
- R8 (inline, workload-resources/dbinstances.yaml:7 and docs/workload-resources.md:29): "workload clusters have no SOPS key" no longer applies on the management cluster; reword to the posture argument (password never enters Git or the cluster; matches Azure and GCP).

## Warning
- W1 (inline, teardown.sh:361): the shell sweep comment claims safety from "step 1 suspends Flux", but suspending Kustomizations does not stop the ACK controllers, and teardown.sh never deletes the self-managed management cluster (unlike the CLI, which removes it in 4c). The controllers now keep running on the management cluster with the CRs present and will recreate whatever the sweep deletes. Scale the ack-system deployments to zero in the non-AWS_ONLY branch after step 1.

## Suggestions
- S1 (inline, tests/test-bootstrap-config.py:203): the cross-check now hard-codes the same two reader-role names on both sides, which is what the test exists to avoid. Derive the expected names from mgmt/aws/infrastructure/workload-resources/ instead.
- S2 (inline, flux-instance.yaml:8): keeping the AWS cluster-vars ConfigMap for parity with mgmt/azure and mgmt/gcp is the smaller diff and keeps the workload Podinfo pattern usable; removing it is also acceptable if bootstrap.toml:120 and teardown.sh:224-225 follow. Author's call, but the two comments must follow the choice.
- S3 (PR description): (a) state the rollout gate, this merges only after the live AWS environment is torn down on main; (b) link a follow-up issue for regenerating docs/aws-infra.svg (out of scope here, the mermaid block is the record); (c) tick the live-run box only with kubectl output showing ACK.ResourceSynced=True for all six CRs on the management cluster.

## Looks good
- Topology and file layout match #346: ack-controllers gains s3-chart 1.11.0 and rds-chart 1.12.0 and drops eks-chart; ack-pod-identity and workload/base/aws-operators removed; workload-resources is one Kustomization with dependsOn ack-controllers.
- Rendered workload-resources: 2 Bucket, 2 DBInstance, 2 Role in ack-system, eu-west-1 objects annotated, names equal bootstrap.toml.
- Empty workload/base overlay reconciles (kubectl kustomize and flux build --dry-run exit 0); the FluxInstance pins 2.x and kustomize-controller 1.2.2 confirms empty directories reconcile and prune without error.
- eks-pod-identity-agent removed from both AWSManagedControlPlane specs; no PodIdentityAssociation or pod-identity reference remains under mgmt/aws/ except the comments noted in R5.
- README.md, docs/extending.md, and the Azure/GCP resource comments were updated to the new layout (not in the R5 stale list).

## Merge gating
- Tear down the live AWS environment on main (at or after 028522d) first: main's teardown still sweeps the three krops-ack-* controller roles, so tearing down after merge would orphan them. Confirm EKS clusters, krops-* roles, krops-reader user, and RDS instances are gone in both regions (never touch the unrelated training cluster in eu-west-1).
- Re-review the author's fixes against this list; approve only when every Required item is closed.
- Squash-merge, then run a fresh mise run bootstrap and verify ACK.ResourceSynced=True for all six workload-resources CRs on the management cluster (not only Flux Ready).

lmilbaum:
Addressed in 94d78c1 (single squashed commit, no rebase needed: the branch already sits on main including #353, so there was no modify/delete conflict left). R2: verbatim policies (post-#353) in docs/aws-iam.md, operations.md points there. R3: adoption-policy on all Bucket/DBInstance/Role/User plus an explanatory section. R4: Rust CLI pod-identity sweep removed, steps renumbered, doc comments updated; cargo fmt/clippy/test pass. R5/R7/R8: stale references fixed. W1: teardown.sh scales ack-system to zero after suspending Flux. S1: test derives names from workload-resources. S2: kept the cluster-vars removal and repointed the comments. S3: PR description updated (rollout gate, live-run criterion); the aws-infra.svg follow-up issue is not filed yet. mise run validate passes.

docs/aws-iam.md
lmilbaum: Fixed in 94d78c1. The branch already sits on main including #353, so no conflict remains. The four tagging actions are in the verbatim S3 policy in docs/aws-iam.md.

docs/operations.md
lmilbaum: Fixed in 94d78c1. The action list is replaced by a pointer to the verbatim policy section in docs/aws-iam.md.

mgmt/aws/infrastructure/workload-resources/buckets.yaml
lmilbaum: Fixed in 94d78c1. services.k8s.aws/adoption-policy: adopt-or-create is on every Bucket, DBInstance, Role and the reader User, and docs/aws-iam.md explains why.

mgmt/aws/addons/flux-apps/flux-instance.yaml
lmilbaum: Kept the cluster-vars removal (option 2). The comments in bootstrap.toml and teardown.sh now cite the workload-resources literals (94d78c1).

docs/architecture.md
lmilbaum: Fixed in 94d78c1. The DB1 edge is now labeled 'DBInstance CR reconciled'; only DB2 keeps the region annotation note.

mgmt/aws/infrastructure/workload-resources/dbinstances.yaml
lmilbaum: Fixed in 94d78c1. The comment now gives the posture argument (password never enters Git or the cluster; same as Azure and GCP).

docs/workload-resources.md
lmilbaum: Fixed in 94d78c1 with the same posture wording as dbinstances.yaml.

teardown.sh
lmilbaum: Fixed in 94d78c1. Step 1 now scales ack-system deployments to zero when the namespace exists, and the sweep comment says so.

tests/test-bootstrap-config.py
lmilbaum: Done in 94d78c1. The test derives RDS ids and reader role names from workload-resources. I used a regex rather than yaml to keep it stdlib-only.

PR #356, opened 2026-09-18
fix: drop the extra literal v from the zarf mise asset_pattern

### LINKED ISSUE #324: zarf mise pin: asset_pattern renders a double-v and fails to install
## What

`mise.toml` pins zarf with:

```toml
[tools."github:zarf-dev/zarf"]
version = "v0.85.0"
asset_pattern = 'zarf_v{{ version }}_{{ os(linux="Linux", macos="Darwin") }}_{{ arch() }}'
bin = "zarf"
```

The GitHub release asset is `zarf_v0.85.0_Linux_arm64`, but `version` here is the
full `v0.85.0`, so the template renders `zarf_vv0.85.0_Linux_arm64` (double `v`)
and the asset lookup fails:

```
mise ✗ github:zarf-dev/zarf@v0.85.0  4ms · failed: No matching asset found
for pattern: zarf_vv0.85.0_Linux_arm64
```

## Impact

- `mise x` / `mise install` with zarf in the active env fails to auto-install it,
  and with `set -e` a failing auto-install aborts the whole command. Reproduced
  on darwin-arm64 and linux/arm64 (see PR #317 review thread: the toolbox
  Dockerfile must use `mise -E gcp which gcloud` instead of `mise x` for this
  reason).
- `mise -E gcp run <task>` happened to install zarf successfully once
  (`zarf_v0.85.0_Linux_arm64`), so the behavior differs between mise subcommands
  on the same pin; that inconsistency is the confusing part.

## Suggested fix

Drop the extra literal `v` from the pattern (version already carries it):

```toml
asset_pattern = 'zarf_{{ version }}_{{ os(linux="Linux", macos="Darwin") }}_{{ arch() }}'
```

or switch to `extractVersion`-style matching. Verify by `mise x -- zarf --version`
on both arm64 and amd64, since the air-gapped workflow builds zarf bundles on an
arm64 runner.

## Found while

Addressing the PR #317 toolbox review (gcp environment); the Dockerfile gcp layer
worked around the failure rather than fixing the pin.

lmilbaum:
## What

`mise.toml`'s zarf pin double-prefixes the version with `v`, since the pin
(`v0.85.0`) already carries it. The asset pattern rendered
`zarf_vv0.85.0_Linux_arm64`, which never matches a real release asset
(`zarf_v0.85.0_Linux_arm64`), so `mise x` / `mise install` fails to
auto-install zarf.

## Fix

Drop the extra literal `v` from the pattern.

## Verification

Ran `mise x -- zarf version` on darwin-arm64 after the change: zarf installs
and reports `v0.85.0`.

Fixes #324


shrinedogg:
Merged ahead of the other mise.toml touchers in the batch (squash 73105385d). Verified against the main pin: the pattern now renders zarf_v0.85.0_Linux_arm64, matching upstream asset names. Fixes #324.

lmilbaum:
Correction to the description, for the record: the merged change does the opposite of the "Fix" section. It keeps the literal `v` in `asset_pattern` and drops it from `version` (`v0.85.0` to `0.85.0`). It also:

- remaps `arch(x64="amd64")`, fixing Linux amd64 hosts, which previously matched no zarf asset;
- changes the Renovate zarf rule to an unprefixed `version = "..."` match (that pattern also matched inside `min_version`; fixed in #443);
- adds `tests/test-mise-zarf-pin.py` and the `zarf-mise-pin` CI job;
- adds a note to `docs/dependencies.md`.

PR #357, opened 2026-09-18
fix: stop a comment block from truncating the toolbox-run.sh exec command

lmilbaum:
## What

A `#`-comment block explaining the `CLOUDSDK_CONFIG` override sat between two backslash-continued lines of the `exec "$CONTAINER_ENGINE" run ...` command in `scripts/toolbox-run.sh` (introduced in #317). In bash, a `\`-continued line followed by a `#`-comment line ends the logical command right there, so the comment silently truncated the command: the `CLOUDSDK_CONFIG` env var, the image argument, and the CLI args were never appended.

Every invocation of `scripts/toolbox-run.sh` (bootstrap/pivot/teardown, any profile) ran the container engine with no image argument, failing with a usage error and never starting the toolbox container.

## Fix

- Move the comment above the `exec` block instead of inside the continuation chain; trim it to a one-line pointer in the source and move the full rationale to `docs/operations.md`.
- Add `tests/test-toolbox-run-exec.py`, wired into `mise run validate`: it fakes the container engine binary, runs the wrapper, and asserts the image and CLI args actually reach it (verified it fails against the pre-fix script and passes against the fix).

## Verification

- `bash -n scripts/toolbox-run.sh` passes
- `python3 tests/test-toolbox-run-exec.py` passes against the fix and fails against the pre-fix `scripts/toolbox-run.sh` (confirmed the image argument is dropped there)
- `mise run validate`: all checks green, including the new test


maniankara:
Hi @lmilbaum I just noticed the same problem and created an issue and pr for it: https://github.com/polarsquad/krops/pull/365
You are first, but the extra python part
> Add tests/test-toolbox-run-exec.py, wired into mise run validate: it fakes the container engine binary, runs the wrapper, and asserts the image and CLI args actually reach it (verified it fails against the pre-fix script and passes against the fix).

I feel thats irrelevant for this bug?

lmilbaum:
> Hi @lmilbaum I just noticed the same problem and created an issue and pr for it: #365 You are first, but the extra python part
> 
> > Add tests/test-toolbox-run-exec.py, wired into mise run validate: it fakes the container engine binary, runs the wrapper, and asserts the image and CLI args actually reach it (verified it fails against the pre-fix script and passes against the fix).
> 
> I feel thats irrelevant for this bug?

@maniankara I agree with you. I tried to cut some corners. It is about time to start working properly.

PR #318, opened 2026-09-16
feat: GKE workload cluster resources via Config Connector

shrinedogg:
## What

Stacked on the PR for the gcp management side. Adds the GKE workload cluster's resources: `workload/gcp-base/` mirrors `workload/azure-base/` with Config Connector (the same pinned 1.156.0 operator bundle) replacing ASO, plus `workload/europe-north1-01/` pointing at it.

- `kcc`: operator + cluster-mode ConfigConnector via GKE-native Workload Identity (`krops-kcc`); the pinned bundle ships its own webhook certs, so no cert-manager on the workload cluster
- `networking`: Private Service Access range + VPC peering for Cloud SQL's private IP
- `storage`: bucket per cluster, S3 posture (uniform access, versioning, PAP enforced)
- `postgres`: Cloud SQL PostgreSQL 17, private IP only, TLS enforced, IAM authentication only (no password anywhere)
- `iam`: per-cluster reader service account + grants; the human identity impersonates it
- `tests/test-gcp-identity-chain.py`: cross-checks the WIF pool/provider/subject couplings, the credential JSONs, the bootstrap.toml pivot wiring and the workload bindings (wired into `mise run validate` and CI)
- docs: `docs/gcp.md` workload section, `docs/workload-resources.md` GCP subsection

Refs #72

## Verification

- `mise run validate` green (all workload overlays build; gcp identity chain passes)
- kustomize build of `workload/gcp-base` and `workload/europe-north1-01` verified
- KCC operator bundle on the workload side is byte-identical to the management side (the pin test enforces the version)

PR #360, opened 2026-09-18
fix(gcp): set the Cloud SQL edition, drop requireSsl, correct the IAM-only claim

lmilbaum:
## What

Follow-up to #318.

- `settings.edition: ENTERPRISE` is now explicit (`db-f1-micro` is Enterprise only, PG 17 can default to Enterprise Plus)
- `requireSsl` dropped, `sslMode: ENCRYPTED_ONLY` stays
- "IAM auth only / no password anywhere" reworded in docs and diagram: the IAM flag does not disable password login and the built-in `postgres` user still exists
- new `tests/test-gcp-postgres.py`, wired into `mise run validate` and CI

## Verification

- the test fails on main and passes here; the identity chain test still passes
- not applied to a live cluster; whether the API accepts the edition/tier combination is unverified


shrinedogg:
Merged first in the GCP follow-up pair (squash f3c3b938e). The author's rebuilt head (based on the post-#329 main) was adopted, plus a clean merge of current main. On the refreshed branch: the new postgres test passes and the identity-chain test stays green. The body's own caveat stands: the edition/tier combination is not yet verified against a live Cloud SQL API.

PR #361, opened 2026-09-18
fix(gcp): order iam after postgres, move reader comments to docs

lmilbaum:
## What

Follow-up to #318.

- `iam` Kustomization now `dependsOn` `postgres` (the SQLUser its `instanceUser` grant serves)
- reader.yaml rationale comments moved to `docs/workload-resources.md`, which also now covers the `cloudsql.viewer` grant the old header omitted
- `tests/test-gcp-identity-chain.py` checks the dependsOn and the docs

## Verification

- the test fails on main and passes here
- touches the same test file as the naming PR; expect a trivial rebase


shrinedogg:
Merged after #360 (squash 98abe3dd4). The author's rebuilt head was adopted (its flux-ks.yaml postgres dependsOn wording), then current main merged in cleanly. On the merged tree the identity-chain test carries both the (h) KCC-naming checks from #359's squash and this PR's (i) ordering/documentation checks, and passes.

PR #362, opened 2026-09-18
test(gcp): check the workload bucket name length

lmilbaum:
## What

Follow-up to #318. Adds a 63-character check for the bucket name to `tests/test-gcp-identity-chain.py` and documents it.

## Verification

- passes on main's manifests; temporarily raising the assumed project-number length made it fail with the expected message
- touches the same test file as the other follow-ups; expect a trivial rebase


shrinedogg:
Merged last in the batch (squash 2d074003c). Refreshed onto the post-#361 main; the identity-chain test conflict resolved keep-both so the (i) ordering checks and this PR's (j) bucket-name-length check coexist. First merge attempt failed cleanly (the refresh had missed #361's squash) and was redone against current main before merging.

PR #363, opened 2026-09-18
fix(bootstrap-rs): stop teardown kubectl stub tests flaking on ETXTBSY

lmilbaum:
The teardown tests write a shell-script kubectl stub and exec it immediately. When another test thread forks while the write fd is still open, exec fails with ETXTBSY. This intermittently failed different teardown tests on unrelated PRs (#329, #357).

The stub helper now probes the script (a no-op via STUB_PROBE) until it is executable before returning.

Verified: cargo test passes 3 times in a row locally.


shrinedogg:
Merged first in the batch (squash 080f6b1cb). The ETXTBSY stub fix is prerequisite hygiene for the teardown-test flakes seen today on unrelated CI runs (#329 and #338 cargo jobs, both green on a single rerun). No conflicts with the batch.

Issue #364, opened 2026-09-18
toolbox-run.sh bootstrap fails with "docker run" requires at least 1 argument


maniankara:
# Bug: `toolbox-run.sh bootstrap` fails with Docker argument error

## Description

Running:

```sh
scripts/toolbox-run.sh bootstrap
```

fails with:

```text
"docker run" requires at least 1 argument.
See 'docker run --help'.
```

## Cause

The `docker run` command was built using backslash continuations. A comment was placed inside the continued command after:

```bash
"${PASS_ENV[@]}" \
```

Because Bash treats `#` as the start of a comment, the remaining command arguments were ignored, including:

```bash
-e CLOUDSDK_CONFIG=/workspace/.gcloud
"$TOOLBOX_IMAGE"
```

Docker therefore received an incomplete command without an image argument.

## Expected Behavior

The script should start the toolbox container and run the requested lifecycle command:

```sh
docker run ... "$TOOLBOX_IMAGE" bootstrap
```

## Actual Behavior

Docker exits before starting the toolbox because no image argument is passed.

## Suggested Fix

Move the explanatory comments outside the continued `docker run` command:

```bash
# CLOUDSDK_CONFIG is canonical-by-design: it is always the repo-local
# .gcloud/ inside the /workspace mount.
exec "$CONTAINER_ENGINE" run --rm ${TTY_ARGS[@]+"${TTY_ARGS[@]}"} \
  -v "$REPO_ROOT:/workspace" \
  -w /workspace \
  -v "$SOCK_SOURCE:/var/run/docker.sock" \
  -v "$REPO_ROOT/.kube:/root/.kube" \
  -e KUBECONFIG="$KUBECONFIG_IN" \
  "${PASS_ENV[@]}" \
  -e CLOUDSDK_CONFIG=/workspace/.gcloud \
  "$TOOLBOX_IMAGE" \
  ${CLI_ARGS[@]+"${CLI_ARGS[@]}"}
```

## Validation

Run:

```sh
bash -n scripts/toolbox-run.sh
scripts/toolbox-run.sh bootstrap
```

PR #365, opened 2026-09-18
Fix #364: Toolbox Docker command construction

maniankara:
Move explanatory comments outside the backslash-continued docker run command so the toolbox image and runtime arguments are passed correctly. This prevents bootstrap from failing with 'docker run requires at least 1 argument.


lmilbaum (changes requested):
This fix already addressed in one of the open PRs

shrinedogg (approved):
Approved.

Verified locally:
- Stub-engine run of the old script drops the image and CLI args from the docker run argument list (60 args, ends at `-e CLOUDSDK_CONFIG`); the fixed script passes the full list (65 args, image + CLI args present).
- `bash -n` passes on both versions; the bug was semantic, which is why CI stayed green.
- Swept the tree for the same comment-inside-continuation pattern: no other occurrences; the raw `docker run` examples in `docs/operations.md` are clean.

Suggestion (non-blocking): the PR description has an unclosed quote: 'docker run requires at least 1 argument.' Minor, no action needed for merge.

Also worth noting in the commit message or the #364 thread: the broken comment came in with #317, so this un-breaks bootstrap for all profiles, not just GCP.

shrinedogg:
Thanks @maniankara for the report and the fix. Since this landed first, the core fix for #364 is in via this PR. The follow-up #357, merged right after, adds a regression test (`tests/test-toolbox-run-exec.py`) that guards against the truncation recurring, plus the `CLOUDSDK_CONFIG` rationale in `docs/operations.md`.

Issue #254, opened 2026-09-11
perf: consolidate CI and release image builds and use native architecture runners


lmilbaum:
## Problem

Optimize the container release workflow to avoid QEMU emulation when building the supported image architectures.

The current `.github/workflows/toolbox-release.yml` runs on `ubuntu-24.04-arm`, sets up QEMU for `amd64`, and builds both `linux/amd64` and `linux/arm64` in one job. ARM64 is already native; AMD64 is the emulated build. Merely selecting an ARM runner therefore does not remove emulation while retaining both architectures.

## Proposal

Build each architecture on a matching native runner, preferably in parallel:

- `linux/amd64` on an AMD64 runner.
- `linux/arm64` on an ARM64 runner.

Publish the resulting platform images by digest, then assemble a single multi-architecture OCI image index for the release. Remove QEMU setup when neither architecture requires emulation.

## Consolidate CI and release image builds

Also review duplication between the Docker image build job in `.github/workflows/bootstrap-rs.yml` and the image build steps in `.github/workflows/toolbox-release.yml`.

Consolidate the shared build configuration into one implementation used for both validation and release. Evaluate a single workflow with separate triggers/jobs versus a reusable workflow called by thin CI and release workflows. Choose the simplest structure that reduces duplicated build setup, architecture selection, caching, and smoke-test commands without obscuring the release process.

Preserve their distinct behavior: pull-request CI validates builds without publishing or requiring registry write/OIDC permissions; an authorized version-tag release publishes, assembles the multi-architecture index, signs, attests, and creates the project release. Keep Rust fmt/clippy/tests from the existing bootstrap-rs workflow intact. Preserve path filters, required-check names where relevant, and fork-PR compatibility.

The goal is fewer duplicated definitions and less maintenance, not merely fewer lines. Avoid introducing a complicated abstraction if a small shared workflow can express the common build logic clearly.

## Requirements

- Preserve support for both architectures and existing release version/tag semantics.
- Use isolated platform-specific cache scopes where needed so parallel builds do not overwrite each other's caches.
- Assemble and publish release tags only after all required platform builds and smoke tests succeed.
- Sign the final published index and retain appropriate SBOM/attestation coverage for both platform artifacts.
- Record the final index digest and platform digests in the project release artifact inventory proposed in #250.
- Coordinate image naming with #249; this optimization does not require changing the public image interface.

## Acceptance criteria

[...]

PR #366, opened 2026-09-18
fix(release): build toolbox image arches natively instead of under QEMU

shrinedogg:
## What

The toolbox release workflow cross-built `linux/amd64` under QEMU on the
`ubuntu-24.04-arm` runner. The amd64 `rustc` segfaults deterministically
under qemu-user-static there (`rustc -vV`, signal 11), so the `v0.2.0`
release failed twice and **no toolbox image was ever published to GHCR**
(`ghcr.io/polarsquad/krops-toolbox` returns `NAME_UNKNOWN`).

This makes each architecture build natively and merge the results:

- `build` matrix: `linux/amd64` on `ubuntu-24.04`, `linux/arm64` on
  `ubuntu-24.04-arm`, in parallel, each pushed as a single-arch image
  tagged `<version>-linux-<arch>` (provenance disabled for a clean merge).
- `merge`: `docker buildx imagetools create` combines the two single-arch
  images and publishes the stable tags `X.Y.Z`, `X.Y`, `latest`.
- `sign-attest`: unchanged behavior. Cosign keyless (workflow OIDC
  identity) plus a Syft SBOM attestation per stable tag; the merged index
  digest is resolved from the registry with `buildx imagetools inspect`.

Bumps the CLI to `0.2.2`: `v0.2.0` points at the broken QEMU commit, and
`v0.2.1` (the first native-build cut) published only its version tag after
docker/metadata-action collapsed the tag list, so the merge and sign jobs now
derive the stable tags in plain bash. Force-moving existing release tags is
off the table. The version gate only fires on tags, so the PR's non-tag pushes
don't run it.

## Verification

- `actionlint` clean; YAML parses; jobs `gate -> build -> merge ->
  sign-attest`.
- The merge mechanics (single-arch push, `imagetools create` to
  `0.2.1`/`0.2`/`latest`, digest extraction, two-platform index) were
  exercised end-to-end against a local `registry:2` before this change.
- `v0.2.2` was tagged from this branch and the live `toolbox-release` run
  (35377265296) is the proof: it published a signed, SBOM-attested
  multi-arch `ghcr.io/polarsquad/krops-toolbox` as `0.2.2`, `0.2`, and
  `latest` (one index, linux/amd64 + linux/arm64; cosign verified against
  the tag identity).
- No gate references the CLI version (checked `tests/`, `airgap/`,
  `scripts/`, `mise.toml`), so the bump is inert outside Cargo files.

Part of #254 (native architecture runners; the CI consolidation half is
out of scope here).


shrinedogg (commented):
Verified end-to-end against the live v0.2.2 release; the workflow does what it claims. All review findings are addressed in 0d52027 and 917b0d0.

Verified locally:
- actionlint clean; all 8 PR checks green on 917b0d0.
- v0.2.2 run 35377265296: gate, Build amd64, Build arm64, Merge, Sign/attest all success.
- imagetools inspect: latest, 0.2.2, 0.2 all at index sha256:e69e9980... with linux/arm64 + linux/amd64 children.
- cosign verify against identity toolbox-release.yml@refs/tags/v0.2.2: passed; SPDX attestations present.
- Tag-derivation bash simulated (stable, prerelease, two-segment): stable publishes X.Y.Z + X.Y + latest, prerelease omits latest.
- Cargo.toml is 0.2.2 at the tagged commit fbbf38c.

Findings and resolution:
- Provenance was not disabled despite the body claiming it: fixed in 0d52027 with provenance: false on the per-arch builds. Takes effect on the next tag build; the already-published v0.2.2 index keeps its two buildkit attestation children, which is cosmetic.
- Docs sweep incomplete (docs/bootstrap-cli.md still said no release was published): fixed in 917b0d0.
- README wrap/paragraph break at the toolbox bullet: fixed in 917b0d0.
- Stale PR body (0.2.1 vs 0.2.2): edited in place.

Left open, deliberately: the intermediate <version>-linux-<arch> tags and the orphaned 0.2.1 tag stay published in GHCR. Pruning them needs delete:packages, which GITHUB_TOKEN does not grant, so that cleanup does not belong in this workflow.

Looks good: native-runner matrix with fail-fast: false and per-arch cache scopes, registry-resolved digest in the sign job, prerelease latest guard, and gate -> build -> merge -> sign ordering so a bad tag never publishes.

PR #367, opened 2026-09-18
fix(ci): fetch full history in the zarf-mise-pin job

lmilbaum:
The "Zarf mise asset pin" job runs `tests/test-cert-manager-kubernetes-support.py --changed-from $BASE_SHA`, which runs `git diff BASE...HEAD`. The default depth-1 checkout lacks the base commit, so git exits 128 and the job fails on main and on every PR.

Sets `fetch-depth: 0` on that job's checkout.


shrinedogg:
Landed as the follow-up to the zarf-mise-pin failures this batch produced: every post-#327 main validate run went red in this job because the diff-based --changed-from gate needs full history, which the job's checkout did not fetch. One-file fix, verified green on the PR before merge (squash bcb77ac78). All 22 PRs in this cycle are now merged; the next scheduled main run should be the health check.

Issue #173, opened 2026-09-03
Extend Renovate digest pinning to every authoritative air-gap image source


lmilbaum:
## Problem

Renovate currently enables `pinDigests` only for `airgap/images.txt`, `airgap/zarf.yaml`, and the toolbox Dockerfile. The repository’s authoritative air-gap image surface also includes direct image references in `airgap/scripts/*.sh`, but Renovate has no general image manager or digest-pinning rule for those references.

The special `kindest/node` regex manager overlaps the scripts but captures only `currentValue` and emits a tag-only replacement, so it cannot add or preserve `currentDigest`.

As a result, `python3 airgap/tests/test-airgap-image-digests.py --all` reports unpinned images that Renovate cannot fix across:

- `airgap/images.txt`
- `airgap/zarf.yaml`
- `airgap/scripts/build-config-artifact.sh`
- `airgap/scripts/build-package.sh`
- `airgap/scripts/stage-and-create-cluster.sh`

## Goal

Make Renovate propose digest pins for every eligible external image in every authoritative air-gap image source while retaining readable tags.

## Proposed scope

- Add narrowly scoped custom Docker extraction for image references in `airgap/scripts/*.sh`.
- Capture optional `currentDigest` and ensure replacements retain `repository:tag@sha256:<digest>`.
- Fix or consolidate the overlapping `kindest/node` manager so it can add and preserve digests.
- Apply `pinDigests: true` to every authoritative air-gap image source recognized by the repository digest gate.
- Keep `localhost:5001/knr-ops-airgap:latest` as a narrow documented exception because it is built locally per run.
- Avoid managing generated or vendored paths already excluded under `airgap/manifests/`, `airgap/rendered/`, `airgap/archives/`, and `airgap/sbom/`.
- Coordinate with #162 so the integration test proves one digest proposal per eligible reference, not merely one proposal per file.
- Coordinate with #165 and #170 so Renovate extraction and committed-state enforcement use the same authoritative source set.
- Validate changes using the pinned Renovate dry-run procedure documented in `AGENTS.md`, including live datasource lookup and replacement-path verification.

## Acceptance criteria

- After stripping existing digests in the integration fixture, every eligible external image reference produces a `pinDigest` update.
- Renovate proposes digest pins for eligible references in both inventories and all authoritative air-gap shell scripts.
- Replacements retain readable tags and valid lowercase SHA-256 digests.
- Repeated references resolve consistently across authoritative sources.
[...]

PR #368, opened 2026-09-18
test(airgap): assert digest pins keep the tag and agree across sources (#173)

lmilbaum:
Adds two assertions to `airgap/tests/test-renovate-digest-pinning.py` (helpers in `tests/renovate_harness.py`) for the #173 criteria it did not cover:

- every `pinDigest` update keeps the tag (`newValue == currentValue`), so replacements stay `repo:tag@sha256:...`
- a repeated `repo:tag` reference gets the same digest in every authoritative source (19 multi-file refs today, 0 inconsistent)

Ran the test locally under Node 24.21.0 with Renovate 44.90.3; it passes. Checked the new helpers against fabricated bad data and both fail as intended.

Refs #173


shrinedogg (approved):
**Verdict: Approve** (no blocking findings)

Adds the two #173 acceptance criteria the integration test did not cover: pinDigest updates retain the tag, and repeated refs resolve to the same digest across the authoritative sources.

Verified locally (Node 24.21.0, Renovate 44.90.3, head 14b497a):

- `python3 airgap/tests/test-renovate-digest-pinning.py` passes: 88 pins proposed (images.txt 45, build-config-artifact.sh 12, build-package.sh 15, stage-and-create-cluster.sh 2, zarf.yaml 14), 0 tag-changing pins, 0 differing cross-file digests.
- 19 multi-file refs, 0 inconsistent, matching the PR description, from an independent run.
- Both helpers probed with fabricated bad data: a pinDigest with a changed tag is flagged, and two files carrying different digests for the same ref are flagged. A ref with no pinDigest update at all is invisible to `inconsistent_pin_digests`, which is fine because `deps_without_pin_digest` already fails that case.
- `ruff check` passes on both files.

Looks good:

- `newValue == currentValue` matches Renovate's lookup behavior: a pinDigest update is emitted with `newValue` set to `config.currentValue`, so any deviation would indicate a broken custom-manager extraction.
- Failure output names the file, ref, and offending values for both new checks, consistent with the existing missing-pin reporting.

PR #371, opened 2026-09-19
fix(airgap): pin cert-manager pod images by the bundled digests

lmilbaum:
## Problem

The scheduled air-gapped run failed at `zarf package deploy` (cert-manager Helm install, `context deadline exceeded`). Behind it were several separate bugs that only showed once the previous one was fixed:

1. **Digest-only registry names.** Zarf stores a `tag@sha256` bundle image by digest only, but rewrites a tag-only pod image to `<tag>-zarf-<crc>`, which was never pushed. Pods sat in ImagePullBackOff (`NotFound`). Not a Kubernetes/cert-manager version mismatch: 1.36.4 failed the same way.
2. **Missing pull secret.** Zarf only creates `private-registry` in namespaces it knows from charts, so the CAPI/CAAPH provider namespaces (created by `kubectl apply`) hit `no basic auth credentials`.
3. **Host-daemon images were untagged after `docker load`.** Saving `name:tag@sha256` leaves no tag and no RepoDigests, so kind/CAPD/`krops-registry` silently pulled from the internet. A connected runner hides this; an air gap would not.
4. **kubeadm needs `pause:3.10.2`** but `kindest/node:v1.37.0` bakes `3.10`, so `kubeadm init` on CAPD nodes hung offline.
5. **Workload Flux controllers were pinned `tag@digest`** but preloaded into the nodes by tag, so containerd tried to pull.
6. **Seeded podinfo chart was 6.14.0** while the workload tree requests 6.15.0.

## Fix

- Every pod image references the bundled digest: cert-manager and flux-operator chart values, FluxInstance controller patches, and `pin_image_digests` for the `clusterctl`-rendered CAPI/CAAPH manifests (fails if any image is left unpinned).
- `copy_registry_secret` copies `private-registry` into the provider namespaces.
- `save_host_images` pulls by digest, aliases the digest-less name, saves it, and fails if the archive lacks the tag. Host-daemon consumers (kind, `krops-registry`, CAPD `customImage`/`preLoadImages`, workload Flux patches) use the tag in the artifact; committed sources keep digests for Renovate.
- `pause:3.10.2` is preloaded into workload nodes.
- The podinfo chart version is read from `workload/local-host/podinfo/helm.yaml`.
- Service-readiness waits (`zarf init`/`deploy --timeout`, Flux Ready, node Ready) are 1 minute. The CAPD workload-cluster wait is 20 minutes and the job cap 60.
- Workload-cluster diagnostics are captured to `/tmp/airgap-workload-debug.txt` on failure and uploaded; later checks are skipped when the cluster is not Available.
- New tests: `test-airgap-cert-manager-digest-values.py` (also covers flux-operator and FluxInstance), `test-airgap-pin-image-digests.py` (`pin_image_digests`, `copy_registry_secret`); `docs/airgap.md` findings 8 to 10.

Renovate does not update the new values/patches; bump them together with `images.txt` (tests fail on drift).

## Verification

Full air-gapped run on this branch passes end to end (mgmt deploy, workload cluster Available, nodes Ready, workload Flux sync, podinfo Running): https://github.com/polarsquad/krops/actions/runs/35440535110


lmilbaum:
Superseded by #372, which uses a branch on the fork (lmilbaum/krops) as the head.

PR #373, opened 2026-09-19
fix(airgap): make the air-gapped deploy pull only bundled images

lmilbaum:
## Problem

In the air-gapped deploy, pods stayed in `ImagePullBackOff` until the Helm timeout, first cert-manager and then flux-operator. Zarf stores a `tag@sha256` bundle image by digest only. It rewrites a tag-only pod image to a `<tag>-zarf-<crc>` name that was never pushed, so the pull can never succeed offline.

## Changes

- **Digest pins in pod specs.** The cert-manager and flux-operator chart values, the FluxInstance controller patches, and the clusterctl-rendered CAPI and CAAPH manifests (via the new `pin_image_digests`) now reference the bundled digest. `pin_image_digests` fails if any `image:` is left unpinned.
- **Registry pull secret.** `copy_registry_secret` copies Zarf's `private-registry` secret into the provider namespaces that Zarf does not know about, so their pods can pull.
- **Host-daemon images.** A loaded image has no RepoDigests, so a digest reference never resolves locally. `save_host_images` saves each image under its digest-less tag and fails if the archive lacks it. The config artifact strips digests from `customImage`, `preLoadImages` and the workload Flux patches to match.
- **Missing image.** `registry.k8s.io/pause:3.10.2` is now bundled and preloaded. kubeadm pulled it and hung offline.
- **Podinfo chart version.** It is read from `workload/local-host/podinfo/helm.yaml` instead of being hardcoded in two places.
- **Faster failure, better diagnostics.** Zarf and Flux waits are 1m, the workload cluster wait is 20m and the job timeout is 60m. Workload cluster debug output is written to `/tmp/airgap-workload-debug.txt` and uploaded as a CI artifact.

## Tests

Two new checks run in `validate.yml` and `offline-run.sh`:
- `test-airgap-cert-manager-digest-values.py` ties the chart values and FluxInstance patches to `airgap/images.txt`.
- `test-airgap-pin-image-digests.py` checks that `pin_image_digests` rewrites images and rejects unpinned ones.

I did not run the air-gapped workflow end to end for this PR. `docs/airgap.md` has a new finding covering the failure and the fix.

Supersedes #371 and #372.


lmilbaum:
## Findings

The full write-up is in `docs/airgap.md` (findings 8 to 11). Summary:

**1. Tag-only pod images never resolve in the air gap.** Zarf stores a `tag@sha256` bundle image by digest only, and its agent rewrites a tag-only pod image to `<tag>-zarf-<crc>`, which was never pushed. cert-manager and then flux-operator sat in `ImagePullBackOff` until the Helm timeout. Every pod image now references the bundled digest: chart values, FluxInstance patches, and the clusterctl-rendered CAPI and CAAPH manifests (`pin_image_digests`, which fails on any unpinned image).

**2. Provider namespaces had no pull secret.** Zarf only creates `private-registry` in namespaces it knows from charts. `copy_registry_secret` now copies it from `flux-system` into the provider namespaces.

**3. Host-daemon images resolve by tag, never by digest.** A loaded image has no RepoDigests, so a `name@sha256` reference silently pulls from the internet. A connected runner hides this and an air gap fails. `save_host_images` saves each image under its digest-less tag and fails if the archive lacks it, and the config artifact strips digests from `customImage`, `preLoadImages` and the workload Flux patches.

**4. `kindest/node` bakes `pause:3.10`, but kubeadm wants `3.10.2`.** `kubeadm init` hung offline. `pause:3.10.2` is now bundled and preloaded. Re-check this whenever the node image is bumped.

**5. Smaller fixes.**
- The seeded podinfo chart version is read from the workload tree.
- Zarf and Flux waits fail after 1m; the workload cluster gets 20m.
- Workload debug output goes to `/tmp/airgap-workload-debug.txt`, uploaded by the workflow.
- The signer identity uses `github.repository` (default `polarsquad/krops`), so a fork verifies its own signature.
- The fork and branch guard on the build job is removed. Triggers stay `schedule` and `workflow_dispatch` only.

## Verification

- Air-gapped workflow on a fork, final commit: **success**, build 4.3 min, deploy 9.0 min, about 14 min total: https://github.com/lmilbaum/krops/actions/runs/35447730320. Deploy timeout lowered from 60 to 30 minutes on that basis; the only earlier upstream success took 15.4 min.
- PR CI: green.
- Every change has a test. New: `test-airgap-host-image-tags.py` (save_host_images, digest stripping, pause image, podinfo version) and `test-airgap-workflow-config.py` (triggers, no guard, timeouts). `test-airgap-verify-identity.py` now covers `AIRGAP_VERIFY_REPO`. I broke the code six ways and each new check failed as it should.

## Known limits

- The digests are duplicated by hand across `images.txt`, the chart values, the FluxInstance patches and `build-package.sh`. Renovate updates only `images.txt`, so bumps must be made together. The tests catch drift but do not prevent it. Generating the pins from `images.txt` is the better follow-up.
- The digest stripping in `build-config-artifact.sh` is pattern based, so a new field or format could slip past it. The post-strip guards cover only the two known cases.

airgap/manifests/flux-instance.yaml
```diff
@@ -70,26 +70,26 @@ spec:
         patch: |-
           - op: replace
             path: /spec/template/spec/containers/0/image
-            value: ghcr.io/fluxcd/source-controller:v1.9.4
+            value: ghcr.io/fluxcd/source-controller:v1.9.4@sha256:8a8ed0a57b8b86f561d5a4309a69f65e62f0cebe4de8801593c5ff35a3bc3c23
```
shrinedogg, line 73: Warning: the header comment at lines 13-19 still describes the old mechanism: controller images "pinned back to TAGS" and tag references rewritten by the Zarf agent to the <tag>-zarf-<hash> names that exist in the registry. The values below are now digest-pinned, and docs/airgap.md finding 8 explicitly says it supersedes finding 3's tag pins. The comment now contradicts both the file's own content and the updated docs; please rewrite it to describe the digest pinning (and why the manifest-list digests now resolve).

airgap/scripts/build-config-artifact.sh
```diff
@@ -199,6 +201,20 @@ open(path, "w").write(txt)
 print(f"patched {path}: preLoadImages added to {count} DevMachineTemplates")
 PY
 
+# Digest-less references for host-daemon images: docs/airgap.md finding 9.
+sed -E -i.bak '/customImage:|^ +- [a-z0-9.\/-]+(:[^@ ]+)@sha256:/ s/@sha256:[a-f0-9]{64}//' "$LH/clusters/docker/cluster-class.yaml"
+rm -f "$LH/clusters/docker/cluster-class.yaml.bak"
+sed -E -i.bak 's#(value: ghcr\.io/fluxcd/[a-z-]+:[^@ ]+)@sha256:[a-f0-9]{64}#\1#' "$LH/addons/flux-apps/flux-instance.yaml"
```
shrinedogg, line 207: Suggestion: the comment above the workload FluxInstance heredoc (lines 71-73) still attributes the tag pinning to "the embedded distribution's manifest-list digests do not resolve in a single-arch registry". After this PR the operative reason is finding 9: these images resolve from docker-loaded preLoadImages, which carry no RepoDigests, so a digest reference never resolves locally. Worth aligning the comment with the finding the strip now depends on.

shrinedogg (commented):
## Review summary

Verdict: no blocking findings. The functional change is verified end to end; the two warnings below are docs-consistency issues left over from the mechanism change.

### Verified locally (checkout of the PR head, diffed against fresh origin/main)

- Downloaded the 5 provider component assets at the exact URLs and shasums pinned in zarf.yaml (all shasums match) and ran pin_image_digests against them with the PR's images.txt: every rendered image is digest-pinned, none left unpinned. Each asset has exactly one image: line, all covered by images.txt.
- All 8 airgap test files pass locally, including the 4 new ones. Note: test-airgap-cert-manager-digest-values.py crashes under macOS system python 3.9 (the str | None annotation) but passes under the repo-pinned mise python 3.14.7, which is what mise run validate and CI use. Env artifact, not a defect.
- bash -n clean on all 5 changed scripts; test-airgap-image-digests.py --all passes.
- zarf v0.85.0 (the pinned version) supports --timeout on both init and package deploy.
- CI green. The skipped "Rendered Flux diff" check is konflate skipping fork PRs by design, not a failure.
- The author's fork run (lmilbaum/krops actions run 35447730320) covers the full air-gapped workflow end to end.

### Warnings

- AGENTS.md:121 still describes the air-gapped workflow as "upstream main only". Removing the fork and branch guard makes that stale, and AGENTS.md's own "Keeping this file current" rule asks for the update in the same PR. (The "two comparison deployments in parallel" phrase on line 123 was already stale on main since the workflow now has a single deploy job; worth fixing in the same edit.)
- See the inline comment on the flux-instance.yaml header: it still documents the tag-only pinning this PR replaces.

### Suggestions

- See the inline comment on build-config-artifact.sh: the workload FluxInstance comment still cites the single-arch registry rationale rather than finding 9's RepoDigests reason.

### Looks good

- pin_image_digests failing loudly on any unpinned image is the right fail-fast design, and it holds against the real deploy inputs (verified above).
- save_host_images aliasing plus the archive-content check closes the silent-pull hole from finding 9 without losing the digest guarantee.
- The digest strip in build-config-artifact.sh has post-strip guards for both known cases, so a format drift fails the build instead of shipping a broken artifact.
- Every behavioral change has a test wired into both validate.yml and mise run validate, and the known-limits section is honest about the manual digest duplication.

airgap/manifests/flux-instance.yaml
lmilbaum: Rewrote the header to describe the digest pinning and why the tag-only rewrite fails (finding 8). Fixed in 5adbe20.

airgap/scripts/build-config-artifact.sh
lmilbaum: Aligned the comment with finding 9 (loaded images have no RepoDigests). Fixed in 5adbe20.

shrinedogg:
## Follow-up on my review

All three findings are fixed in 5adbe20, verified on a checkout of that commit:

- AGENTS.md: the workflow paragraph now says "on any repository that carries it" and describes a single egress-blocked deploy with one evidence artifact, which matches the workflow file at that commit (one deploy job, egress deny plus the pcap check). The stale "two comparison deployments" phrase is gone.
- airgap/manifests/flux-instance.yaml header: now describes the digest pinning and why the tag-only rewrite fails (finding 8), matching the file's own values and docs/airgap.md.
- airgap/scripts/build-config-artifact.sh comment: now cites finding 9 (docker-loaded images carry no RepoDigests, so digest references never resolve locally).

Verification at 5adbe20: full `mise run validate` green (all airgap tests including the four new ones, bootstrap cross-check, every kustomize overlay) and `test-airgap-image-digests.py --all` green. The delta from the PR head d76f550 to 5adbe20 is comment and markdown only, with no executable change, so the green CI on d76f550 and your fork's air-gapped run 35447730320 remain valid evidence for the amended commit; no e2e re-run is needed.

## One thing before merge

5adbe20 is not in this PR. The PR still tracks d76f550 on lmilbaum/krops:fix/airgap-cert-manager-digest-values; the amended commit was pushed to the upstream repo's branch of the same name (the leftover head of closed #371). So the fixes do not appear in the PR diff, and the threads' "Fixed in 5adbe20" replies do not resolve against this PR's head.

Please push 5adbe20 to your fork's fix/airgap-cert-manager-digest-values (it amends d76f550, so pushing it means rewriting your own branch, which is fine here), then delete the upstream branch so the next fix does not land there either. I have not touched either branch.

airgap/manifests/flux-instance.yaml
shrinedogg: Verified at 5adbe20 (header now describes the digest pinning per finding 8). Note: 5adbe20 is not yet in this PR — see my follow-up comment.

airgap/scripts/build-config-artifact.sh
shrinedogg: Verified at 5adbe20 (comment now cites finding 9 and the RepoDigests mechanism). Note: 5adbe20 is not yet in this PR — see my follow-up comment.

lmilbaum:
Thanks for catching that. I had pushed 5adbe20 to the upstream branch by mistake.

- Pushed 5adbe20 to `lmilbaum/krops:fix/airgap-cert-manager-digest-values`. The PR head is now 5adbe20, so the fixes and the "Fixed in 5adbe20" replies resolve against this PR.
- Deleted the leftover `fix/airgap-cert-manager-digest-values` branch from `polarsquad/krops`.

The two review threads were already resolved.

PR #375, opened 2026-09-20
fix(bootstrap-rs): fail fast when AGE_PUBLIC_KEY contradicts the age key file

### LINKED ISSUE #350: fix(bootstrap): stale AGE_PUBLIC_KEY silently overrides the age.key pubkey and plants the wrong key in the sops-age secret
## Description

`AGE_PUBLIC_KEY` (an optional env var) silently overrides the public key derived from `AGE_KEY_FILE`, and nothing validates the two agree. A stale `AGE_PUBLIC_KEY` in `.env` (left over from a rotated key) makes the bootstrap plant the *wrong* key into the `sops-age` secret, so Flux cannot decrypt any `*.sops.yaml` after handoff.

The failure mode is nasty because it is not caught at bootstrap:

- The bootstrap only needs the key to *construct* the secret; it never tests decryption.
- The `sops-age` secret's key is named `keys.<pubkey>.agekey` — Flux's source-controller only uses entries whose key matches the public key embedded in the SOPS files. With a mismatched name, every SOPS-encrypted Kustomization fails with `no keys found in sops-age secret` (or the equivalent dry-run error) and the reconciliation chain stalls at the first encrypted resource.
- `AGE_PUBLIC_KEY` is read via `value("AGE_PUBLIC_KEY")` (empty string counts as unset, which is fine) and then preferred over `extract_age_pubkey(age_key_content)` in `preflight_github` (main.rs:820-824). The error message only appears when *neither* source yields a key.

## How it bit

On a clean 2026-09-18 AWS e2e run, the checkout's `.env` carried `AGE_PUBLIC_KEY` for an old age identity (`age18lc…`) while `age.agekey` / `.sops.yaml` were on the current one (`age1amq…`). Caught only by manual preflight cross-checking; the run was started with the var omitted so the derived key was used.

## Suggested fix

In the preflight where `age_pubkey` is resolved, when `AGE_PUBLIC_KEY` is set AND `extract_age_pubkey(age_key_content)` yields a *different* non-empty key, fail fast with a message naming both values (no secret material). Optionally also verify the pubkey matches the recipient(s) in `.sops.yaml` for the selected environment's `*.sops.yaml` paths.

## Alternatives

- Drop the `AGE_PUBLIC_KEY` override entirely; the key file is the single source of truth.
- If the override must stay (e.g. for setups where the key file is not readable by the operator), document that it must exactly match the key file, and add the cross-check above.

shrinedogg:
## Problem

`AGE_PUBLIC_KEY` (optional env var) silently overrode the public key derived from `AGE_KEY_FILE`, and nothing validated the two agree. A stale `AGE_PUBLIC_KEY` in `.env` (left over from a rotated key) made the bootstrap plant the wrong key into the `sops-age` secret: Flux names the secret entry `keys.<pubkey>.agekey` after that key, source-controller then matches no SOPS recipient, and every encrypted Kustomization fails with `no keys found in sops-age secret`. The bootstrap itself looked green because the key is only used to construct the secret, never to test decryption.

## Fix

New pure helper `resolve_age_pubkey` in `bootstrap-rs/src/main.rs`, called from `preflight_github` where `age_pubkey` is resolved:

- `AGE_PUBLIC_KEY` set AND the key file yields a DIFFERENT non-empty public key: fail fast, naming both public keys (no secret material), with remediation text.
- Override set and matching (or the key file yields no extractable public key): override wins, as before. Documented setups relying on the override keep working.
- Override unset or empty string: derived from the key file, as before.

Also documents the must-match constraint on the `AGE_PUBLIC_KEY` row in `docs/bootstrap-cli.md`.

## Tests

Four new unit tests: matching override accepted; mismatch rejected with both keys named in the error; unset override derives from the key file; empty-string override counts as unset; override kept when the key file yields no public key.

Sabotage-verified: restoring the pre-fix silent-override semantics makes the mismatch test fail (the stale key is returned unchanged).

Gates: `cargo fmt --check`, `cargo clippy --all-targets`, `cargo test` (92 passed), and `mise run validate` all green.

## Scope note

`bootstrap-common.sh` carries the same pre-fix pattern (`AGE_PUBKEY="${AGE_PUBLIC_KEY:-...}"`), but the shell scripts are the legacy path pending retirement (#92/#95/#100) and the issue scopes the fix to `bootstrap-rs`; left as-is here.

Fixes #350


shrinedogg (commented):
Verified clean against issue #350. Reviewed the diff on 0ed40e4 and re-ran the validation locally rather than trusting the description:

- Semantics of `resolve_age_pubkey` match the issue: two non-empty keys that differ bail naming both public keys; a matching override passes; unset or empty-string override derives from the key file; the override is never silently dropped (kept when the key file yields no extractable pubkey).
- Sabotage check: removing the mismatch arm (restoring pre-fix silent-override semantics) makes `resolve_age_pubkey_rejects_mismatched_override` fail with the stale key returned unchanged; source restored afterwards.
- End-to-end probe: built the binary and ran the real preflight (`KROPS_PROFILE=aws`) with a fresh `age-keygen` key file and a mismatched `AGE_PUBLIC_KEY`. Exits 1 before any mutation step, the error names both public keys, and no `AGE-SECRET-KEY` material appears in the output.
- Gates: `cargo fmt --check`, `cargo clippy --all-targets`, and `cargo test` (92 passed, matching the PR claim) all green.
- Sweep check: `age_pubkey` has a single consumer (the `keys.<pubkey>.agekey` entry in the `sops-age` secret), so there is no other call site carrying the old pattern. `bootstrap-common.sh:59` still has the pre-fix override, which the PR explicitly scopes out as the legacy path retiring under #92/#95/#100; consistent with the issue.
- `docs/bootstrap-cli.md` now documents the must-match constraint on the `AGE_PUBLIC_KEY` row.

No findings.

PR #376, opened 2026-09-20
fix(bootstrap): poll through the clean-first-run pivot prerequisites (#348, #349)

### LINKED ISSUE #348: fix(bootstrap): bootstrap→pivot chain fails on a clean first run (pivot Phase 0 before Flux has created the mgmt Cluster object)
## Description

`krops-bootstrap` (default `BOOTSTRAP_PIVOT=1`) chains the pivot immediately after the bootstrap completes. For `aws`, the pivot's Phase 0 (`pivot_check_management_cluster`) requires the management `Cluster` object to already exist in the kind bootstrap cluster, but creating that object is *Flux's* job after bootstrap hands off to GitOps (CAPA HelmRelease + providers + AWSClusterControllerIdentity + eu-north-1 overlay, a multi-minute dependency chain).

Result on a clean first run: bootstrap prints "Bootstrap complete! Flux is now reconciling from …" and the pivot fails within seconds:

```
>>> Bootstrap complete! Flux is now reconciling from https://github.com/polarsquad/krops
>>> Watch progress with: flux get kustomizations --watch
ERROR: Cluster 'eu-north-1-management' not found in the bootstrap cluster.
       The management cluster definition must be reconciled first:
       aws:        merged to main, Flux-in-kind creates it (~15-25 min)
```

Exit code 1. Recovery is a manual re-run of the same command once Flux has created the object (the "Pivot recovery" path), which works because `clusterctl move` and the bootstrap steps are rerun-safe.

Observed 2026-09-18 during a full AWS e2e run (clean account, toolbox image `krops-toolbox:dev` built from b27855c). The kind bootstrap phase took ~4 min; the management `Cluster` object appeared ~7 min after bootstrap completion (Flux still resolving CAPA provider → identity → AWSCluster).

## Why it happens

`pivot.sh` (the original script) has the same structure, but its operator ran it *manually* later, so the human wait absorbed the reconciliation delay. The CLI port (#95/#98) removed the human step and made the pivot the default exit, without adding a wait between "Flux controllers ready" and `pivot_check_management_cluster`.

Note: Phase 1 (`pivot_wait_for_management_cluster`, the 40m kubeconfig poll) only runs *after* Phase 0 passes, so its timeout cannot cover the object-creation delay.

## Suggested fix

In the bootstrap→pivot chain, after Flux controllers are ready, poll for the management `Cluster` object (the object Phase 0 checks) with a generous timeout (e.g. reuse `mgmt-ready-timeout` plus slack, or a dedicated `MGMT_DEFINITION_TIMEOUT`), before entering `run_pivot`. The wait should distinguish "Flux is still reconciling" from "Flux is stuck" (surface failed Kustomization conditions, as `watch_local_reconciliation` does for local-host).

## Alternatives

- Document the two-step flow explicitly (bootstrap, wait for `kubectl get cluster` in kind, re-run), and make the Phase 0 error message tell operators exactly that (it hints at it, but the `aws:` line suggests waiting ~15-25 min for a merge that has already happened).
- Give the CLI a real `pivot` subcommand (the docs' "there is no separate pivot subcommand" makes `mise run pivot` vs `krops-bootstrap pivot` confusing; the wrapper maps `pivot` to a bare `krops-bootstrap` invocation, which re-runs the whole bootstrap first).

### LINKED ISSUE #349: fix(pivot): Phase 2 node wait fails on nodeless EKS target (kubectl wait node --all errors before the node group registers nodes)
## Description

The pivot's Phase 2 (`pivot_export_kubeconfig`, main.rs:1843-1856) ends with:

```
kubectl --kubeconfig <target> wait --for=condition=Ready node --all --timeout=15m
```

For an AWS (EKS) management cluster this is the first operation against the target, and it fails immediately on a clean first run:

```
>>> Waiting for management-cluster nodes to be ready...
error: no matching resources found
Error: 'kubectl --kubeconfig /root/.kube/krops-mgmt.yaml wait --for=condition=Ready node --all --timeout=15m' failed with exit status: 1
```

## Why

`kubectl wait --all` errors with "no matching resources found" when the selector matches zero objects — it does not wait for objects to appear. On EKS the kubeconfig secret (which Phase 1 waits on, then Phase 2 exports) exists as soon as the EKS control plane is ACTIVE. The worker node group, however, is created by CAPA's MachinePool *after* control-plane ready, so at Phase 2 the target cluster has zero nodes registered. The wait therefore errors before the node group has even started launching.

Observed 2026-09-18 during a full AWS e2e run (clean account):

- `default_eu-north-1-management-control-plane` went `ACTIVE` (v1.34) → kubeconfig secret appeared → pivot Phase 2 exported it.
- Node group `default_eu-north-1-management-pool` (desired 2× t4g.medium) was created 09:47:50, ~1-2 min after the control plane was ACTIVE; at that point Phase 2's node wait had already failed.
- The pivot never reached Phase 4 (`clusterctl move`); the kind bootstrap cluster stayed authoritative and no AWS workers were lost.

The same line exists in `pivot.sh:162` — so the original script shares the defect. In script-era operation it was masked by the human running `pivot.sh` manually, several minutes after `bootstrap.sh`, by which time the node group had registered nodes. The CLI port made the pivot the default exit, so the timing gap now hits every clean first run.

## Impact

Every clean first-run AWS bootstrap fails at pivot Phase 2 (after EKS spend has started: control plane + node group). Recovery is re-running the lifecycle once nodes exist (documented "Pivot recovery"), which works because the phases are rerun-safe. The kind bootstrap cluster is not deleted, so no state is lost; but the operator sees a scary failure mid-pivot.

## Suggested fix

Replace the `wait node --all` with a poll that tolerates a nodeless cluster: wait (up to the same 15m) until at least one `Ready` node exists *and* all registered nodes are Ready, e.g.:

```
until kubectl get node -o name | grep -q .; do sleep 10; done   # first node appears
kubectl wait --for=condition=Ready node --all --timeout=15m
```

(or a single `wait_for_resource`-style loop counting Ready nodes against `MGMT_CLUSTER`'s desired). Note the desired replica count comes from the MachinePool (2 for the AWS mgmt pool), so "all registered nodes Ready" is the right success condition once the node group has reported capacity; optionally also require the EKS node group `ACTIVE` for extra certainty.

## Alternatives

- Move the node wait to *after* Phase 4 (the move): the move only needs the target API + CAPI (installed in Phase 3), not workers. The node wait's purpose is to have a live target before Flux is seeded (Phase 5), so it could move between Phase 4 and Phase 5 with the same poll fix.

shrinedogg:
Fixes #348
Fixes #349

## Summary

Two timing gaps failed every clean first AWS run (observed on the 2026-09-18 e2e). Both were masked in the script era by the human running `pivot.sh` minutes after `bootstrap.sh`; the CLI port made the pivot the default exit, so the gaps now hit every clean run.

**#348 — bootstrap→pivot chain fails before Flux has created the mgmt Cluster object.**
The pivot's Phase 0 required the management `Cluster` object to already exist in the kind bootstrap cluster, but creating it is Flux's job after the handoff (the CAPA HelmRelease, providers, identity, and clusters overlay — observed ~7 min after bootstrap completion). The fail-fast check is now a poll with the Phase 1 budget (`MGMT_READY_TIMEOUT` / `MGMT_POLL_INTERVAL`, no new knobs). On timeout it lists the flux-system Kustomizations, so a stuck Flux is distinguishable from one still reconciling, and fails with rerun-safe guidance.

Placement note: the issue asks for the poll "before entering `run_pivot`". The wait sits at the top of the pivot, after the context preflight, which is that same boundary — `run_pivot` is only entered from the bootstrap chain — and keeps two properties the literal placement would lose: a wrong-context invocation still fails fast instead of polling for 40m first, and `pivot.sh` (which owns the per-profile cluster name) gets the identical structure, so `bootstrap.sh` needs no change.

**#349 — Phase 2 node wait fails on a nodeless EKS target.**
`kubectl wait --for=condition=Ready node --all --timeout=15m` errors "no matching resources found" while the target has zero nodes, and the CAPA MachinePool registers nodes 1-2 min after the control plane goes ACTIVE. The bare wait is now a poll with the same 15m budget: at least one node exists AND every registered node is Ready. On timeout it prints the node listing and `clusterctl describe cluster` with rerun-safe guidance.

## Parity (bootstrap-rs ↔ pivot.sh)

Both fixes land in both implementations, keeping step order, messages, and the env interface in sync (per the teardown parity rule). The CLI parses `kubectl get nodes -o json` (`nodes_readiness`); the script counts Ready rows from the same listing. The timeout diagnostics use `kubectl get kustomizations` rather than the flux CLI because `flux` is only in local-host's required tools.

## Tests

- New Rust unit tests drive the poll loops with scripted probes on tokio's paused clock (`test-util` as a dev-only feature; no new crates, lockfile unchanged): object-appears-after-delay succeeds, stuck-Flux surfaces the Kustomization diagnostic and fails naming the object and budget, nodeless-then-registers succeeds, permanently-nodeless never succeeds early, both timeout paths fail with actionable messages, and `poll_until`'s past-budget probe math is pinned.
- pivot.sh's new functions probed behaviorally with stubbed `kubectl`/`clusterctl`: object appears on the 3rd poll → success; definition timeout → exit 1 + Kustomization diagnostic + guidance; nodeless→NotReady→Ready → success; nodeless timeout → exit 1 + node listing + `clusterctl describe` + guidance (14 assertions, all pass).

## Verification

- `cargo fmt --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked` (101 passed) in `bootstrap-rs/`
- `bash -n pivot.sh` and `mise run validate` from the repo root (exit 0)
- Rebased on `924a005` (the merged #350 fix); disjoint regions, no conflicts

The destructive happy path (a real clean AWS e2e) was not run; it is the next acceptance run's job. Docs updated: `docs/bootstrap-cli.md` (controls table + phase summary), `docs/operations.md` (Pivot recovery).


shrinedogg (commented):
Verdict: clean. Both fixes verified in bootstrap-rs and pivot.sh; no blocking findings. One non-blocking suggestion below.

Verified locally (PR head b6932d4, every claim re-run; green CI not relied on):

- Parity: the Phase 0 Cluster poll and the nodeless-tolerant Phase 2 node wait land in both implementations with the same step order, messages, and env interface. MGMT_READY_TIMEOUT / MGMT_POLL_INTERVAL are reused (no new knobs), and MGMT_NODE_READY_TIMEOUT is a non-overridable 15m constant on both sides. The fail-fast Cluster check is gone from both pivot_preflight and run_pivot with no remaining callers or stale doc references.
- bootstrap-rs: cargo fmt --check, cargo clippy --locked --all-targets -- -D warnings, cargo test --locked: 101 passed, including the 9 new tests (poll_until probe-count pinning, cluster-definition success and stuck paths, nodes_readiness JSON boundaries including zero nodes and unparseable kubectl output, nodeless-then-registers, both timeout messages). Cargo.lock is untouched; tokio test-util is dev-only.
- pivot.sh: bash -n passes, shellcheck -S warning is clean, and I ran an independent stub harness (stubbed kubectl/clusterctl/date on PATH, the functions sourced from pivot.sh) covering: duration_to_seconds matches the Rust parser on 40m/2h/90s/bare seconds; Cluster appearing on the 3rd poll succeeds with exactly 3 probes and no diagnostic; the stuck path exits 1 naming the object and budget, prints the Kustomization listing, and gives rerun-safe guidance; nodeless then NotReady then Ready succeeds; a permanently nodeless target burns the budget, never succeeds early, then exits 1 with the node listing, clusterctl describe, and guidance. 18 assertions, all pass.
- Boundary inputs: zero nodes keeps polling (also pinned by nodes_wait_never_succeeds_while_nodeless), zero Cluster objects polls to timeout, and a failed/unreachable kubectl parses as keep-waiting rather than erroring the poll. Mixed Ready/NotReady listings wait correctly in both implementations (Rust checks the Ready condition per item; the script counts awk '$2 ~ /^Ready/' rows, which excludes NotReady and Unknown).
- Docs: docs/bootstrap-cli.md (controls table, phase summary) and docs/operations.md (Pivot recovery) describe the new waits accurately.
- Timeout diagnostics print object names, namespaces, and budgets only; no secret material.

Suggestion (non-blocking, follow-up is fine): the description cites 14 shell-side probe assertions, but nothing shell-side is committed, so nothing guards pivot.sh's new functions against regression (mise run validate covers pivot.sh with bash -n only). The repo already runs committed behavioral shell tests in validate (tests/test-toolbox-run-exec.py stubs binaries on PATH and asserts behavior); the same pattern would fit the two new wait functions and keep the shell half as protected as the Rust half.

Note for the record: using `kubectl get kustomizations` (not the flux CLI) for the stuck-Flux diagnostic is right, since pivot preflight requires kubectl but not flux, and the Flux Kustomization CRD carries Ready/Status printer columns, so the listing is actionable.

The destructive happy path (a real clean AWS e2e) was not run and is correctly deferred to the next acceptance run.

Issue #355, opened 2026-09-18
gcp/aws/azure: WireMock-virtualized e2e tests for CAPA/CAPZ/CAPG controllers


lmilbaum:
## Context

Live acceptance runs (like #319 for GCP) need real cloud credentials and a real project/subscription/account, so there's no fast, repeatable CI signal that catches config drift or controller wiring regressions between live runs, across any of krops's three cloud environments (`aws`, `azure`, `gcp` in `bootstrap.toml`).

This plan adds a **virtualized e2e test per cloud**: WireMock intercepts each environment's controllers' outbound cloud API calls on a kind cluster and replays recorded, stateful stub responses, so CI can exercise the real authentication chain and reconciliation logic without touching real cloud APIs. This is a regression safety net, not a replacement for any cloud's live acceptance run. It never validates that real infrastructure actually provisions correctly, only that krops's controllers are wired correctly and issue the calls we expect.

All three clouds are designed together from the start, sharing one harness shape, rather than building GCP first and generalizing later.

## What differs per cloud

| | Controllers | Auth mechanism | Endpoint-override support |
|---|---|---|---|
| AWS | CAPA (`capa-system`) + ACK (S3/RDS/IAM, on the management cluster) | Static IAM access keys, SOPS-encrypted, no WIF/federation chain | Likely yes: AWS SDK Go v2 (used by both CAPA and ACK) officially supports `AWS_ENDPOINT_URL` / per-service `AWS_ENDPOINT_URL_<SERVICE>` env vars, the same mechanism LocalStack-style testing relies on. Needs a spike to confirm it's honored by CAPA's/ACK's specific pod configs, but lowest risk of the three. |
| Azure | CAPZ (`capz-system`) + ASO (Azure Service Operator) | Arc-connect + workload identity federation (FIC) | Plausible: Azure SDK for Go (`azcore`) supports `cloud.Configuration` with a custom `Services` map for per-ARM-provider endpoint overrides (used for Azure Stack/sovereign clouds), but whether CAPZ/ASO's Helm charts or CRs expose a way to inject that at deploy time is unconfirmed, needs its own spike. |
[...]

PR #391, opened 2026-09-20
docs(azure): WireMock e2e Phase 0 spike findings: ASO config surface yes, CAPZ via CoreDNS rewrite

shrinedogg:
Azure Phase 0 spike for the WireMock-virtualized e2e plan. Scratch kind cluster, `clusterctl init --infrastructure azure:v1.27.0` (repo pin, bundles ASO v2.19.0), bare WireMock 3.13.2 with the throwaway-CA pattern from the AWS findings. Dummy GUID credentials throughout; no real Azure contact except the unassisted token probes that serve as negative evidence (all rejected by real AAD).

Decision gate:

| Controller | Config-surface override | Fallback |
|---|---|---|
| ASO v2.19.0 | yes (`aso-controller-settings` Secret: `AZURE_AUTHORITY_HOST` / `AZURE_RESOURCE_MANAGER_ENDPOINT` / `AZURE_RESOURCE_MANAGER_AUDIENCE`; same three templated by clusterctl and exposed as ASO Helm values). Proven to arrival: token POST with WireMock audience scope, `ResourceGroup` PUT at the WireMock ARM endpoint. | Custom authority additionally needs CoreDNS rewrite of `login.microsoftonline.com` for MSAL's instance-discovery anchor (ASO has no disable knob) plus static discovery/openid/token stubs |
| CAPZ v1.27.0 | no. `azureEnvironment` takes 4 named clouds only (`AzureStackCloud` rejected at runtime), endpoint env vars provably ignored (real AADSTS90002 with the env vars set, journal empty), no CR fields | CoreDNS rewrite of `login.microsoftonline.com` + `management.azure.com` works end to end (auth chain + ARM skus + ASO-managed RG all arrive). `HTTPS_PROXY` is not viable: WireMock answers CONNECT with a hollow 200 and cannot tunnel TLS |

Phase 1 mechanism: combined ASO endpoint configuration + CoreDNS rewrite. Bonus finding: CAPZ's ASOAPI-created resources and per-cluster ASO credentials inherit the global ASO endpoint settings automatically, so the `AzureASOManaged*` path krops uses is covered by the one Secret patch.

New harness findings beyond the AWS list (WireMock 3.13.2 `--port` vs `--http-port`, `MissingResourceException` on bad flags, journal reset via `DELETE /__admin/requests`, keep-alive requires a manager restart after the rewrite, the probe-relax fix for the loaded-host leader-election flap) are in the doc's harness and caveats sections.

Deliverable: `docs/wiremock-e2e-spike-findings-azure.md`, registered in the mkdocs nav and the README documentation table. Verified with `mise run docs-build` and `mise run docs-test`. Scratch cluster deleted.

Refs #355


shrinedogg (commented):
Review verdict: clean, no findings. Both components are covered by a confirmed mechanism and the doc does not overclaim: ASO by the aso-controller-settings surface (token POST with the WireMock audience, ResourceGroup PUT at the WireMock ARM endpoint), CAPZ by the CoreDNS rewrite (discovery, openid-config, token, and the skus GETs all recorded arriving, with absoluteUrl still management.azure.com), and CAPZ's lack of a config surface is documented as three evidenced negatives rather than glossed over.

Verified against upstream sources at the pinned versions:

- CAPZ pin matches the repo: mgmt/azure/capi-providers/capz-system/providers.yaml pins v1.27.0.
- ASO bundling confirmed: CAPZ v1.27.0 config/aso/kustomization.yaml pulls the ASO v2.19.0 release bundle, image mcr.microsoft.com/k8s/azureserviceoperator:v2.19.0.
- ASO config surface confirmed in all three places claimed: ASO v2.19.0 v2/pkg/common/config/config.go defines the three keys; CAPZ config/aso/settings.yaml templates all three from clusterctl variables; the ASO Helm values.yaml exposes the same three. The release infrastructure-components.yaml mounts them as optional secretKeyRef env entries on the ASO Deployment.
- CAPZ negative confirmed at source: azure/scope/clients.go switches on exactly the four named clouds, and the default branch prints c.CloudEnvironment() before it is assigned, which explains the empty %q in the captured error. Only AZURE_SUBSCRIPTION_ID is read from the process env, and identity.go constructs the credential with an explicit Cloud block, so azidentity's env fallback never engages. The rendered capz-controller-manager Deployment carries no other Azure env wiring.
- ALLOW_MULTI_ENV_MANAGEMENT defaults to false per config.go ("This defaults to false for security reasons"), supporting the bonus finding that per-cluster credentials inherit the global endpoints.
- DisableInstanceDiscovery: zero hits grepping the full ASO v2.19.0 Go source tree, matching "no disable knob".
- azcore v1.23.1 transport_default_http_client.go sets Proxy: http.ProxyFromEnvironment, so the conclusion that WireMock's hollow CONNECT 200 kills the HTTPS_PROXY path is sound.
- The CAPZ user-agent sample ending v1.27.0-dirty is authentic, not a typo: the manager binary inside registry.k8s.io/cluster-api-azure/cluster-api-azure-controller:v1.27.0 contains the baked v1.27.0-dirty version string.
- Version table matches reality: SDK pins agree with go.mod at both tags (CAPZ azcore v1.23.1 / azidentity v1.14.1; ASO azcore v1.21.1 / azidentity v1.13.1), and cert-manager v1.21.1 is clusterctl v1.14.2's CertManagerDefaultVersion.
- CRD details check out: no enum on azureEnvironment (plain string), and clientSecret is a plain {name, namespace} ObjectReference with no key field.
- Hygiene: dummy GUIDs only (all-zero / all-one / all-two patterns), no tenant or subscription leakage, no em-dashes or emojis. mkdocs nav and README documentation table both registered. Scratch cluster deleted (kind lists no wiremock-azure-spike). CI green including the strict docs build.

